PR #7, opened 2018-09-25
Enable CI through Travis

junghans:


junghans:
Yes

sslattery:
Will this also run the unit tests? Curious as to if we can add `ctest` to the script. Also is there a way to verify that this works? I'm assuming this preliminary setup will run on some free GitHub CI.

junghans:
You can do everything, you just `script:` block.

junghans:
Ok, it runs the tests now and add coverage reporting.

PR #8, opened 2018-09-26
Sorting Cleanup

sslattery:
This PR does the following:

- [x] `LinkedCellList` becomes its own object. It has its own associated implementation of `permute` for `AoSoA` objects.

- [x] Improves the testing of `LinkedCellList`

- [x] Templates sorting data on the Cabana memory spaces rather than kokkos memory spaces.

- [x] Cleans the sort interface functions names for consistency.


junghans:
Hmm, I still don't understand why the gitlab CI isn't triggered, most have something to do with the fact that the PR is coming for a fork, I guess.

core/src/Cabana_Sort.hpp
```diff
@@ -114,173 +98,26 @@ class BinningData
...
 
 namespace Impl
 {
 //---------------------------------------------------------------------------//
 // Create a permutation vector over a range subset using a comparator over the
 // given Kokkos View of keys.
 template<class KeyViewType, class Comparator>
-BinningData<typename KeyViewType::memory_space>
+BinningData<
+    typename KokkosSpaceToCabana<typename KeyViewType::memory_space>::type>
```
rfbird: That is a really nice enhancement, good stuff

core/src/impl/Cabana_CartesianGrid.hpp
```diff
@@ -20,52 +20,62 @@ namespace Cabana
...
-    Scalar _max_x;
-    Scalar _max_y;
-    Scalar _max_z;
-    Scalar _dx;
-    Scalar _dy;
-    Scalar _dz;
-    Scalar _rdx;
-    Scalar _rdy;
-    Scalar _rdz;
+    double _min_x;
```
rfbird: As per my comment on Slack, I don't really like templating on data stype..but not do I like being married down to double (I think it makes this Cartesian Grid less generic, even though it makes sense for the Vertlet being the user)

core/src/impl/Cabana_CartesianGrid.hpp
```diff
@@ -75,11 +85,25 @@ class CartesianGrid
         num_z = _nz;
     }
 
+    // Get the number of cells in a given direction.
+    KOKKOS_INLINE_FUNCTION
```
rfbird: Did we want to wrap this `Kokkos_xxxx` with `Cabana_xxxx`?

core/unit_test/tstSort.hpp
```diff
@@ -385,7 +385,7 @@ void testBinByMemberDataOnly()
     // Bin the aosoa by the 1D member. Use one bin per data point to
     // effectively make this a sort. Don't actually move the particle data
     // though - just create the binning data.
-    auto bin_data = Cabana::binByMember( aosoa.slice<1>(), num_data-1 );
+    auto bin_data = Cabana::binByKey( aosoa.slice<1>(), num_data-1 );
```
rfbird: Is this planned to be the last version of the `sort` interface before we lock it down?

rfbird: (not that I have a problem with it, I just hope to play around with it a bit for my PIC use case if possible before marrying down to it. As per our last discussion on this I think we had my requirements captured)

core/src/impl/Cabana_CartesianGrid.hpp
sslattery: I'm torn if we should propagate the Cabana macros to implementation details

core/unit_test/tstSort.hpp
sslattery: This PR is hopefully the last update to the code API before tagging a release candidate. This captures your requirements I think but I wanted to make it more consistent in terms of naming convention.

PR #9, opened 2018-09-26
Enable gitlab CI

junghans:
Could be used for GPU support later on


junghans:
This will be another CI abd you can register your GPU machine with gitlab.com as a runner, so that we get Cuda testing as well.

PR #10, opened 2018-09-28
cmake: first try to remove tribits

junghans:


junghans:
Converted all tests, Cuda still untested.

rfbird:
Eventually we need to find a way to determine which GPU devices are supported.

it increasingly looks like it might be best to just inherit whatever Kokkos did?

rfbird:
Looks good to me ! I'll make an issue about the `pthreads` incompatibility. 

Also, just so it's noted somewhere, one has to build Cabana with the same compiler you used for Kokkos..

PR #18, opened 2018-10-03
Adding Pthreads backend

sslattery:
This adds the Pthreads execution space to Kokkos and therefore Cabana. To enable, build Kokkos with Pthreads and then add `-DCabana_Enable_Pthread=ON` to your configure. Details:

1. Adds Pthreads as a backend build options
2. Refactors CMake to default to Serial as on and the rest of the backends as off. A user must explicitly enable a backend rather than it being enabled if the supporting library is found. 
3. Remove `QUIET` option from package discovery - I like to see what configure is looking for.
4. Adds these new options to the Travis CI
5. Adds a pthreads build to Travis. OpenMP and Pthread runtimes seem to stomp on each other so this seemed like a good idea.


sslattery:
@junghans I tried to add a separate build to Travis for Pthreads - did I do this right?

sslattery:
Ultimately these tests are still failing in the same manner I saw on my linux box when I had a kokkos build with both OpenMP and Pthread enabled - basically the error comes down to the runtime initializing threads and is therefore not a Cabana problem. Does it make sense to have a build that has both Pthreads and OpenMP? If not we should either only test serial and OpenMP on Travis or also setup a separate kokkos build that just has Pthreads.

.travis.yml
```diff
@@ -19,10 +19,20 @@ addons:
...
+          -DCabana_ENABLE_Serial=ON -DCabana_ENABLE_OpenMP=ON
+          -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_EXAMPLES=ON
           -DEXTRA_LINK_FLAGS="--coverage" -DCMAKE_CXX_FLAGS="--coverage" .. &&
     make -j2 VERBOSE=1 &&
+    make test &&
+    cd ../pthread_build &&
+    cmake -DCMAKE_PREFIX_PATH=$HOME/kokkos
+          -DCabana_ENABLE_Pthread=ON -DCabana_ENABLE_Serial=OFF
+          -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_EXAMPLES=ON .. &&
+    make -j2 VERBOSE=1 &&
```
junghans: Why are we doing two separate builds instead of one with everything enabled?

.travis.yml
```diff
@@ -19,10 +19,20 @@ addons:
...
+  - export OMP_NUM_THREADS=4 &&
+    mkdir pthread_build && mkdir serial_openmp_build &&
+    cd serial_openmp_build &&
+    cmake -DCMAKE_PREFIX_PATH=$HOME/kokkos
+          -DCabana_ENABLE_Serial=ON -DCabana_ENABLE_OpenMP=ON
+          -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_EXAMPLES=ON
           -DEXTRA_LINK_FLAGS="--coverage" -DCMAKE_CXX_FLAGS="--coverage" .. &&
     make -j2 VERBOSE=1 &&
+    make test &&
+    cd ../pthread_build &&
```
junghans: I would do the second build in a different environment!

CMakeLists.txt
```diff
@@ -18,15 +18,28 @@ set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 find_package(KOKKOS REQUIRED)
 
 option(Cabana_ENABLE_Serial "Build Cabana with Serial support" ON)
-find_package(OpenMP QUIET)
-option(Cabana_ENABLE_OpenMP "Build Cabana with OpenMP support" ${OpenMP_FOUND})
+
+option(Cabana_ENABLE_Pthread "Build Cabana with Pthread support" OFF)
+if( Cabana_ENABLE_Pthread )
+  find_package(Threads)
```
junghans: Do we really need to find the thread package? Isn’t it inherited from Kokkos.

sslattery: I guess I would rather be thorough here

sslattery:
The Pthreads build is still failing. I think we actually need two separate kokkos builds one with serial+openmp and one with serial+threads. I think the issue is the kokkos runtime is initializing both openmp and pthreads and they seem to be clashing

junghans:
@sslattery: fixed, but we still need to report this upstream!

CMakeLists.txt
junghans: Ok

PR #22, opened 2018-10-04
cmake: run OpenMP test with different # threads

junghans:


sslattery:
Is there a way to do this for Pthreads too?

junghans:
@sslattery Done!

And how do you set the number of threads for pthreads?

rfbird:
This now installs headers into `$PREFIX/include/Cabana`, is that intended? I think before it was `$PREFIX/include`

junghans:
@rfbird I don't know, we can easily do it with the `Cabana` subdir.

rfbird:
I think the way it is right now breaks the includes

junghans:
Huh, how?

rfbird:
Hmm well I guess that's misleading, I just mean it's in a different place to before

Before I would just look in `Cabana/include` now I have to look in `Cabana/include/Cabana`

Not a huge deal

junghans:
@sslattery let me know how you want it and I will fix it.

sslattery:
@junghans I would like things in the include dir as they were before - I think @rfbird is correct in that it will break header links in installed builds. Also our headers are prefixed with `Cabana_` so we shouldn't need a prefix directory

core/unit_test/CMakeLists.txt
```diff
@@ -44,8 +44,11 @@ if(Cabana_ENABLE_Pthread)
   foreach(_test AoSoA Slice DeepCopy Tuple Sort LinkedCellList NeighborList Parallel)
     add_executable(${_test}_test_Pthread Pthread/tst${_test}_Pthread.cpp unit_test_main.cpp)
     target_link_libraries(${_test}_test_Pthread cabanacore cabana_core_gtest)
-    add_test(NAME ${_test}_test_Pthread COMMAND ${_test}_test_Pthread --gtest_color=yes)
-    set_tests_properties(${_test}_test_Pthread PROPERTIES FAIL_REGULAR_EXPRESSION "${FAIL_RE}")
+    foreach(_thread 1 2)
+      add_test(NAME ${_test}_test_Pthread_${_thread} COMMAND
+        ${_test}_test_Pthread --gtest_color=yes --kokkos-threads=${_thread})
```
sslattery: Nice - yeah was going to tell you this was the way to add multiple pthreads to the kokkos runtime. We can do this instead of setting `OMP_NUM_THREADS` as well.

sslattery:
so 3eba49d adds a CabanaConfig.cmake file? @rfbird did you test out using this file?

core/unit_test/CMakeLists.txt
junghans: On the other hand, with this, we could "simplify" the testing in Cmake by using a triple nested `foreach()` loop, do you want that?

sslattery: if you think that will be cleaner and easier to maintain go for it - otherwise I think this is fine the way it is

junghans: Done!

sslattery:
@junghans ready to merge?

junghans:
@sslattery yes, but please test Cuda on more time.

sslattery:
Heading to the beach for the weekend. Will do the final cuda testing Sunday

PR #25, opened 2018-10-08
.travis.yml: enable ccache for clang

junghans:


junghans:
Yeah, it is automatic for `gcc`, but not `clang`.

PR #26, opened 2018-10-08
Version 0.1 Tutorial

sslattery:
This adds basic examples for the core 0.1 release elements:

- [x] Initialization/Finalization (Hello World)
- [x] Tuple
- [x] SoA
- [x] AoSoA
- [x] Slice
- [x] DeepCopy
- [x] Sort
- [x] LinkedCellList
- [x] VerletList


core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
```diff
@@ -0,0 +1,219 @@
+/****************************************************************************
```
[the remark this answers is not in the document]
sslattery: I would say that this in fact demonstrates "portability" - least in the context of version 0.1. Because we have no parallel for concept in this release we will need to rely on the users to use their own programming model. I think these examples show the same code can be used with the different programming model (kokkos or otherwise).

What I could do is combine these into a single file and use `#ifdef CABANA_ENABLE_somebackend` the protect different implementations. Note sure exactly how I would implement the rest of that but I bet it is possible and would give the user a single file to look at.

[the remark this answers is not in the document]
rfbird: I personally think that one file could be better, but I'm happy to be flexible

core/example/04_aosoa_openmp/aosoa_openmp.cpp
```diff
@@ -0,0 +1,225 @@
...
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#include <Cabana_Core.hpp>
+
+#include <iostream>
+
+//---------------------------------------------------------------------------//
+// AoSoA example using OpenMP.
```
guangyechen: It may be better to make it more explicit by stateing that "AoSoA example using OpenMP parallel for for particle initialization".

core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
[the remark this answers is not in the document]
sslattery: So it's not clear to my be looking at this how I would combine these into one file without making a mess. These examples demonstrate our portable data structures. It's hard to cleanly demonstrate portable parallelism without our `parallel_for`. Once we push that in version 0.2 I think then we will have a single clean example with `AoSoA`, `Slice`, etc. for the user to look at.  

What if we refactored these examples in version 0.2 once we have a `parallel_for`?

core/example/02_tuple/tuple_example.cpp
```diff
@@ -0,0 +1,118 @@
...
+         4) its base class and non-static data members (if any) are themselves
+            also trivial types
+
+      *: The types in a Cabana tuple are defined via a type parameter
+         pack using the Cabana::MemberTypes class
+
+      *: Cabana tuples are designed to store and access multidimensional array
+         data with accessors provided to obtain references to individual array
+         values within the tuple.
+    */
```
guangyechen: For the tuple example, It may be worth adding that "Cabana tuple's memory layout is consistent with order of declaration, i.e., what declared first is first in memory. (Note that libstdc++ has reverse order for std::tuple.)

core/example/03_struct_of_arrays/soa_example.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include <iostream>
+
+//---------------------------------------------------------------------------//
+// SoA example.
+//---------------------------------------------------------------------------//
+void soaExample()
+{
+    /*
+      Cabana structs-of-arrays (SoAs) are groups of Tuples with the data
+      specified in an order such that the tuple index is stride-1 in
```
guangyechen: "tuple index is stride-1..." may be "tuple's member array index is stride 1..."?

core/example/03_struct_of_arrays/soa_example.cpp
```diff
@@ -0,0 +1,142 @@
...
+      memory. This results in each dimension of each multidimensional tuple
+      member being stored in a contiguous memory block the size of the number
+      of tuples. The number of tuples in an SoA is defined as the "vector
```
guangyechen: "number of tuples" may be "number of each tuple member"?

core/example/03_struct_of_arrays/soa_example.cpp
```diff
@@ -0,0 +1,142 @@
...
+      length".
+
+      For example, consider declaring the following member type to store two
+      arrays of doubles in each tuple:
+
+          using types = Cabana::MemberTypes<double[3][2],double[4]>;
```
guangyechen: It may be more complete and consistent with the code below by adding a scalar member.

core/example/03_struct_of_arrays/soa_example.cpp
```diff
@@ -0,0 +1,142 @@
...
+      data layout:
+
+          struct MyEquivalentSoA
+          {
+              double d0[2][3][veclen];
+              double d1[4][veclen];
+          };
+
+      Note: The data in this struct definition with an equivalent memory
+      layout is stride-1 in the vector index.
```
guangyechen: I would prefer using "array" instead of "vector" here, as in "native c/c++ array", which is used in the struct.

core/example/03_struct_of_arrays/soa_example.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+      Note: The higher dimensional indices have their order reversed (see the
+      order of the data of member `d0` in the equivalent struct
+      definition). This is to create a Fortran-equivalent or column-major
+      ordering of the data relative to the vector index.
+
```
guangyechen: It is not clear that why this is the case. There is especially no reason to create a Fortran-equivalent struct definition. I would prefer an memory layout order that is consistent with SOA declaration, if possible.

core/example/03_struct_of_arrays/soa_example.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                          float[4],
+                                          int>;
+
+    /*
+      Next declare the vector length of our SoA. This is how many tuples the
+      SoA will contain. A reasonable number for performance should be some
+      multiple of the vector length on the machine you are using.
+    */
+    const int VectorLength = 4;
+
```
guangyechen: As a side thought: I am wondering if there is a way to automatically figure out the vector length supported by the device where the code compiles? If there is, then one can just say 
const int VectorLength = DefaultVeclen

core/example/03_struct_of_arrays/soa_example.cpp
sslattery: Each member is one component of a tuple - so in this case the array size or vector length can also be viewed as the number of tuples in the SoA

core/example/03_struct_of_arrays/soa_example.cpp
sslattery: good call

core/example/03_struct_of_arrays/soa_example.cpp
sslattery: This was fixed with #28

core/example/03_struct_of_arrays/soa_example.cpp
sslattery: So we have presets for this - the user doesn't have to define it. In the future we can automatically select the vector length based on the device type flags we get from kokkos. Should be pretty easy to do. I will make an issue for this.

core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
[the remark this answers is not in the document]
sslattery: @rfbird does this work for you?

sslattery:
@guangyechen do you plan on doing any more review on this one?

core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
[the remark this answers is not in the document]
rfbird: Sure sounds fine to me. I don't personally care, what worries me is that someone will see this, think they have to write device code and we'll never get a second chance to capture them as a user. More of a management concern that a developer one

guangyechen:
@sslattery Sorry that I just got back from a long discussion with Bob. Yes, I am planning on finishing it maybe tonight. Thanks.

core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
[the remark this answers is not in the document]
guangyechen: Maybe we could put some of the simple examples into one dir, which I believe can be useful for some people. The important ones will be those with parallel_for which may be in another dir.

core/example/03_struct_of_arrays/soa_example.cpp
guangyechen: It cause some confusion to me because a tuple is an object that could hold one or many "particles". Are you using "particles" and "tuples" interchangeably?

core/example/tutorial/06_deep_copy/deep_copy.cpp
```diff
@@ -0,0 +1,130 @@
...
+            for ( int a = 0; a < src_aosoa.arraySize(s); ++a )
+                soa.get<1>(a,i) = 1.0 * (a + i);
+
+        for ( int a = 0; a < src_aosoa.arraySize(s); ++a )
+            soa.get<2>(a) = a + 1234;
+    }
+
+    /*
+      Deep copy the data from the source to the destination. A deep copy is
+      possible if the AoSoAs have the same member types and they are the same
```
guangyechen: It would be better adding a "two" in front of "AoSoAs" on line 87. The first time I looked at it, I was thinking of within a AoSoA, how could it be having the same member types, as it is clearly not the case (some doubles, floats, and ints).

core/example/tutorial/07_sorting/sorting.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+        std::cout << "Tuple " << t
+                  << ", member 0: " << tp.get<0>() << std::endl;
+
+        std::cout << "Tuple " << t
+                  << ", member 1: " << tp.get<1>() << std::endl;
+    }
+    std::cout << std::endl;
+}
+
```
guangyechen: Is there a way to get the number of values in each bin?

guangyechen:
@sslattery It appears that you have used the term "tuple" in most places we used to call particles. Since tuple is a type/object in standard C++, it could cause some confusions, especially in the context of SoA, where the SoA itself can be looked at as one tuple object (which may hold many particles).

core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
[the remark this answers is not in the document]
sslattery: I'm starting to think should we even have Cuda/OpenMP variants at all at this point without our `parallel_for`? I do want users to know that they can use this data structure with different programming models but in ideal cases they shouldn't have to do much of anything with a different programming model when it comes to particles.

What I could do is just keep the serial example for now and explain in comments how to change the memory space in which things are allocated and that the data structures can be used with programming models compatible in those memory spaces.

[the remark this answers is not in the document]
sslattery: Ultimately I agree with Bob's point - we don't want potential users to think they need to use other programming models for their particles.

core/example/tutorial/06_deep_copy/deep_copy.cpp
sslattery: I have clarified this comment

core/example/tutorial/07_sorting/sorting.cpp
sslattery: Yes - the sorting/binning returns a `Cabana::BinningData` object. The function `bin_data.binSize( bin_id )` gets the number of values in the bin with the given id while the permutation vector gives the bins themselves and offsets tells you where each bin starts in the permutation vector.

I'll add some to the example explaining how to get the binning data information.

sslattery:
I view the subset of data in an AoSoA associated with a single particle to be a `Tuple` and therefore an `SoA` is an array of `Tuple` with a special data layout (where AoS is really just an array of `Tuple`). From that point of view what we have is nearly identical to the C++ STL except we support multidimensional data accessors.

So in other words, think of `Tuple` as the core single element or little packet of data and AoS, `SoA`, and `AoSoA` as different ways of grouping `Tuple` objects together in memory. By building from a `Tuple` example I was hoping that the tutorial would expose the user to that concept.

core/example/tutorial/04_aosoa_cuda/aosoa_cuda.cpp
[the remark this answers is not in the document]
sslattery: OK - just pushed a commit to this effect

PR #27, opened 2018-10-09
Making binning data and permutation interface consistent 

sslattery:
The use of begin and end for binned/sorted subranges was inconsistent. This changes stores begin/end directly in the binning data and reuses it for the permutation.


sslattery:
@rfbird Can I get a quick review on this one?

PR #39, opened 2018-11-05
Custom Cabana Slice Layout for Kokkos::View

sslattery:
Adding new kokkos view layout and offset computation for slices to eliminate stride 1 multiplication in indexing.


core/src/Cabana_Slice.hpp
```diff
@@ -22,6 +22,333 @@
...
+                       size_t N6 = 0, size_t S6 = 0,
+                       size_t N7 = 0, size_t S7 = 0 )
+    : LayoutStride( N0, S0,
+                    N1, S1,
+                    N2, S2,
+                    N3, S3,
+                    N4, S4,
+                    N5, S5,
+                    N6, S6,
+                    N7, S7 )
```
dalg24: I believe you should be able to use variadic templates and just forward all arguments to the constructor of `Kokkos::LayoutStride`.

dalg24: Actually, just remove that constructor definition and replace it with `using LayoutStride::LayoutStride;`

dalg24:
Why did you mark my comments are resolved?  You haven't addressed them, or did I miss something??

sslattery:
I refactored this to where they were no longer relevant. The layout no longer inherits from `LayoutStride`

PR #40, opened 2018-11-06
Add optimization flags by default

### LINKED ISSUE #37: Change Default Compiler Flags to favor aggressive optimization
As per some previous offline discussion, I think it might be good to do some combination of the following:

1) Make the default build type `release` (which likely gives `-O3`)
2) Consider adding compiler flags to add platform specific code (`-march=native`, `-xHost` etc)

@junghans  has already done this for another project, so it should be fairly straight forward to port over

junghans:
Fix #37


junghans:
@rfbird, `icc` support `-march=native` as well, but is it the same as `-xHost`?

rfbird:
> @rfbird, `icc` support `-march=native` as well, but is it the same as `-xHost`?

It's not clear to me if it's directly equivalent (but it's definitely close). I think it's fine for a first step

CMakeLists.txt
```diff
@@ -58,6 +63,51 @@ find_package(Doxygen)
...
+#------------------------------------------------------------------------------#
+# Architecture
+#------------------------------------------------------------------------------#
+if(BUILD_MARCH AND BUILD_MCPU)
+  message(FATAL_ERROR "BUILD_MARCH and BUILD_MCPU are incompatible")
+endif()
+
+# Try -march first. On platforms that don't support it, GCC will issue
+# a hard error, so we'll know not to use it.
+if(BUILD_MARCH)
```
sslattery: is `BUILD_MARCH` a native CMake variable?

junghans: No, we still need to a cache doc string for it!

sslattery: ah OK - didn't see it elsewhere in the configure

junghans:
Don’t merge yet I still want to move the coverage injection into an option, too!

sslattery:
@junghans can you add `[WIP]` to the pull request title indicating you still have work left to do?

junghans:
@rfbird @sslattery ready to merge

junghans:
@sslattery ours will overwrite it. We could certainly put the whole thing inside a `if(NOT CMAKE_CXX_FLAGS)` conditional.

sslattery:
@rfbird what are your thoughts on this? I guess if we document that our `MARCH` setting will set this flag the user should know to not do it themselves

junghans:
And one can always disable it with `-DCOMPILER_SUPPORTS_MARCH=OFF`, too.

sslattery:
Also should we only do this for release builds - i.e. does it do any optimizations that would affect debugging?

junghans:
@rfbird wanted `-march` and `-xHost`, so your call. I am happy with just making _release_ the default.

sslattery:
I'm fine with us setting optimization flags for release builds or the CMake variants of release builds. I guess I would like to see no optimization flags automatically added by the library when a user specifies a pure debug build.

rfbird:
@sslattery  I think you make a good point, without having a good way to disable this we may well hose cross compiling. 

I think 1) lets make release default, 2) lets turn optimization flags off when debug is enabled, and optionally 3) have an option to say "I'm smarter than you don't do 'the right thing'"

junghans:
CMake has `None Debug Release RelWithDebInfo MinSizeRel` for which one do you want to deactivate stuff? And what about user / vendor defined Buildtypes (of which we don't know the names)?

rfbird:
Urgh.

Enable only for release and have release be the default?

junghans:
Done

sslattery:
Thanks @junghans - I think that is a reasonable approach for setting the flags.

PR #41, opened 2018-11-07
Kokkos improvements

junghans:
https://github.com/kokkos/kokkos/issues/1841 got fixed via https://github.com/kokkos/kokkos/pull/1871

### DIFF
--- a/cmake/FindKOKKOS.cmake
+++ b/cmake/FindKOKKOS.cmake
@@ -1,5 +1,5 @@
 # - Find kokkos
-# Creates a PkgConfig::KOKKOS imported target
+# Creates a Kokkos::kokkos imported target
 #
 #  KOKKOS_INCLUDE_DIRS  - where to find kokkos.h, etc.
 #  KOKKOS_LIBRARIES     - List of libraries when using kokkos.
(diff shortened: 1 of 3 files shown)


cmake/FindKOKKOS.cmake
```diff
@@ -1,12 +1,21 @@
...
-# Creates a Kokkos::kokkos imported target
+# Creates a PkgConfig::KOKKOS imported target
 #
 #  KOKKOS_INCLUDE_DIRS  - where to find kokkos.h, etc.
 #  KOKKOS_LIBRARIES     - List of libraries when using kokkos.
 #  KOKKOS_FOUND         - True if kokkos found.
 #  KOKKOS_SETTINGS_DIR - path to kokkos_generated_settings.cmake
 #
 
+find_package(PkgConfig)
```
junghans: @rfbird @sslattery should we require pkg-config and drop all the code below?

junghans:
pkg-config support in kokkos was added in https://github.com/kokkos/kokkos/pull/1840

cmake/FindKOKKOS.cmake
sslattery: Will that work across platforms? I'll admit I've never used it so not sure if it will work on Mac or the HPC machines

junghans: I just tested this on a bunch of machines and it seems the `IMPORTED_TARGET` capability isn't full there in CMake-3.9, so for the sake of not having to bump the minimal CMake version, I will rewrite that a bit.

sslattery:
Does this make any changes to CMake configuration scripts for Cabana? The travis stuff didn't change so I am assuming not.

junghans:
@sslattery in the end, with `IMPORTED_TARGET` being broken in CMake-3.9, I only changed `FindKOKKOS.cmake` to pick up hints from `pkg-config` (if available).

PR #43, opened 2018-11-08
MPI Communication

sslattery:
Adds support for MPI and basic communication plans and operations:

1. `CommunicationPlan` - base class for all communication plans. Contains the base implementation which determines the number of imports and exports as well as neighbor ranks based on export data.

2. `Distributor` - provides a communication plan for `migrate` to move data from one uniquely-owned distribution to another uniquely-owned distribution. This derives from `CommunicationPlan`.

3. `Halo` - provides a communication plan for `gather` and `scatter` to manipulate ghosted data. This also derives from `CommunicationPlan`

Each has a decent set of unit tests that caught some bugs. I have also tested this with a GPU-aware OpenMPI implementation although we are currently only using CUDA UVM (we will be adding regular CUDA memory soon).

### DIFF
--- a/core/src/CMakeLists.txt
+++ b/core/src/CMakeLists.txt
@@ -2,21 +2,57 @@ configure_file(CabanaCore_config.hpp.cmakein CabanaCore_config.hpp)
 
 #-----------------------------------------------------------------------------
 
-file(GLOB HEADERS_PUBLIC *.hpp)
-file(GLOB SOURCES *.cpp)
+set(HEADERS_PUBLIC
+  Cabana_AoSoA.hpp
+  Cabana_Core.hpp
+  Cabana_DeepCopy.hpp
+  Cabana_ExecutionPolicy.hpp
+  Cabana_LinkedCellList.hpp
+  Cabana_Macros.hpp
+  Cabana_MemberTypes.hpp
+  Cabana_NeighborList.hpp
+  Cabana_Parallel.hpp
+  Cabana_Slice.hpp
+  Cabana_SoA.hpp
+  Cabana_Sort.hpp
+  Cabana_Tuple.hpp
+  Cabana_Types.hpp
+  Cabana_VerletList.hpp
+  Cabana_Version.hpp
+  )
+
+#-----------------------------------------------------------------------------
+# MPI-dependent code
+if(Cabana_ENABLE_MPI)
+  list(APPEND HEADERS_PUBLIC
+    Cabana_CommunicationPlan.hpp
+    Cabana_Distributor.hpp
+    Cabana_Halo.hpp
+    )
+endif()
+
+#-----------------------------------------------------------------------------
+# implementation details
+set(HEADERS_IMPL
+  impl/Cabana_CartesianGrid.hpp
+  impl/Cabana_Index.hpp
+  impl/Cabana_IndexSequence.hpp
+  impl/Cabana_PerformanceTraits.hpp
+  impl/Cabana_TypeTraits.hpp
+  )
+set(SOURCES_IMPL
+  impl/Cabana_Version.cpp
+  )
 
 #-----------------------------------------------------------------------------
 
-file(GLOB HEADERS_IMPL impl/*.hpp)
-file(GLOB SOURCES_IMPL impl/*.cpp)
-
-install(FILES ${HEADERS_PUBLIC} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}) 
+install(FILES ${HEADERS_PUBLIC} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 install(FILES ${HEADERS_IMPL} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/impl/)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/CabanaCore_config.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}) 
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/CabanaCore_config.hpp DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
 #-----------------------------------------------------------------------------
 
-add_library(cabanacore ${SOURCES} ${SOURCES} ${SOURCES_IMPL})
+add_library(cabanacore ${SOURCES_IMPL})
 target_include_directories(cabanacore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
 target_link_libraries(cabanacore Kokkos::kokkos)
 if(Cabana_ENABLE_MPI)
--- a/core/src/Cabana_CommunicationPlan.hpp
+++ b/core/src/Cabana_CommunicationPlan.hpp
@@ -1,0 +1,610 @@
+/****************************************************************************
+ * Copyright (c) 2018 by the Cabana authors                                 *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef CABANA_COMMUNICATIONPLAN_HPP
+#define CABANA_COMMUNICATIONPLAN_HPP
+
+#include <Kokkos_Core.hpp>
+
+#include <mpi.h>
+
+#include <type_traits>
+#include <vector>
+#include <exception>
+#include <algorithm>
+#include <numeric>
+#include <cassert>
+
+namespace Cabana
+{
+//---------------------------------------------------------------------------//
+/*!
+  \class CommunicationPlan
+
+  \brief Communication plan base class.
+
+  \tparam MemorySpace Memory space in which the data for this class will be
+  allocated.
+
+  The communication plan computes how to redistribute elements in a parallel
+  data structure using MPI. Given a list of data elements on the local MPI
+  rank and their destination ranks, the communication plan computes which rank
+  each process is sending and receiving from and how many elements we will
+  send and receive. In addition, it provides an export steering vector which
+  describes how to pack the local data to be exported into contiguous send
+  buffers for each destination rank (in the forward communication plan).
+
+  Some nomenclature:
+
+  Export - elements we are sending in the forward communication plan.
+
+  Import - elements we are receiving in the forward communication plan.
+
+  \note If a communication plan does self-sends (i.e. exports and imports data
+  from its own ranks) then this data is first in the data structure. What this
+  means is that neighbor 0 is the local rank and the data for that rank that
+  is being exported will appear first in the steering vector.
+*/
+template<class MemorySpace>
+class CommunicationPlan
+{
+  public:
+
+    // Cabana memory space.
+    using memory_space = MemorySpace;
+
+    // Kokkos execution space.
+    using execution_space = typename memory_space::execution_space;
+
+    /*!
+      \brief Constructor.
+
+      \param comm The MPI communicator over which the distributor is defined.
+
+      \return The MPI communicator for this plan.
+    */
+    CommunicationPlan( MPI_Comm comm )
+        : _comm( comm )
+    {}
+
+    /*!
+      \brief Get the MPI communicator.
+    */
+    MPI_Comm comm() const
+    { return _comm; }
+
+    /*!
+      \brief Get the number of neighbor ranks that this rank will communicate
+      with.
+
+      \return The number of MPI ranks that will exchange data with this rank.
+    */
+    int numNeighbor() const
+    { return _neighbors.size(); }
+
+    /*!
+      \brief Given a local neighbor id get its rank in the MPI communicator.
+
+      \param neighbor The local id of the neighbor to get the rank for.
+
+      \return The MPI rank of the neighbor with the given local id.
+    */
+    int neighborRank( const int neighbor ) const
+    { return _neighbors[neighbor]; }
+
+    /*!
+      \brief Get the number of elements this rank will export to a given neighbor.
+
+      \param neighbor The local id of the neighbor to get the number of
+      exports for.
+
+      \return The number of elements this rank will export to the neighbor with the
+      given local id.
+     */
+    std::size_t numExport( const int neighbor ) const
+    { return _num_export[neighbor]; }
+
+    /*!
+      \brief Get the total number of exports this rank will do.
+
+      \return The total number of elements this rank will export to its
+      neighbors.
+    */
+    std::size_t totalNumExport() const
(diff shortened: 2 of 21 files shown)


sslattery:
@dalg24 @junghans @sjplimp @rfbird This one is finally ready for review. I am not opposed to splitting this PR into 3 PRs - one for each of the 3 new classes. Please take a look when you have a chance. Feedback on the APIs would be useful.

.travis.yml
```diff
@@ -47,6 +47,7 @@ script:
     cmake -DCMAKE_PREFIX_PATH=$HOME/kokkos
           -DCabana_ENABLE_Serial=OFF ${CMAKE_OPTS[@]}
           -DCabana_ENABLE_MPI=ON
+          -DMPIEXEC_EXECUTABLE=/usr/bin/mpirun
```
junghans: Isn't that automatic?

sslattery: Suprisingly it didn't work. On my other machines the CMake MPI module automatically set this variable but it was not being defined in the travis environment. Adding it fixed the problem.

junghans: Hmm, the CMake on travis is very new, so maybe the variable got renamed!

sslattery: That would seem odd as I got my instructions from the latest docs: https://cmake.org/cmake/help/latest/module/FindMPI.html

junghans: I will have a look!

junghans: Got it, Travis run CMake-3.9 and it is called `MPIEXEC` (not `MPIEXEC_EXECUTABLE`) in that version.

junghans: `MPIEXEC` is listed under [Backward-Compatibility](https://cmake.org/cmake/help/v3.11/module/FindMPI.html#backward-compatibility), so we should use that one.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+                    break;
+                }
+
+            // If this is a new neighbor (i.e. someone we didn't send anything
+            // to) record this.
+            if ( !found_neighbor )
+            {
+                _neighbors.push_back( status[i].MPI_SOURCE );
+                _num_import.push_back( import_sizes[i] );
+            }
```
dalg24: How about something like
```C++
auto const source = status[i].MPI_SOURCE;
auto result = std::find(_neighbors.begin(), _neighbors.end(), source);
if (result != std::end(_neighbors))
{
    _num_import[i+self_offset] = import_sizes[i];
}
else
{
    _neighbors.push_back( source );
    _num_import.push_back( import_sizes[i] );
}
```

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+        _num_import.assign( num_n, 0 );
+
+        // Count the number of sends to each neighbor.
+        for ( int n = 0; n < num_n; ++n )
+            _num_export[n] = std::count( element_export_ranks.begin(),
+                                         element_export_ranks.end(),
+                                         _neighbors[n] );
+
+        // Post receives for the number of imports we will get.
+        std::vector<MPI_Request> requests(0);
```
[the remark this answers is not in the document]
sslattery: 👍

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+                MPI_Send( &_num_export[n],
+                          1,
+                          MPI_UNSIGNED_LONG,
+                          _neighbors[n],
+                          1001,
+                          _comm );
+
+        // Wait on receives.
+        std::vector<MPI_Status> status( requests.size() );
+        MPI_Waitall( requests.size(), requests.data(), status.data() );
```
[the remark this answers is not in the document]
sslattery: 👍

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+                                         element_export_ranks.end(),
+                                         _neighbors[n] );
+
+        // Post receives for the number of imports we will get.
+        std::vector<MPI_Request> requests(0);
+        requests.reserve( num_n );
+        for ( int n = 0; n < num_n; ++n )
+            if ( my_rank != _neighbors[n] )
+            {
+                requests.resize( requests.size() + 1 );
```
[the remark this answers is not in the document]
sslattery: 👍

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+        // Count the number of sends this rank will do to other ranks.
+        std::vector<int> num_neighbor( comm_size, 0 );
+        for ( const auto export_rank : element_export_ranks )
+            if ( export_rank >= 0 )
+                ++num_neighbor[ export_rank ];
+
+        // Extract the export ranks and number of exports and then flag the
+        // send ranks.
+        _neighbors.resize( 0 );
+        _num_export.resize( 0 );
```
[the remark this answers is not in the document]
sslattery: 👍

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+        std::vector<std::size_t> offsets( num_n, 0 );
+        for ( int n = 1; n < num_n; ++n )
+            offsets[n] = offsets[n-1] + _num_export[n-1];
+
+        // Create the export steering vector for writing local elements into
+        // the send buffer.
+        std::vector<std::size_t> counts( num_n, 0 );
+        Kokkos::View<std::size_t*,Kokkos::HostSpace>
+            host_steering( "host_steering", _total_num_export );
+        for ( std::size_t i = 0; i < element_export_ranks.size(); ++i )
```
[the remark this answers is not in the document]
sslattery: I made this parallel

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+      \param element_export_ids The local ids of the elements to be
+      exported. This corresponds with the export ranks vector and must be the
+      same length if defined. If use_iota is false this vector must be
+      defined. If use_iota is true, this vector is ignored.
+    */
+    void createExportSteering(
+        const bool use_iota,
+        const std::vector<int>& element_export_ranks,
+        const std::vector<std::size_t>& element_export_ids =
+        std::vector<std::size_t>() )
```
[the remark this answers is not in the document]
sslattery: I refactored these functions to use overloading with different signatures

sslattery:
@dalg24 I refactored the inheritance structure here. The base class now has functions that are essentially protected but they must be made public to use class data with CUDA. I changed the create functions in the derived classes as well. This makes it clear that an entirely new communication plan is being created.

core/src/Cabana_Distributor.hpp
```diff
@@ -0,0 +1,503 @@
...
+
+//---------------------------------------------------------------------------//
+// Synchronously move data between a source and destination AoSoA by executing
+// the forward communication plan.
+template<class Distributor_t, class AoSoA_t>
+void distributeData(
+    const Distributor_t& distributor,
+    const AoSoA_t& src,
+    AoSoA_t& dst,
+    int mpi_tag,
```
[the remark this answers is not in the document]
sslattery: There are default arguments in the functions the user will call - the implementation detail assumes it will be passed a tag by these functions.

core/src/Cabana_Distributor.hpp
```diff
@@ -0,0 +1,503 @@
...
+              int>::type * = 0 )
+{
+    // Check that src and dst are the right size.
+    if ( src.size() != distributor.numExportElement() )
+        throw std::runtime_error("Source is the wrong size for migration!");
+    if ( dst.size() != distributor.totalNumImport() )
+        throw std::runtime_error("Destination is the wrong size for migration!");
+
+    // Get the number of components in the slice. The source and destination
+    // should be the same.
```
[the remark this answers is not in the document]
sslattery: This is explicitly enforced by them having the same type `Slice_t` (i.e. the number of components is built into the type definition).

core/src/Cabana_Halo.hpp
```diff
@@ -0,0 +1,557 @@
...
+          const std::size_t num_local,
+          const IdViewType& element_export_ids,
+          const RankViewType& element_export_ranks,
+          const std::vector<int>& neighbor_ranks )
+        : CommunicationPlan<MemorySpace>( comm )
+    {
+        if ( element_export_ids.size() != element_export_ranks.size() )
+            throw std::runtime_error("Export ids and ranks different sizes!");
+
+        _num_local = num_local;
```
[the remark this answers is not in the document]
sslattery: it is used in `numLocal()` which in turn is used in `scatter` and `gather`. The data is laid out such that the local data comes first in the AoSoA/Slice and the ghosts come second.

core/unit_test/tstCommunicationPlan.hpp
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    template<class RankViewType, class IdViewType>
+    void createSteering(
+        const RankViewType& element_export_ranks,
+        const IdViewType& element_export_ids )
+    {
+        this->createExportSteering(
+            element_export_ranks, element_export_ids );
+    }
```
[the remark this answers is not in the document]
sslattery: yeah I had made this change so that they could be called in the unit test assuming they would be `protected`. Once I got on a GPU machine to test I realized they would need to be public and therefore hadn't changed this back to directly calling the object.

sslattery:
@junghans What do you think is the best way here to enable `CommunicationPlan`, `Distributor`, `Halo`, and their tests/examples only when MPI is enabled?

core/src/Cabana_Distributor.hpp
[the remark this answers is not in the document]
dalg24: Then say "is the same" instead of "should".

core/src/Cabana_DeepCopy.hpp
```diff
@@ -70,7 +70,6 @@ inline void deep_copy(
     // Return if both pointers are null.
     if ( dst_data == nullptr && src_data == nullptr )
     {
-        Kokkos::fence();
```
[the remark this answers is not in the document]
sslattery: I removed this from the PR

core/src/Cabana_Slice.hpp
```diff
@@ -499,8 +499,6 @@ class Slice
 template<typename >
 struct is_slice : public std::false_type {};
 
-// True only if the type is a member slice *AND* the member slice is templated
-// on an AoSoA type.
```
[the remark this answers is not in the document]
sslattery: I removed this from the PR

core/unit_test/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ endforeach()
 ##--------------------------------------------------------------------------##
 ## MPI tests.
 ##--------------------------------------------------------------------------##
-if(Cabana_ENABLE_MPI)
+if(${Cabana_ENABLE_MPI})
```
junghans: That is not needed!

junghans: You can just use `if(Cabana_ENABLE_MPI)`.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+        // the send buffer.
+        std::vector<std::size_t> counts( num_n, 0 );
+        Kokkos::View<std::size_t*,Kokkos::HostSpace>
+            host_steering( "host_steering", _total_num_export );
+        for ( std::size_t i = 0; i < element_export_ranks.size(); ++i )
+            for ( int n = 0; n < num_n; ++n )
+                if ( element_export_ranks[i] == _neighbors[n] )
+                {
+                    host_steering( offsets[n] + counts[n] ) =
+                        (use_iota) ? i : element_export_ids[i];
```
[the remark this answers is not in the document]
sslattery: should we do this now or leave for future optimizations?

core/unit_test/CMakeLists.txt
sslattery: I fixed this

core/src/CMakeLists.txt
```diff
@@ -2,21 +2,57 @@ configure_file(CabanaCore_config.hpp.cmakein CabanaCore_config.hpp)
...
+  Cabana_NeighborList.hpp
+  Cabana_Parallel.hpp
+  Cabana_Slice.hpp
+  Cabana_SoA.hpp
+  Cabana_Sort.hpp
+  Cabana_Tuple.hpp
+  Cabana_Types.hpp
+  Cabana_VerletList.hpp
+  Cabana_Version.hpp
+  )
```
dalg24: Say I create a new header and forget to list it here.  Will it be caught when building and running the tests?

sslattery: It may although that is tough to tell because it is header only so it may just pull it in. It will likely show up more when external libraries link in and the missing header is not installed.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,602 @@
...
+      Whenever the communciation plan is applied, this is the total number of
+      elements expected to be input on the sending ranks (in the forward
+      communication plan). This will be different than the number returned by
+      totalNumExport() if some of the export ranks used in the construction
+      are -1 and therefore will not particpate in an export operation.
+
+      \return The number of export elements.
+    */
+    std::size_t numExportElement() const
+    { return _num_export_element; }
```
dalg24: I understand we need something like this to validate the user input when executing the communication plan but I don't like the name of this member function because I don't find it very descriptive.

sslattery:
Things to finish:

- [ ] Fix name of `exportSteering()` -> `getExportSteering()`

- [ ] Fix name of `numExportElement()` -> `exportSize()`

- [ ] Add MPI_Barrier to all communication functions to avoid potential race conditions for same MPI tag calls and the add `async` versions in the future

core/src/CMakeLists.txt
junghans: No, it won’t, hence globbing and subfolders might be preferred!

sslattery: Lets push this as is with regards to the build system them and move this conversation to #55 - once we come to a conclusion there we can implement in a new PR

core/src/Cabana_CommunicationPlan.hpp
sslattery: I have changed the name as we discussed

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -0,0 +1,460 @@
...
+
+      \return The steering vector for the exports.
+
+      The steering vector places exports in contiguous chunks by destination
+      rank. The chunks are in consecutive order based on the local neighbor id
+      (i.e. all elements going to neighbor with local id 0 first, then all
+      elements going to neighbor with local id 1, etc.). Only indices of ghost
+      elements are in the list of exports.
+    */
+    Kokkos::View<std::size_t*,kokkos_memory_space> exportSteering() const
```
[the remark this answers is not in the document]
sslattery: I have changed the name as we discussed

PR #45, opened 2018-11-15
MPI Dependency

sslattery:
Adds an optional MPI dependency to the build and open-mpi to the Travis CI. This PR has no code that uses MPI - just adds the library and test infrastructure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,12 +64,17 @@ if(Doxygen_FOUND)
   doxygen_add_docs(doxygen core)
 endif()
 
+option(Cabana_ENABLE_MPI "Build Cabana with MPI support" OFF)
+if(Cabana_ENABLE_MPI)
+  find_package(MPI REQUIRED)
+endif()
+
 #------------------------------------------------------------------------------#
 # Architecture
 #------------------------------------------------------------------------------#
 if(CMAKE_BUILD_TYPE STREQUAL "Release")
   set(Cabana_BUILD_MARCH "" CACHE STRING "Arch to use with -march= (if empty CMake will try to use 'native') in release build and only release build")
-  
+
   # Try -march first. On platforms that don't support it, GCC will issue
   # a hard error, so we'll know not to use it.
   if(Cabana_BUILD_MARCH)
@@ -77,7 +82,7 @@ if(CMAKE_BUILD_TYPE STREQUAL "Release")
   else()
     set(INTERNAL_Cabana_BUILD_MARCH "native")
   endif()
-  
+
   include(CheckCXXCompilerFlag)
   check_cxx_compiler_flag("-march=${INTERNAL_Cabana_BUILD_MARCH}" COMPILER_SUPPORTS_MARCH)
   if(COMPILER_SUPPORTS_MARCH)
@@ -87,7 +92,7 @@ if(CMAKE_BUILD_TYPE STREQUAL "Release")
   else()
     unset(INTERNAL_Cabana_BUILD_MARCH)
   endif()
-  
+
   option(Cabana_ENABLE_COVERAGE_BUILD "Do a coverage build" OFF)
   if(Cabana_ENABLE_COVERAGE_BUILD)
       message(STATUS "Enabling coverage build")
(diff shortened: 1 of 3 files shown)


sslattery:
@junghans Gave this an initial try. Looks like its failing to link on the C++ standard library in Travis.

Can you see anything obvious I did wrong? Basically, I added openmpi to the environment and then set the compiler under the MPI wrapper to be whichever we are using (GCC or clang).

junghans:
@sslattery setting `CXX` to `mpicxx` isn't the CMake way, let me see if I can fix it.

CMakeLists.txt
```diff
@@ -64,20 +64,31 @@ if(Doxygen_FOUND)
...
 endif()
 
+option(Cabana_ENABLE_MPI "Build Cabana with MPI support" OFF)
+if( Cabana_ENABLE_MPI )
+  find_package(MPI)
+  if( NOT MPI_FOUND )
+    message(FATAL_ERROR "Cabana MPI support requested but MPI not found." )
+  endif()
+else()
+  set(MPI_CXX__LIBRARIES)
```
dalg24:
```suggestion
  set(MPI_CXX__LIBRARIES)
  set(MPI_CXX_LIBRARIES)
```

core/src/CMakeLists.txt
```diff
@@ -17,6 +17,6 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/CabanaCore_config.hpp DESTINATION ${CM
 #-----------------------------------------------------------------------------
 
 add_library(cabanacore ${SOURCES} ${SOURCES} ${SOURCES_IMPL})
-target_include_directories(cabanacore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-target_link_libraries(cabanacore Kokkos::kokkos)
+target_include_directories(cabanacore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR} ${MPI_CXX_INCLUDE_PATH})
+target_link_libraries(cabanacore Kokkos::kokkos ${MPI_CXX__LIBRARIES})
```
dalg24:
```suggestion
target_link_libraries(cabanacore Kokkos::kokkos ${MPI_CXX_LIBRARIES})
```

junghans:
From which CMake version on is the `MPI::MPI_CXX` Target supported?

junghans:
Good point, then we should use the imported target!

sslattery:
So on Travis it looks like it is still using `g++` or `clang` to do the compiling. Does the way you have linked in the CMake supersede the use of `mpicxx`?

sslattery:
OK just verified with myself that this works by injecting some MPI code into the unit tests to try it out. This is all I wanted to do on this PR. I'll work on package structure and then start on the communication PRs next. If both of you are fine with this please give a review and then we can merge. @junghans @dalg24 thanks for the help.

sslattery:
Also once we merge I will update the [build instructions](https://github.com/ECP-copa/Cabana/wiki/Build-Instructions)

core/src/CMakeLists.txt
```diff
@@ -19,4 +19,8 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/CabanaCore_config.hpp DESTINATION ${CM
 add_library(cabanacore ${SOURCES} ${SOURCES} ${SOURCES_IMPL})
 target_include_directories(cabanacore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
 target_link_libraries(cabanacore Kokkos::kokkos)
+if(Cabana_ENABLE_MPI)
+  find_package(MPI REQUIRED)
```
dalg24: Not sure I'd put the `find_package(MPI)` here.  You might want to do that in the root directory `CMakeLists.txt` so you can register test with `${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ...`

Also do not forget the `if (NOT MPI_CXX_FOUND) message(FATAL_ERROR ...`

junghans: The former is a good point as tests might not live in a subfolder!

The latter, I was thinking `REQUIRED` will check for all enabled languages, incl. `CXX`.

sslattery: @junghans Do you want to make these changes or should I?

junghans: @sslattery I did the former, I don't think the latter is necessary.

sslattery:
@dalg24 do you approve?

core/src/CMakeLists.txt
dalg24: > The latter, I was thinking REQUIRED will check for all enabled languages, incl. CXX.

Oh that's right

PR #49, opened 2018-12-13
Update Parallel For Loops

sslattery:
This PR does 4 key things:

1. Adds an `ExecutionSpace` concept
2. Unifies the definitions of `MemorySpace` and the new `ExecutionSpace` - these just alias Kokkos for now but when we move to version 1.0 I want to just use things in the Kokkos namespace
3. Creates a new `SimdPolicy` and associated `simd_parallel_for` for vectorized for loops
4. Removes `neighbor_parallel_for` from the `Experimental` namespace

The key design takeaway here is that we are only adding things that Kokkos doesn't already do. For example, note that the `neighbor_parallel_for` actually takes a `Kokkos::RangePolicy` as an argument. If a user wants to do a 1d portable loop over a Slice or AoSoA they should use a `Kokkos::RangePolicy` and a `Kokkos::parallel_for`, or any other valid Kokkos approach.


core/src/Cabana_Parallel.hpp
```diff
@@ -143,244 +49,92 @@ class StructAndArrayParallelTag {};
...
-            Impl::getArrayBounds<ExecutionPolicy::array_size>(
-                begin, end, struct_bounds, s );
-            if ( a < array_bounds.second )
-                functor(
-                    Impl::Index<ExecutionPolicy::array_size>::i(s,a) );
-        };
-
-    // Execute the functor.
-    Kokkos::parallel_for( str, k_policy, functor_wrapper );
+            auto s = team.league_rank() + exec_policy.structBegin();
```
guangyechen: I wonder if use `kokkos_policy team_policy( 1, exec_policy.numStruct(), VectorLength );` instead of `kokkos_policy team_policy( exec_policy.numStruct(), 1, VectorLength );`, we can just use `auto s = tream.team_rank ();`?

core/src/Cabana_ExecutionPolicy.hpp
```diff
@@ -13,53 +13,219 @@
...
-    // Range constructor.
+    /*!
+      \brief Range constructor.
+      \param begin The begininning of the 1D range.
+      \param begin The ending of the 1D range.
+    */
     RangePolicy( const std::size_t begin, const std::size_t end )
-        : _begin( begin )
-        , _end( end )
+        : base_type(begin,end)
```
dalg24: Any reason why you wouldn't just do
```C++
using Kokkos::RangePolicy::RangePolicy;
```
Are they constructors from the base you don't want to make visible?

core/src/Cabana_ExecutionPolicy.hpp
```diff
@@ -13,53 +13,219 @@
...
+
+      \param container The container over which to build the range policy.
+    */
     template<class Container>
     RangePolicy( Container container )
-        : _begin( 0 )
-        , _end( container.size() )
+        : base_type(0,container.size())
+    {}
+};
```
dalg24: I think you should get rid of this class altogether.  I don't see how neither the constructor from the extent of the range nor the one from a generic container bring enough added value to justify the existence of that class.

core/src/Cabana_ExecutionPolicy.hpp
```diff
@@ -13,53 +13,219 @@
...
+template<int VectorLength>
+class StructRange
+{
+  public:
+
+    template<typename I>
+    KOKKOS_FORCEINLINE_FUNCTION
+    static constexpr
+    typename std::enable_if<std::is_integral<I>::value,std::size_t>::type
+    structBegin( const I& begin )
```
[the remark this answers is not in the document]
dalg24: Also really don't see the point of templating all these member functions.  It introduces unnecessary code bloat.  Just commit to an integral type, `std::size_t` is probably fine, otherwise if you really want a template, template the entire class on `IndexType` and instantiate it on `ExecutionSpace::index_type` or deduce it from the Kokkos policies using whatever traits they come with.

core/src/Cabana_ExecutionPolicy.hpp
```diff
@@ -13,53 +13,219 @@
...
+
+  \tparam VectorLength The inner array size of the AoSoA.
+*/
+template<int VectorLength>
+class StructRange
+{
+  public:
+
+    template<typename I>
+    KOKKOS_FORCEINLINE_FUNCTION
```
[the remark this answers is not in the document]
sslattery: changed this to a standard inline function

core/src/Cabana_ExecutionPolicy.hpp
sslattery: I think the begin/end constructor is the only one I have that mirrors a base constructor. The single argument version may also be in `Kokkos::RangePolicy` although the container one is not.

core/src/Cabana_ExecutionPolicy.hpp
sslattery: For now I do like the idea of having two different objects to represent our 2 distinct parallel loop strategies over the AoSoA data structure. I know that one of them is exactly a `Kokkos::RangePolicy` (and I have renamed this to `Cabana::LinearPolicy`) but I like having a single basic class in which we can encapsulate our ideas for documentation purposes.

I suppose we could just bring `Kokkos::RangePolicy` into the Cabana namespace but then how would we document this as part of the API/Design of the library?

sslattery:
Other than the pending discussion on `RangePolicy`/`LinearPolicy` are there any more comments here @guangyechen @rfbird? I'd like to merge this as soon as possible so I can do a PR cleaning up the neighbor list parallel for which will depend on this PR.

core/src/Cabana_ExecutionPolicy.hpp
dalg24: > The single argument version may also be in `Kokkos::RangePolicy`

It is not.

One thing I don't like about it is it goes against the behavior one would expect with C++ - that is, default arguments are for trailing parameters.  And it does not save you many keystrokes `RangePolicy(n)` vs `RangPolicy(0, n)`,

core/src/Cabana_ExecutionPolicy.hpp
dalg24: By just calling it what it is and instead focus on spelling out what Cabana brings that's not already in Kokkos.

core/unit_test/tstParallel.hpp
```diff
@@ -157,59 +177,149 @@ void runTest()
...
+    using AoSoA_t = Cabana::AoSoA<DataTypes,TEST_MEMSPACE>;
+
+    // Create an AoSoA.
+    int num_data = 155;
+    AoSoA_t aosoa( num_data );
+
+    // Create a vectorized execution policy using the begin and end of the
+    // AoSoA.
+    Cabana::SimdPolicy<TEST_EXECSPACE,AoSoA_t::vector_length>
+        policy_1( 12, 135 );
```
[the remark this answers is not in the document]
sslattery: done

sslattery:
Yes the peel/remainder is hiding in the `SimdPolicy`. You give Kokkos arguments in the `ThreadVectorRange` for begin and and which will either be `0` to `VectorLength` or the peel/remainder loop. These are always runtime options for the `ThreadVectorRange`

core/src/Cabana_Parallel.hpp
sslattery: have discussed this. For now what we have is functional and should get reasonable performance. We can do these explorations in future performance studies and it will not affect the API

core/src/Cabana_ExecutionPolicy.hpp
sslattery: OK I have removed range policy - people should use a `Kokkos::RangePolicy` with a `Kokkos::parallel_for` if they want to do linear loops over the data structure.

core/src/Cabana_ExecutionPolicy.hpp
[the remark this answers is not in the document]
sslattery: Removed templates.

PR #51, opened 2018-12-16
Remove pointer and stride arrays from AoSoA

sslattery:
There were some fixed-size arrays in the AoSoA that were not needed anymore. I could see these causing issues in default copy construction at some point anyway.


sslattery:
@rfbird can you look at this one and give and approve if you're ok with it? Should be very quick to assess.

PR #52, opened 2018-12-18
Performance tests

rfbird:
This PR attempts to capture the progress that has been made on the peakflops benchmark.

I am to layout a logical progression of performance tests that demonstrate how the peakflops performance varies with the complexity of the implementation. 

Currently it is split into 3 major parts (each which can have subparts):

1) Vanilla C++ : this is the baseline performance for which we can compare everything else
2) Kokkos only : this demonstrates that the abstraction to Kokkos does [should?] not cost us any performance
3) Cabana : demonstrates the variation in performance for the different interfaces 

These can then be moved to formal performance tests (using gtest etc), where #1 has to get a minimum performance of ~32 (based on hardware), and the subsequent tests are considered a fail if they do not achieve some fraction (90%?) of that baseline.

Once we have dedicated hardware, this will make for a nice way to track performance regressions against the API and underlying implementations


guangyechen:
It may be worth pointing out that the current version of Kokkos does not support MemoryTratis<Kokkos::Restrict> out of box. As a result, it does not reach peakflops (e.g. 32 flop/cycle on haswell). This maybe resolved in a future version of Kokkos.

core/example/benchmark/CMakeLists.txt
```diff
@@ -1,7 +1,15 @@
+option(Cabana_ENABLE_EXAMPLES "Build examples" OFF)
```
junghans: You should define the option in main CMake, it is double defined below!

rfbird: Thanks, that was a copy paste error I'll take it out!

guangyechen:
#39 just passed the peakflops test for `slice.access()`! The key there is using compile-time stride for addressing. We should see the peakflops after #39 is merged.

junghans:
Can you exclude the performance test from coverage counting, too? (See `.codecov.yml` in the repository root directory)

sslattery:
@rfbird what is the status of this? This one has been around for a long time.

rfbird:
Honestly, I thought we merged this. I think something like this should go in. There's a lot more we could do, but I think as-is this has value

sslattery:
If you think this is ready for review can you take out of WIP?

junghans:
@rfbird this needs a merge!

core/example/benchmark/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
+
 if(Cabana_ENABLE_Serial)
   add_executable(NeighborListMDPerfTest md_neighbor_perf_test.cpp)
   target_link_libraries(NeighborListMDPerfTest cabanacore)
 endif()
 
-add_executable(PeakFlops Cabana_peakflops.cpp)
-target_link_libraries(PeakFlops cabanacore)
+#add_executable(PeakFlops Cabana_peakflops.cpp)
```
sslattery: If this is in the new performance test directory let's just get rid of these lines.

core/example/benchmark/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
+
 if(Cabana_ENABLE_Serial)
   add_executable(NeighborListMDPerfTest md_neighbor_perf_test.cpp)
```
sslattery: In the future I would like this to be in a performance test as well. Basically, everything in `example/benchmark/` should be a performance test eventually.

core/performance_test/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+option(Cabana_ENABLE_PERFORMANCE_TESTING "Build Performance Tests" OFF)
+
+if(NOT Cabana_ENABLE_PERFORMANCE_TESTING)
+    message("Not doing Performance Tests")
+    return()
+endif()
```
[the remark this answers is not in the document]
sslattery: agreed

core/performance_test/peakflops/01_cpp_simple.cpp
```diff
@@ -0,0 +1,206 @@
...
+x_[6] = 14.501436
+x_[7] = 1.216459
+-------------------
+For Sandy/Ivy Bridge you need to unroll by 3:
+
+Only FP Add has dependency on the previous iteration of the loop
+FP Add can issue every cycle
+FP Add takes three cycles to complete
+Thus unrolling by 3/1 = 3 completely hides the latency
+FP Mul and FP Load do not have a dependency on the previous iteration and you can rely on the OoO core to issue them in the near-optimal order. These instructions could affect the unroll factor only if they lowered the throughput of FP Add (not the case here, FP Load + FP Add + FP Mul can issue every cycle).
```
sslattery: So you do three "loops" within one?

sslattery: Do you do this unrolling in this file?

core/performance_test/peakflops/01_cpp_simple.cpp
```diff
@@ -0,0 +1,206 @@
...
+  float vec[VECLENTH];
+};
+
+//struct data * axpy_10(struct data *restrict a, struct data *restrict x0, struct data *restrict x1, struct data *restrict x2, struct data *restrict x3, struct data *restrict x4, struct data *restrict x5, struct data *restrict x6, struct data *restrict x7, struct data *restrict x8, struct data *restrict x9,struct data *restrict c, long n) {
+  struct data * axpy_10(struct data *__restrict__ a, struct data *__restrict__ x0, struct data *__restrict__ x1, struct data *__restrict__ x2, struct data *__restrict__ x3, struct data *__restrict__ x4, struct data *__restrict__ x5, struct data *__restrict__ x6, struct data *__restrict__ x7, struct data *__restrict__ x8, struct data *__restrict__ x9,struct data *__restrict__ c, long n) {
+  long i;
+  int j;
+
+  asm volatile ("# ax+c loop begin");
+  for(i = 0; i<n; i++){
```
sslattery: Is this where the unroll happens?

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <gtest/gtest.h>
+#include "common.h"
+//#include "rdtsc.h"
+
+typedef Kokkos::View<float[VECLENTH],Kokkos::MemoryTraits<Kokkos::Restrict> > view_type;
+
+/*
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+*/
+
+
+struct data{
+    float vec[VECLENTH];
+};
+
+//struct data * axpy_10(struct data *restrict a, struct data *restrict x0, struct data *restrict x1, struct data *restrict x2, struct data *restrict x3, struct data *restrict x4, struct data *restrict x5, struct data *restrict x6, struct data *restrict x7, struct data *restrict x8, struct data *restrict x9,struct data *restrict c, long n) {
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+//struct data * axpy_10(struct data *__restrict__ a, struct data *__restrict__ x0, struct data *__restrict__ x1, struct data *__restrict__ x2, struct data *__restrict__ x3, struct data *__restrict__ x4, struct data *__restrict__ x5, struct data *__restrict__ x6, struct data *__restrict__ x7, struct data *__restrict__ x8, struct data *__restrict__ x9,struct data *__restrict__ c, long n) {
+
+view_type //__attribute__ ((noinline))
+axpy_10(
+        const view_type & __restrict__  a,
```
sslattery: Does using `__restrict__` here matter when they type is a `Kokkos::View`?

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+    view_type x3_("x3",N);
+    view_type x4_("x4",N);
+    view_type x5_("x5",N);
+    view_type x6_("x6",N);
+    view_type x7_("x7",N);
+    view_type x8_("x8",N);
+    view_type x9_("x9",N);
+    view_type c_("c",N);
+
+    /*
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+    for(j=0; j<VECLENTH; j++){
+        x0(j) = x0(j)+x1(j)+x2(j)+x3(j)+x4(j)+x5(j)+x6(j)+x7(j)+x8(j)+x9(j)+(float)n;
+    }
+    return x0;
+}
+
+
+TEST(kokkos, simple) {
+    long n = static_cast<long>(2e6); //1000000000; //MAXBYTES/sizeof(float);
+    /*long seed = (argc > 2 ? atol(argv[2]) : 76843802738543);*/
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+        x2_(i) = erand48(rg);
+        x3_(i) = erand48(rg);
+        x4_(i) = erand48(rg);
+        x5_(i) = erand48(rg);
+        x6_(i) = erand48(rg);
+        x7_(i) = erand48(rg);
+        x8_(i) = erand48(rg);
+        x9_(i) = erand48(rg);
+    }
+    // for (i = 0; i < VECLENTH; i++) {
```
sslattery: Is this for debug purposes? If so you may want to place in some kind of debug print macro.

core/performance_test/peakflops/02_kokkos_simple_view.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    if ( flops_clock > expected_flops_clock * ERROR_MARGIN )
+    {
+        acceptable_fraction = true;
+    }
+
+    EXPECT_TRUE(acceptable_fraction);
+
+    /*
+    delete a_;
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -62,7 +55,9 @@ struct data_t
 // Helper functions.
 //---------------------------------------------------------------------------//
 // Move function using the array-of-struct-of-arrays synatx.
-void movePx(data_t *__restrict__ a,  data_t *__restrict__ x0,
+
+void //__attribute__ ((noinline))
```
sslattery: Is the attribute needed here?

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -146,8 +202,54 @@ void moveSlices(SliceType a,  SliceType x0, SliceType x1, SliceType x2,
         }
     }
 }
+*/
 
-//---------------------------------------------------------------------------//
+/*
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -146,8 +202,54 @@ void moveSlices(SliceType a,  SliceType x0, SliceType x1, SliceType x2,
...
+
+#pragma omp declare simd simdlen(VECLENTH) uniform(s) linear(i) uniform(a) uniform(c) uniform(x)
+template<typename SliceType>
+//void access_vector_mul(SliceType* x, SliceType* a, SliceType* c, const int s, const int i)
+float access_vector_mul(SliceType& x, SliceType& a, SliceType& c, int s, int i)
+{
+    return a.access(s,i) * x.access(s,i) + c.access(s,i);
+}
+/*
+void access_vector_variant() {
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -159,13 +261,69 @@ void moveSlicesWithAccess(SliceType a,  SliceType x0, SliceType x1, SliceType x2
     int s;
     int j;
 
+    /*
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -159,13 +261,69 @@ void moveSlicesWithAccess(SliceType a,  SliceType x0, SliceType x1, SliceType x2
...
-    for(i = 0; i<n; i++)
+    for ( s = 0; s < num_struct; ++s )
     {
-        for ( s = 0; s < num_struct; ++s )
+        for(i = 0; i<n; i++)
         {
+#pragma omp simd
             for(j=0; j<VECLENTH; j++)
             {
+                //(&x0.access(0,0))[s*x0_stride + j] = (&a.access(0,0))[s*a_stride + j] * (&x0.access(0,0))[s*x0_stride + j] + (&c.access(0,0))[s*c_stride + j]; // Works
```
sslattery: Is this code going to be used? If not remove it.

core/performance_test/CMakeLists.txt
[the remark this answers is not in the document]
rfbird: fixed

core/example/benchmark/CMakeLists.txt
rfbird: fixed

core/performance_test/peakflops/01_cpp_simple.cpp
rfbird: Honestly, not sure. This is a note from @guangyechen . But I agree I don't see an unroll by 3? I assume it's hidden in the `n`

core/performance_test/peakflops/02_kokkos_simple_view.cpp
rfbird: fixed

core/performance_test/peakflops/02_kokkos_simple_view.cpp
rfbird: fixed

core/performance_test/peakflops/02_kokkos_simple_view.cpp
rfbird: fixed

core/example/CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-option(Cabana_ENABLE_EXAMPLES "Build examples" OFF)
```
junghans: Why?

core/performance_test/peakflops/02_kokkos_simple_view.cpp
rfbird: fixed

core/performance_test/peakflops/02_kokkos_simple_view.cpp
rfbird: fixed

core/example/benchmark/CMakeLists.txt
```diff
@@ -2,9 +2,3 @@ if(Cabana_ENABLE_Serial)
...
   target_link_libraries(NeighborListMDPerfTest cabanacore)
 endif()
-
-# PeakFlops uses x86 asm volatile
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "([xX]86|[aA][mM][dD]64)")
```
junghans: Does it work on Power9 now?

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: This is a tricky one. GCC generates better assembly with the attribute, but in an ideal world you'd never tell it not to inline... 

(gcc bug?)

core/performance_test/peakflops/02_kokkos_simple_view.cpp
rfbird: It certainly doesn't hurt (and, I think, has different semantics). I'm not sure if it's performance critical though

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: hmmmm - so I see it is commented out which is why I originally was wondering if we needed it. I would say either remove if not needed. If needed, but not portable, wrap in some type of macro.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -13,16 +13,9 @@
 
 #include <cstdlib>
```
sslattery: How is this file related to what is in the new `performance_test` directory? I think `peakflops` should only appear in a single location.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -105,49 +100,72 @@ void movePx(data_t *__restrict__ a,  data_t *__restrict__ x0,
...
     }
+    */
 }
 
-//---------------------------------------------------------------------------//
+
+
+#pragma omp declare simd simdlen(VECLENTH) uniform(s) linear(i) uniform(a)
+template<typename SliceType>
+float access_vector_mul_direct(SliceType& x, SliceType& a, SliceType& c, int s, int i)
```
sslattery: This function does not seem to get used.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -105,49 +100,72 @@ void movePx(data_t *__restrict__ a,  data_t *__restrict__ x0,
...
+{
+    return a(i) * x(i) + c(i);
+}
+
+
+#pragma omp declare simd simdlen(VECLENTH) uniform(s) linear(i) uniform(a) uniform(c) uniform(x)
+template<typename SliceType>
+float access_vector_mul(SliceType& x, SliceType& a, SliceType& c, int s, int i)
```
sslattery: This function does not seem to get used.

core/example/benchmark/CMakeLists.txt
rfbird: Nope. I moved that check down into `core/performance_test/peakflops/CMakeLists.txt `, but maybe it should be higher? 

My thinking was "only peakflops needs this check, so it should be as close to that as possible" (i.e some performance tests might work on p9)

core/example/CMakeLists.txt
rfbird: Because I'm a bad person. I think this got accidently smushed in the merge. reverted

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: Seems reasonable. I don't really want to call it `Cabana.cpp` either :P. suggestions?

core/performance_test/peakflops/01_cpp_simple.cpp
guangyechen: Those are some old note taken from a few years ago. The number of "unrolling" is very cpu dependent, and the note may need be cleaned up.

guangyechen: @sslattery To answer your question: yes, here we do unrolling of 10 for haswell cpus.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: Well I guess my point is why duplicate it? If we have peak flops as a performance test and it carries all the functionality of what is in `example/benchmark` then we should just delete what is in `example/benchmark`.

core/performance_test/peakflops/01_cpp_simple.cpp
rfbird: Updated comments in file

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: removed

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: removed

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: Oh, my misunderstanding. Honestly, I thought I did:

https://github.com/rfbird/Cabana/tree/performance_tests/core/example/benchmark

Maybe I messed up the merge, so it's "not there" rather than "deleted"?

sslattery: Could be - one of the reasons I was curious how clean the merge/rebase was.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -105,49 +103,55 @@ void movePx(data_t *__restrict__ a,  data_t *__restrict__ x0,
...
+                    x6.access(s).get<0>(j) = _a * x6.access(s).get<0>(j) + _c;
+                    x7.access(s).get<0>(j) = _a * x7.access(s).get<0>(j) + _c;
+                    x8.access(s).get<0>(j) = _a * x8.access(s).get<0>(j) + _c;
+                    x9.access(s).get<0>(j) = _a * x9.access(s).get<0>(j) + _c;
+                }
             }
-        }
     }
     asm volatile ("# ax+c loop end");
+    /* // TODO: do we need to re-enable this?
```
sslattery: What is this loop here for?

rfbird: @guangyechen we do want this, right?

guangyechen: The motivation was to use the results of above loops so that compiler do not optimize them away.

rfbird: The `asm volatile` also does that, right?

guangyechen: I don't know. If it does, then we can remove the 2nd loop.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
guangyechen: @rfbird I guess the simplest thing to do is to remove it for now.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: have you guys tried this out?

guangyechen: Either way (removing or keeping it here) is fine here, because those `x0`-`x9` are used in the next call `moveSlicesWithAccess`, and the compiler will not optimize it away.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: Is this going to be removed or uncommented?

CMakeLists.txt
```diff
@@ -55,11 +55,13 @@ if(NOT ENABLE_DEVICES)
 endif()
 message(STATUS "Enable Devices: ${ENABLE_DEVICES}")
 
+
```
sslattery: Remove these extra lines in this file.

core/performance_test/peakflops/01_cpp_simple.cpp
```diff
@@ -0,0 +1,202 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "common.h"
+/*#include "rdtsc.h"*/
```
sslattery: is this header needed?

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: @guangyechen @rfbird lets make a call here and resolve this one

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: @guangyechen @rfbird lets make a call here and resolve this one. If it stays please document in the comments why it is there.

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -299,10 +304,14 @@ void run()
     std::cout<<dc1<<" clocks 1"<<std::endl;
     std::cout<<flops/dc1<<" flops/clock 1\n";
     std::cout << std::endl;
-    std::cout << "Slice Single Index" << std::endl;
+
+    double flops_clock = flops / dc1;
```
sslattery: What about checking the performance of the other 2 ways of accessing the data?

sslattery:
also please rebase to resolve conflicts

core/performance_test/peakflops/01_cpp_simple.cpp
rfbird: fixed

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: re-enabling so it's consistent with the other files

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: Do we want to add that?

I had originally left it out because:

1. It doesn't test any Cabana functionality
2. If they don't work, it's basically impossible for the existing test to pass (so it's somewhat subsumed)

I'm happy to add it though!

rfbird:
> also please rebase to resolve conflicts

WRT the rebase, I recall that I already merged in master during the last set of changes, so a rebase probably no longer makes sense?

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: I mad the call to remove the comment attribute, and leave the `// NOTE` above

core/performance_test/peakflops/03_Cabana_peakflops.cpp
sslattery: I would argue we don't test any Cabana functionality in the benchmark if we don't check the other 2. Shouldn't we expect peak flops when using slice with `access`?

rfbird: Having thought about it a bit more, I agree (and in fact my comment above wasn't even strictly true). I'll add it now -- good catch.

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# PeakFlops uses x86 asm volatile
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "([xX]86|[aA][mM][dD]64)")
+
+    set(GTEST_SOURCE_DIR ${CMAKE_SOURCE_DIR}/gtest)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGTEST_HAS_PTHREAD=0")
```
junghans:
```suggestion
    target_compile_definitions(${TEST_NAME} "DGTEST_HAS_PTHREAD=0")
```

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+    include_directories(${GTEST_SOURCE_DIR})
+
+    include_directories(${CMAKE_CURRENT_BINARY_DIR})
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+    foreach(_test 01_cpp_simple 02_kokkos_simple_view 03_Cabana_peakflops)
+        # TODO: tidy unit_test_main.cpp path
```
junghans: Could you use `${GTEST_SOURCE_DIR}` for that?

core/performance_test/peakflops/common.h
```diff
@@ -3,6 +3,7 @@
 
 #include <gtest/gtest.h>
 
+// NOTE: The user may have to override these for a given platform
```
sslattery: Should we add a configure option for this?

guangyechen:
This pr is related to issue "Performance regression benchmarks #4".

core/performance_test/peakflops/CMakeLists.txt
rfbird: Do you mean

`foreach(_test ${GTEST_SOURCE_DIR})`? If so, yes we could, be I don't really love the globbing over a simple small list

core/performance_test/peakflops/common.h
rfbird: Right now that's why they're wrapped in `#ifndef` so the user can specify `-DEXPECTED_FLOPS` etc. Are you suggesting exposing an explicit cmake `option()`? I'm fine either way, as it's a pretty "advanced" option that most users don't care about 

Eventually it might be nice to try and auto detect it...

sslattery: good point - you can just pass this in at compile time with `-D` through cmake flags

sslattery:
CI hanging due to memory constraints on testing system - will address tomorrow

sslattery:
retest this please

sslattery:
retest this please

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+set(GTEST_SOURCE_DIR ${CMAKE_SOURCE_DIR}/gtest)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGTEST_HAS_PTHREAD=0")
```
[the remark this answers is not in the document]
rfbird: Thanks for catching this. Is it even needed? AFAIK this is here for historic reasons, but it seems like the rest of Cabana no longer has it?

sslattery:
retest this please

junghans:
@sslattery I think we need to add `Cabana_ENABLE_Performance_Testing=ON` to Jenkins, too.

sslattery:
Yes we should add that but we will need to update Jenkins in a different PR - @dalg24 do you concur?

junghans:
I just saw there is `Cabana_ENABLE_PERFORMANCE_TESTING` and `Cabana_ENABLE_Performance_Testing` which both need to be enabled.

sslattery:
Do we want/need both? Looks like there was some duplication here.

rfbird:
> Do we want/need both? Looks like there was some duplication here.

Agreed. Let me get it down to just one -- any preferences which?

sslattery:
The pattern so far has been to use all uppercase (e.g. `Cabana_ENABLE_TESTING`, `Cabana_ENABLE_EXAMPLES`) so I would go with `Cabana_ENABLE_PERFORMANCE_TESTING`

junghans:
Now travis fails with:
```
[ 59%] Building CXX object core/performance_test/peakflops/CMakeFiles/02_kokkos_simple_view_test.dir/02_kokkos_simple_view.cpp.o
cd /home/travis/build/ECP-copa/Cabana/build/core/performance_test/peakflops && /usr/lib/ccache/g++   -I/home/travis/build/ECP-copa/Cabana/gtest -I/home/travis/build/ECP-copa/Cabana/build/core/performance_test/peakflops -I/home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops -I/home/travis/build/ECP-copa/Cabana/core/src -I/home/travis/build/ECP-copa/Cabana/build/core/src -isystem /home/travis/kokkos/include -isystem /usr/lib/openmpi/include -isystem /usr/lib/openmpi/include/openmpi  -Wall -Wextra -pedantic -Werror -DGTEST_HAS_PTHREAD=0   --std=c++11 -fopenmp -o CMakeFiles/02_kokkos_simple_view_test.dir/02_kokkos_simple_view.cpp.o -c /home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops/02_kokkos_simple_view.cpp
/home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops/01_cpp_simple.cpp:97:0: error: ignoring #pragma omp simd [-Werror=unknown-pragmas]
 #pragma omp simd
 ^
```
so we need to add a check to CMake for a minimum GCC version (and add a newer gcc to Travis).

rfbird:
> Now travis fails with:
> 
> ```
> [ 59%] Building CXX object core/performance_test/peakflops/CMakeFiles/02_kokkos_simple_view_test.dir/02_kokkos_simple_view.cpp.o
> cd /home/travis/build/ECP-copa/Cabana/build/core/performance_test/peakflops && /usr/lib/ccache/g++   -I/home/travis/build/ECP-copa/Cabana/gtest -I/home/travis/build/ECP-copa/Cabana/build/core/performance_test/peakflops -I/home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops -I/home/travis/build/ECP-copa/Cabana/core/src -I/home/travis/build/ECP-copa/Cabana/build/core/src -isystem /home/travis/kokkos/include -isystem /usr/lib/openmpi/include -isystem /usr/lib/openmpi/include/openmpi  -Wall -Wextra -pedantic -Werror -DGTEST_HAS_PTHREAD=0   --std=c++11 -fopenmp -o CMakeFiles/02_kokkos_simple_view_test.dir/02_kokkos_simple_view.cpp.o -c /home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops/02_kokkos_simple_view.cpp
> /home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops/01_cpp_simple.cpp:97:0: error: ignoring #pragma omp simd [-Werror=unknown-pragmas]
>  #pragma omp simd
>  ^
> ```
> 
> so we need to add a check to CMake for a minimum GCC version (and add a newer gcc to Travis).

Agreed. My definition of `VLEN` has also gone, so I'll fix that up too and we are hopefully good

core/performance_test/peakflops/CMakeLists.txt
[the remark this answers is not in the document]
junghans: @sslattery do you know?

.travis.yml
```diff
@@ -85,6 +85,7 @@ script:
           -DCabana_ENABLE_Serial=OFF ${CMAKE_OPTS[@]}
           -DCabana_ENABLE_MPI=ON
           -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_EXAMPLES=ON
+          -DCabana_ENABLE_Performance_Testing=ON -DCabana_ENABLE_PERFORMANCE_TESTING=ON
```
junghans: Remove on the variables!

core/performance_test/peakflops/CMakeLists.txt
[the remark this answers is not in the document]
sslattery: no...

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+# we only want to run the below if we have access to "#pragma omp simd"
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles(
+      "int main(void) {
+      #pragma omp simd
+      return 0;
+      }" HAVE_OMP_SIMD)
+
+if (NOT HAVE_OMP_SIMD)
+    # No point running these tests without omp simd
```
junghans: I would a `message(WARNING "...")` here

junghans:
Now some tests fail with clang:
```
he following tests FAILED:
	 27 - 01_cpp_simple_performance_test (Failed)
	 28 - 02_kokkos_simple_view_performance_test (Failed)
	 29 - 03_Cabana_peakflops_performance_test (Failed)
```

rfbird:
@junghans , I talked it over with Stuart, and lets do the following: 

1) Only run the performance tests for a single configuration: GCC with aggressive optimization enabled. 
2) Represent that as a separate travis stage (called "performance"), so it's super obvious if we break the performance but not the functionality. Overtime we can add more tests to the performance stage

What do you think?

junghans:
Sounds good to me!

junghans:
I fixed Travis.

sslattery:
retest this please

core/example/benchmark/CMakeLists.txt
```diff
@@ -3,14 +3,8 @@ if(Cabana_ENABLE_Serial)
   target_link_libraries(NeighborListMDPerfTest cabanacore)
 endif()
 
-# PeakFlops uses x86 asm volatile
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "([xX]86|[aA][mM][dD]64)")
-  add_executable(PeakFlops Cabana_peakflops.cpp)
-  target_link_libraries(PeakFlops cabanacore)
-endif()
-
-if(Cabana_ENABLE_MPI AND Cabana_ENABLE_Cuda AND Cabana_ENABLE_OpenMP)
```
sslattery: This logic was the correct one - looks like this got changed in the rebase

rfbird: Do you mean we want:
`if(Cabana_ENABLE_MPI AND Cabana_ENABLE_Cuda AND Cabana_ENABLE_OpenMP)`
over 
`if(Cabana_ENABLE_MPI AND Cabana_ENABLE_Cuda AND (Cabana_ENABLE_Serial OR Cabana_ENABLE_OpenMP OR Cabana_ENABLE_Pthread) )`
or something else?

sslattery: Yes we want `if(Cabana_ENABLE_MPI AND Cabana_ENABLE_Cuda AND Cabana_ENABLE_OpenMP)`

rfbird: fixed

core/performance_test/peakflops/CMakeLists.txt
junghans: Please no silent skipping.....

rfbird: Good catch. fixed

rfbird: fixed

core/performance_test/peakflops/CMakeLists.txt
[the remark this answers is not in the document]
junghans: move to `add_definitions()`.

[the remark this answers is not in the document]
junghans: `target_compile_definitions()` is even better.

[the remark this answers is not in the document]
rfbird: Attempting to remove

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    message(WARNING "Compiler doesn't seem to support `#pragma simd`, skipping performance tests")
+    return()
+endif()
+
+set(GTEST_SOURCE_DIR ${CMAKE_SOURCE_DIR}/gtest)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
junghans: Huh, now we should remove that line!

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+include_directories(${GTEST_SOURCE_DIR})
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+set(gtest_args --gtest_color=yes)
```
junghans: Is this used anywhere?

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    # No point running these tests without omp simd
+    message(WARNING "Compiler doesn't seem to support `#pragma simd`, skipping performance tests")
+    return()
+endif()
+
+set(GTEST_SOURCE_DIR ${CMAKE_SOURCE_DIR}/gtest)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+include_directories(${GTEST_SOURCE_DIR})
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
```
junghans: I don't think this is needed.

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+# Copied from unit test, could be DRYd
```
junghans: Remove this comment...

core/performance_test/peakflops/CMakeLists.txt
rfbird: haha. dur. I'll fix

core/performance_test/peakflops/03_Cabana_peakflops.cpp
```diff
@@ -295,29 +316,39 @@ void run()
...
+    EXPECT_TRUE( check_expected_flops( flops / dc3 ) );
+
 }
 
 
+
 //---------------------------------------------------------------------------//
 // Main.
 //---------------------------------------------------------------------------//
+/*
```
junghans: Remove the commented out code.

core/performance_test/peakflops/CMakeLists.txt
rfbird: removed

core/performance_test/peakflops/CMakeLists.txt
rfbird: ill try removing (just `CMAKE_CURRENT_BINARY_DIR` right?)

core/performance_test/peakflops/03_Cabana_peakflops.cpp
rfbird: done

core/performance_test/peakflops/CMakeLists.txt
rfbird: done

core/performance_test/peakflops/CMakeLists.txt
junghans: Yes.

junghans:
@rfbird I fixed a typo in the Travis file and now the tests actually get run and .... fail!

junghans:
Let's see if that fixed it.

junghans:
Now, it compiles with `-O3`, but it still fails.

rfbird:
> Now, it compiles with `-O3`, but it still fails.

```
/home/travis/build/ECP-copa/Cabana/core/performance_test/peakflops/01_cpp_simple.cpp:97:0: warning: ignoring #pragma omp simd [-Wunknown-pragmas]
 #pragma omp simd
```

It doesn't actually apply the optimization. I guess we need `-fopenmp` or something to enable it? it's interesting that the try compile didn't catch it, is there a way to add `-wErrror` to the try compile?

junghans:
I added `-Werror`. Do we need a newer gcc for the `omp simd` to work.

rfbird:
> I added `-Werror`. Do we need a newer gcc for the `omp simd` to work.

We need ` -fopenmp-simd` (or `fopenmp`), I'll add it

rfbird:
> I added `-Werror`. Do we need a newer gcc for the `omp simd` to work.

I meant `-Werror` for the `try_compile` block that checks if the compiler supports `#pragma omp simd`

junghans:
@rfbird it still says:
```
-- Performing Test HAVE_OMP_SIMD - Failed
CMake Warning at core/performance_test/peakflops/CMakeLists.txt:18 (message):
  Compiler doesn't seem to support `#pragma simd`, skipping performance tests
```
Ideas?

rfbird:
> @rfbird it still says:
> 
> ```
> -- Performing Test HAVE_OMP_SIMD - Failed
> CMake Warning at core/performance_test/peakflops/CMakeLists.txt:18 (message):
>   Compiler doesn't seem to support `#pragma simd`, skipping performance tests
> ```
> 
> Ideas?

Does the changes you added to `CMAKE_REQUIRED_FLAGS` have some un-intended side affects? It uses to work for me locally, but now I see the same with gcc 9. I guess if I can see verbose output from the `check_compiles` we will find out pretty easily

rfbird:
> > @rfbird it still says:
> > ```
> > -- Performing Test HAVE_OMP_SIMD - Failed
> > CMake Warning at core/performance_test/peakflops/CMakeLists.txt:18 (message):
> >   Compiler doesn't seem to support `#pragma simd`, skipping performance tests
> > ```
> > 
> > 
> > Ideas?
> 
> Does the changes you added to `CMAKE_REQUIRED_FLAGS` have some un-intended side affects? It uses to work for me locally, but now I see the same with gcc 9. I guess if I can see verbose output from the `check_compiles` we will find out pretty easily

It looks like gcc is strict about the context and is upset about the lack of a loop. I'll add one

junghans:
Ok, now it compiles, but still fails the test!

rfbird:
> Ok, now it compiles, but still fails the test!

Can we make any guarantees about the hardware on that box?

```
37344280 clocks
8.568916 flops/clock
x0_[0] = 2000000.000000
x0_[1] = 2000018.375000
x0_[2] = 2000004.000000
x0_[3] = 2000108.250000
x0_[4] = 2000007.125000
x0_[5] = 2000016.000000
x0_[6] = 2000008.000000
x0_[7] = 2000007.750000
Expected 32.000000 
```

We have options that do things like set the expected output, most notably `CABANA_PERFORMANCE_EXPECTED_FLOPS` and `CABANA_PERFORMANCE_VECLENGTH`, so it may now be a case of selecting the correct config

junghans:
It builds now, do we want to merge the performance phase back into the main test phase?

rfbird:
> It builds now, do we want to merge the performance phase back into the main test phase?

I think it's probably time, yes

junghans:
@rfbird can you have a look what is wrong now?

rfbird:
> @rfbird can you have a look what is wrong now?

I looks like the performance tests are running in the canary (and therefor I assume don't see `-O3`) and the performance *sucks*:

`0.222702 flops/clock 1`.

I've got it set to need 1flop/clock (vs 32!) .. I guess even that is too high????

sslattery:
I think 0 is probably fine here

rfbird:
> I think 0 is probably fine here

I'm torn. In that case .. what is the point? The functionality is already covered in the existing tests etc. It does make sense if we plan to turn it on again somewhere down the road I guess

sslattery:
0 to me means we don't care about performance in these builds but we do care that the benchmark isn't broken in terms of functionality by a PR merge. We may check these features in other unit tests but having this in the CI would make sure that any interface changes get caught and updated in the performance tests before a merge is done. Otherwise we would have a failing performance build, not because of bad performance, but because it doesn't compile or it throws.

PR #59, opened 2019-01-04
updating license for 2019

sslattery:


sslattery:
will rebase on #57 for after it is merged so new tutorials also get new license years.

sslattery:
@junghans Just thought it would be consistent if they all changed. Do most leave the original date the file was made?

sslattery:
ping

dalg24:
No opinion on this.  Did you and @junghans come to an agreement?

cmake/Cabana.pc.cmakein
```diff
@@ -16,7 +16,7 @@
 
 # Use this in a Makefile:
 # myapp: myapp.cpp
-# 	$(CC) `pkg-config --cflags --libs Cabana` -o $@ $<
+#   $(CC) `pkg-config --cflags --libs Cabana` -o $@ $<
```
junghans: A tab here was actually correct!

sslattery:
Good idea on the script - let's implement when we get to next year

cmake/Cabana.pc.cmakein
sslattery: Does this affect the correctness of this file? This looks like my editor converted tabs to spaces or something.

junghans: This a makefile rule, which needs a tab there, having spaces is just very error-prone. We could put `<TAB>` there to make it obvious if that would make it easier.

sslattery: would this tab be to avoid copy and paste errors? I just now realized that was in a comment

PR #62, opened 2019-01-04
refactoring cmake c++ macros to protect include of MPI-only code

sslattery:
MPI-only headers were including in `Cabana_Core.hpp` without any protection for when MPI was not enabled. This PR refactors the use of C++ macros generated by CMake to enable this.


core/src/CabanaCore_config.hpp.cmakein
```diff
@@ -1,8 +1,10 @@
 #ifndef CABANA_CORE_CONFIG_HPP
 #define CABANA_CORE_CONFIG_HPP
 
-#define Cabana_VERSION_STRING "@PROJECT_VERSION@"
+#define CABANA_VERSION_STRING "@PROJECT_VERSION@"
 
-#define Cabana_GIT_COMMIT_HASH "@Cabana_GIT_COMMIT_HASH@"
+#define CABANA_GIT_COMMIT_HASH "@Cabana_GIT_COMMIT_HASH@"
+
+#define CABANA_ENABLE_MPI @Cabana_USE_MPI@
```
dalg24:
```suggestion
# cmakedefine01 CABANA_ENABLE_MPI
```

CMakeLists.txt
```diff
@@ -67,8 +67,12 @@ endif()
 option(Cabana_ENABLE_MPI "Build Cabana with MPI support" OFF)
 if(Cabana_ENABLE_MPI)
   find_package(MPI REQUIRED)
+  set(Cabana_USE_MPI 1)
```
dalg24: No need for that variable.

core/src/CabanaCore_config.hpp.cmakein
sslattery: good call - cleaned this up and committed with you as co-author

PR #63, opened 2019-01-09
Fixing error in neighbors for comm plan input in tutorials

sslattery:
An incorrect neighbor set was generated for the communication plans in the halo and distributor tutorials. This was causing MPI hangs on some systems.


junghans:
Travis segfaults?

sslattery:
Have been looking at the segfaults for a while now - unable to reproduce locally so far

PR #66, opened 2019-01-24
Add support for ECP-PI

junghans:
/cc @tgamblin @ginsburgnm @dnic

### DIFF
--- a/.ecp-gitlab-ci.yml
+++ b/.ecp-gitlab-ci.yml
@@ -1,0 +1,110 @@
+# see https://docs.gitlab.com/ce/ci/yaml/README.html for all available options
+variables:
+  CI_DEBUG_SERVICES: trace
+  SCHEDULER_PARAMETERS: "-P STF002 -J ginsburgtest -W 0:03 -nnodes 2"
+
+stages:
+  - buildKokkos
+  - build
+#  - test
+
+.BuildKokkos:
+  stage: buildKokkos
+  script:
+    - CI_PROJECT_DIR=${PWD}
+    - module load gcc
+    - j="$(grep -c processor /proc/cpuinfo 2>/dev/null)" || j=0; ((j++))
+    - for i in ${BACKENDS}; do KOKKOS_OPTS+=( --with-${i,,[A-Z]} ); done
+    - git clone --depth=1 https://github.com/kokkos/kokkos.git &&
+      pushd kokkos &&
+      mkdir build &&
+      pushd build &&
+      ../generate_makefile.bash --prefix=${CI_PROJECT_DIR}/kokkos.install ${KOKKOS_OPTS[@]} &&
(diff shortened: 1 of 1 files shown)


ginsburgnm:
@junghans 
Looks reasonable to me for build purposes, I just want to make sure you're aware, none of these jobs will submit to the LSF scheduler. as they are tagged "nobatch". It looks like it's all build anyway though, so it probably makes sense that you are not submitting to the scheduler yet.

This means though: `SCHEDULER_PARAMETERS: "-P STF002 -J ginsburgtest -W 0:03 -nnodes 2"`
will be ignored. It's fine if you want to keep it there for reference to when you do want to submit to the scheduler; but, you will need to change the project name ( `-P STF002` ) that you are charging(as it will be rejected if you attempt to charge to that project). You should probably change the job name ( `-J ginsburgtest` ) to something more meaningful to you as my name probably isn't too helpful. `-W 0:03` is walltime of 3 minutes, which probably will not suffice for real runs, but for short circuit testing is probably useful. -nnodes is number of nodes, so I have no opinion on what is appropriate for your team.

glad this is looking to be helpful to you; let me know if/when you run into strange things :+1:

junghans:
@ginsburgnm, thanks for pointing that out, @sslattery and I will enable the mpi tests soon. These will run as `batch`.

junghans:
ping @sslattery

.ecp-gitlab-ci.yml
```diff
@@ -0,0 +1,110 @@
...
+    - for i in ${BACKENDS}; do KOKKOS_OPTS+=( --with-${i,,[A-Z]} ); done
+    - git clone --depth=1 https://github.com/kokkos/kokkos.git &&
+      pushd kokkos &&
+      mkdir build &&
+      pushd build &&
+      ../generate_makefile.bash --prefix=${CI_PROJECT_DIR}/kokkos.install ${KOKKOS_OPTS[@]} &&
+      make -j${j} &&
+      make install &&
+      popd &&
+      popd
```
dalg24: Why did you chose to use `pushd` and `popd` rather than `cd`?  You don't seem to take any advantage of being able to return to the paths in the directory stack.  Or am I missing something?

junghans: Just my personal preference, plus I like that it shows what directory I am in.

junghans: https://unix.stackexchange.com/questions/272965/pushd-popd-vs-cd-cd-in-bash-and-zsh

junghans:
@dalg24 please review!

dalg24:
@junghans [pipeline is stuck](https://code.ornl.gov/ecpcitest/ecp-copa/cabana/pipelines/35314) on onrl mirror repo.

Issue #67, opened 2019-01-25
c++: error: unrecognized command line option '--diag_suppress=esa_on_defaulted_function_ignored'


junghans:
Cuda fans, any idea about this error:
```
c++: error: unrecognized command line option '--diag_suppress=esa_on_defaulted_function_ignored'
```

I tried a different gcc already....

Details:
```
Running with gitlab-runner  (c364eff5)
  on ascent_copa_setuid 47b77dfd
Using SetUID Shell executor...
Running on login1...
Fetching changes...
Removing kokkos.install/
Removing kokkos/
HEAD is now at 6e6c9eb try newer gcc
Checking out 6e6c9eb9 as ci-cuda...
Skipping Git submodules setup
Downloading artifacts for BuildKokkos Cuda (38102)...
About to register the Shell executor type...
About to register the Batch Runner...
Downloading artifacts from coordinator... ok        id=38102 responseStatus=200 OK token=xheXy1AZ
$ CI_PROJECT_DIR=${PWD}
$ module load cmake
$ module load cuda
$ module load gcc/6.4.0

Lmod is automatically replacing "xl/16.1.1-1" with "gcc/6.4.0".


Due to MODULEPATH changes, the following have been reloaded:
  1) spectrum-mpi/10.2.0.10-20181214

$ for i in ${BACKENDS}; do CMAKE_OPTS+=( -DCabana_ENABLE_${i}=ON ); done
$ j="$(grep -c processor /proc/cpuinfo 2>/dev/null)" || j=0; ((j++))
$ mkdir build && cd build && cmake -DCMAKE_PREFIX_PATH=${CI_PROJECT_DIR}/kokkos.install -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_Serial=OFF -DCabana_ENABLE_EXAMPLES=OFF ${CMAKE_OPTS[@]} .. && make -k -j${j} VERBOSE=1 && make test CTEST_OUTPUT_ON_FAILURE=1
-- The CXX compiler identification is GNU 6.4.0
-- Check for working CXX compiler: /sw/ascent/gcc/6.4.0/bin/c++
-- Check for working CXX compiler: /sw/ascent/gcc/6.4.0/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
[...]

PR #73, opened 2019-02-15
Fix bug in configuration if MPIEXEC_MAX_NUMPROCS is equal to 1

dalg24:
An attempt to fix the configuration error reported [here](https://github.com/ECP-copa/Cabana/issues/67#issuecomment-464126658).

@jgalarowicz Please see if this resolves your problem.

Disclaimer: I don't have a current build of Cabana so I haven't actually tried the patch :)


sslattery:
Looking at the CI output your fix actually reproduced the error

dalg24:
Fixed

dalg24:
Merging as it [seems to have resolved the issue](https://github.com/ECP-copa/Cabana/issues/67#issuecomment-464235695).

PR #81, opened 2019-03-12
Adding templated type alias for AoSoA slices

### LINKED ISSUE #76: Add a typedef for Slices in the AoSoA
Currently there is no typedef within an `AoSoA` for a slice of a given member returned by a call to `AoSoA::Slice<M>()`. This can then force the users to write out the type in certain circumstances. We want an interface that gives the type as `AoSoA::slice_type<M>` where a templated type alias is used.

sslattery:
A user can now define the type of a slice as follows:
```
using aosoa_type = Cabana::AoSoA<MemberTypes,MemorySpace>;

typename aosoa_type::member_slice_type<member_index> slice;

slice = aosoa.slice<member_index>();
```

Closes #76

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -112,6 +112,15 @@ class AoSoA
     using member_pointer_type =
         typename std::add_pointer<member_value_type<M> >::type;
 
+    // Member slice type at a given member index M.
+    template<std::size_t M>
+    using member_slice_type =
+        Slice<member_data_type<M>,
+              memory_space,
+              DefaultAccessMemory,
+              vector_length,
+              sizeof(soa_type) / sizeof(member_value_type<M>)>;
+
   public:
 
     /*!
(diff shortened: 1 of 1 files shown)


dalg24:
> A user can now define the type of a slice as follows:
> 
> ```
> using aosoa_type = Cabana::AoSoA<MemberTypes,MemorySpace>;
> 
> typename aosoa_type::member_slice_type<member_index> slice;
> 
> slice = aosoa.slice<member_index>();
> ```

Prefer another template alias to the C-style typedef declaration.  Also I assume you forgot the variable name in the 3rd line.

streeve:
This works for CabanaMD (full neighbor list). For the half neighbor lists only the force needs to be atomic - do you think it's reasonable to expose the memory access here?

sslattery:
When you grab a slice the memory is always the default type. If you want an atomic slice you just reassign:
```
typename member_slice_type<m>::atomic_access_slice atomic_slice = slice;
```
In other words, this type alias gives you what is returned by `AoSoA.slice<>()` and you manipulate the memory as needed with the assignment operations to slices of different types.

PR #83, opened 2019-03-12
Verlet List Update

### LINKED ISSUE #71: MD: 2D neighbor list
Add new algorithm (already in ExaMiniMD)

### LINKED ISSUE #75: Reuse LinkedCellList in VerletList Construction
If a user has already created a LinkedCellList for another purpose they should be able to reuse it when building a VerletList as long as that LinkedCellList was built with the same parameters.

sslattery:
* Adds 2D Verlet list storage in addition to the existing CSR storage. These are created by a new layout template parameter on the Verlet list. Closes #71 
* ~~Allows for the reuse of a `LinkedCellList` when constructing the Verlet list. This should save a lot of time for those cases where a `LinkedCellList` is constructed and used elsewhere.~~ (will not do per discussion in #75) Closes #75

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -20,6 +20,67 @@
 
 namespace Cabana
 {
+//---------------------------------------------------------------------------//
+// Verlet List Memory Layout Tag.
+//---------------------------------------------------------------------------//
+// CSR Layout
+struct VerletLayoutCSR {};
+
+// 2D Layout.
+struct VerletLayout2D {};
+
+//---------------------------------------------------------------------------//
+// Verlet List Data.
+//---------------------------------------------------------------------------//
+template<class MemorySpace, class LayoutTag>
+struct VerletListData;
+
+template<class MemorySpace>
+struct VerletListData<MemorySpace,VerletLayoutCSR>
+{
+    // The memory space in which the neighbor list data resides.
+    using memory_space = MemorySpace;
+
+    // Number of neighbors per particle.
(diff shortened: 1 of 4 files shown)


core/src/Cabana_VerletList.hpp
```diff
@@ -20,6 +20,67 @@
 
 namespace Cabana
 {
+//---------------------------------------------------------------------------//
+// Verlet List Memory Layout Tag.
+//---------------------------------------------------------------------------//
+// CSR Layout
+struct VerletLayoutCSR {};
```
dalg24: Compressed Row Storage?

streeve:
2D list works (and has much better GPU performance)

streeve:
I'm getting empty neighbor lists when I try to reuse the linkedcell - still looking into it

core/src/Cabana_VerletList.hpp
streeve: Compressed sparse row?

sslattery:
@streeve Keep me filled in on the linked cell error. I have been thinking about how to make the interfaces simpler to ensure the linked cell list one passes to the verlet list is consistent with the one that the verlet list would have made on its own.

sslattery:
@streeve Did you resolve your error when re-using the list?

streeve:
Yes, I was able to reuse the LinkedCellList once I made sure the inputs were identical and did not permute before building the VerletList

core/src/Cabana_VerletList.hpp
sslattery: yes - Compressed Sparse Row

PR #84, opened 2019-03-12
Removing change log in favor of GitHub release tracking

sslattery:
GitHub has a very nice release tracking mechanism [here](https://github.com/ECP-copa/Cabana/releases). I recently updated this to also look like a change log including links to all of the pull requests that enabled new features. I'd like to use the GitHub release page rather than a change log. 

This PR removes the change log in the repo so we don't have to put the same information in 2 places when we release.


dalg24:
I fear that these descriptions that one can attach to a GH release are not under version control.  I'd recommend you keep a copy in the repo.

sslattery:
@dalg24 and I discussed and although I hate duplication of effort it is hard to argue against having a change log in version control. Going to do another PR which updates the change log for version 0.2

sslattery:
Closing in favor of #85

PR #85, opened 2019-03-13
Fix 0.2.0 CMake Version and Change Log

sslattery:
Updates the CMake version and adds updates to the change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Change Log
 
+## [0.2.0]()
+
+**New Features**
+
+- An optional MPI dependency has been added. Note that when CUDA is enabled the MPI implementation is expected to be CUDA-aware. [#45](https://github.com/ECP-copa/Cabana/pull/45)
+- Particle redistribution via MPI. Implemented in the `Cabana::Distributor` [#43](https://github.com/ECP-copa/Cabana/pull/43)
+- Particle halo exchange via MPI. Implemented in the `Cabana::Halo` [#43](https://github.com/ECP-copa/Cabana/pull/43)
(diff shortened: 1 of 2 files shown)


dalg24:
You want to merge these into master.  But what's the plan about release 0.2?  Are you going to update the tag?

sslattery:
Yes we would update the tag with these changes

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## [0.2.0]()
```
dalg24: What the deal with this syntax?  This is essentially an empty (invalid) URL.

sslattery: yeah that is weird. I have no idea why I did that. maybe it should link to the GitHub release page

junghans:
I would make it v0.2.1, updating public tag is just bad practice.

CHANGELOG.md
sslattery: fixed

dalg24:
Answering here to https://github.com/ECP-copa/Cabana/issues/104#issuecomment-475797916 

@sslattery said:
> What about #85?

This PR was trying to fix:
* the runtime `Cabana::get_version()` that returns `"0.1.0"` but there `Cabana::git_commit_hash()` does the right thing
* the version number displayed on Doxygen but who's going to look at the 0.2.0 documentation after we release 0.3.0 which should be coming soon enough

Also this PR was done against master which was probably not the best unless we acted before a bunch other stuff got merged.  I suggest we forget about fixing 0.2.0 rebase #85 onto the current master, drop the to the version number change and cherry-pick the updates of the change log.  Then force-push and edit the subject line of the PR.

dalg24:
I can take care of it if we agree on that.

sslattery:
@dalg24 I agree with that strategy.

PR #88, opened 2019-03-13
Adding Support for Kokkos::CudaSpace

### LINKED ISSUE #38: Add regular CUDA Memory Space
We need to support the regular CUDA memory space in Cabana. The NVIDIA guys have informed me that this will achieve superior performance when using CUDA-aware MPI and GPUDirect implementations for message passing. We already have `deepCopy` so we just need to expose this in the interface and update the unit tests to copy data explicitly to the executing device.

### LINKED ISSUE #86: slicing is broken for non-UVM CUDA
There is a bad `view(0).template ptr<M>()` pattern being used to get pointers for slices. `view(0)` is an illegal access on the CPU. There should be something like `SoA::static_ptr` that can be used as `soa_type::template static_ptr<M>(view.data())`.

sslattery:
Adds support for the CUDA space in Kokkos:

- Fixes bug in constructing slices
- Adds proper data movement in unit tests to allow for testing with cuda space
- Adds UVM space tests that are separate from CUDA space tests

Closes #38 
Closes #86


core/src/Cabana_AoSoA.hpp
```diff
@@ -319,18 +319,19 @@ class AoSoA
             0 == sizeof(soa_type) % sizeof(member_value_type<M>),
             "Slice stride cannot be calculated for misaligned memory!" );
 
-        member_pointer_type<M> data_ptr =
+        member_pointer_type<M> ptr =
```
ibaned: Is this conditional still needed when using `staticPtr`? Doesn't `staticPtr(nullptr)==nullptr`?

sslattery: true! I'll get rid of that conditional. That was the result of rebasing on master which had a previous bug fix in this area.

CMakeLists.txt
```diff
@@ -54,6 +54,15 @@ if(NOT ENABLE_DEVICES)
 endif()
 message(STATUS "Enable Devices: ${ENABLE_DEVICES}")
 
+# if we are using Cuda, also add a "CudaUVM" device to use that memory space
+# in the test suite. Also enable a Cabana UVM option so the tests get triggered.
+foreach(_device ${ENABLE_DEVICES})
+  if(_device STREQUAL Cuda)
+    list(APPEND CABANA_SUPPORTED_DEVICES CudaUVM)
+    set(Cabana_ENABLE_CudaUVM ON)
```
dalg24: Where are these used?

core/src/Cabana_SoA.hpp
```diff
@@ -337,20 +336,27 @@ struct SoA<MemberTypes<Types...>,VectorLength>
...
         Impl::StructMember<M,vector_length,member_data_type<M> >& base = *this;
         return &base;
     }
+
+    // Get a pointer to a given SoA.
+    template<std::size_t M>
+    static void* staticPtr( SoA* p )
+    {
+        Impl::StructMember<M,vector_length,member_data_type<M> >& base = *p;
+        return &base;
```
dalg24: Would you explain how `soa_type::template staticPtr<M>(ptr)` differs from `static_cast<void*>(ptr)`?
What am I missing?

dalg24: I missed the multiple inheritance of `SoAImpl` from `SructMember<0, T0>`,  `SructMember<1, T1>`, etc.
This is an implicit conversion to a sub-object of the struct.

I think we could just do `return static_cast<Impl::StructMember<M,vector_length,member_data_type<M>>*>(p);`

sslattery:
I would like to note that the Cuda tests that are failing can be fixed when adding this the the mpirun launch line: `--mca btl_smcuda_use_cuda_ipc 0`. If you inspect the Travis output closely you will see that there is a cuda/mpi issue after the test finishes which shows up as a segfault. I had the CADES people fix this on condo so that these flags were not necessary - I'll look into how they did the configuration.

sslattery:
Other than the flag fix mentioned above this is done.

sslattery:
Should we consider adding `--mca btl_smcuda_use_cuda_ipc 0` to the test command line or should we find some other solution regarding MPI configuration? FWIW I see this error on my own linux box in addition to Travis

dalg24:
@sslattery I took the liberty to cherry-pick https://github.com/ECP-copa/Cabana/pull/100/commits/d36f39e4bf18b7f11f306ecdec6e324822417d7b and push onto your branch.  This seems to resolve the issues in the automated CUDA testing.

CMakeLists.txt
sslattery: This get used in the unit test `CMakeLists.txt` - any time you see `if(Cabana_ENABLE_${_device})` it will then include a UVM device. The way we have our device settings means that we need two testing device types when CUDA is enabled - one with UVM and one without. This line adds the extra UVM device.

dalg24: OK

core/src/Cabana_SoA.hpp
sslattery: added a commit for this

core/unit_test/CudaUVM/tstTuple_CudaUVM.cpp
```diff
@@ -0,0 +1,13 @@
...
+ *                                                                          *
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#include <CudaUVM/TestCudaUVM_Category.hpp>
+#include <tstTuple.hpp>
```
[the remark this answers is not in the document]
sslattery: resolved with #99

core/unit_test/tstAoSoA.hpp
```diff
@@ -243,42 +251,51 @@ void testRawData()
...
+                p4[ s * st4 + i ] = (s + i) * 5.0;
+
+                // Member 2 has some extra dimensions so add those to the
+                // indexing. Note this is layout left.
+                for ( int j = 0; j < m2e0; ++j )
+                    for ( int k = 0; k < m2e1; ++k )
+                        p2[ s * st2 + j * 16 * m2e1 + k * 16 + i ] =
+                            (s + i + j + k) * 3.0;
+            }
+        } );
```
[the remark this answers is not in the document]
sslattery: added a fence

core/unit_test/tstAoSoA.hpp
```diff
@@ -243,42 +251,51 @@ void testRawData()
...
-            // Member 2 has some extra dimensions so add those to the
-            // indexing. Note this is layout left.
-            for ( int j = 0; j < m2e0; ++j )
-                for ( int k = 0; k < m2e1; ++k )
-                    p2[ s * st2 + j * 16 * m2e1 + k * 16 + i ] =
-                        (s + i + j + k) * 3.0;
-        }
-    }
+    Kokkos::parallel_for(
+        "raw_data_fill",
```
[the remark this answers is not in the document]
sslattery: I add them out of habit - its actually helped me find errors in the past if Kokkos is compiled in debug mode. Actually it helped me find one this week in this PR because it can be outputted in a Kokkos error message.

sslattery:
oh yeah - was wondering why I had to add that file again. I figured the rebase blew it away but it was really in a different directory.

Issue #92, opened 2019-03-14
Add a Label to AoSoA and Slice Objects


sslattery:
The `Kokkos::View` has a string label. The `AoSoA` and `Slice` should have one as well (and it should be applied to the `Kokkos::View` used as an implementation detail). It would be nice if the slice label included the label of the AoSoA from which it was derived in addition to a suffix indicating its member index. For example, an AoSoA with label "MyData" when sliced over member 1 would have a label "MyData_1". This label should also be propagated to mirrors.

sslattery:
Note that labeling for slices is not possible at this time. The purpose of the label is to provide information in the Kokkos memory tracker. Slices use unmanaged memory and Kokkos Views with unmanaged memory do not have labels because they are not tracked in the memory tracker.

sslattery:
I'll still add a slice label as it may be useful - it just won't get used in the memory tracker.

PR #93, opened 2019-03-15
AoSoA Mirror

### LINKED ISSUE #90: Add mirror view support for AoSoA
For more efficient heterogenous computing support for `AoSoA` similar to `Kokkos::create_mirror_view` and `Kokkos::create_mirror_view_and_copy` is needed.

sslattery:
This adds a mirror capability equivalent to that provided for `Kokkos::View`. This closes #90 and is intended to make the implementation of #88 easier.

Also note that this new capability will need to be updated when #92 is resolved.

### DIFF
--- a/core/src/Cabana_DeepCopy.hpp
+++ b/core/src/Cabana_DeepCopy.hpp
@@ -134,10 +134,9 @@ namespace Experimental
   \brief Create a mirror view of the given AoSoA in the given memory
   space. Same space specialization returns the input AoSoA.
 
-  \note The semantics of using the word view in the name of this function
-  indicate that memory allocation will only occur if the requested mirror
-  memory space is different from that of the input AoSoA. If they are the
-  same, the original AoSoA (e.g. a view of that AoSoA) is returned.
+  \note Memory allocation will only occur if the requested mirror memory space
+  is different from that of the input AoSoA. If they are the same, the
+  original AoSoA (e.g. a view of that AoSoA) is returned.
  */
 template<class Space, class SrcAoSoA>
 inline
@@ -158,8 +157,7 @@ create_mirror_view_and_copy(
   space and deep copy the AoSoA into the mirror. Different space
   specialization allocates a new AoSoA and performs the deep copy.
 
-  \note The semantics of using the word view in the name of this function
-  indicate that memory allocation will only occur if the requested mirror
+  \note Memory allocation will only occur if the requested mirror
   memory space is different from that of the input AoSoA. If they are the
   same, the original AoSoA (e.g. a view of that AoSoA) is returned.
  */
--- a/core/src/Cabana_DeepCopy.hpp
+++ b/core/src/Cabana_DeepCopy.hpp
@@ -127,6 +127,119 @@ inline void deep_copy(
 }
 
 //---------------------------------------------------------------------------//
+/*!
+  \brief Create a mirror of the given AoSoA in the given memory
+  space. Same space specialization returns the input AoSoA.
+ */
+template<class Space, class SrcAoSoA>
+SrcAoSoA
+create_mirror_aosoa(
+    const Space&,
+    const SrcAoSoA& src,
+    typename std::enable_if<(is_aosoa<SrcAoSoA>::value &&
+                             std::is_same<typename SrcAoSoA::memory_space,
+                             typename Space::memory_space>::value)>::type* = 0 )
+{
+    return src;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief Create a mirror of the given AoSoA in the given memory
+  space. Different space specialization allocates a new AoSoA.
+ */
+template<class Space, class SrcAoSoA>
+AoSoA<typename SrcAoSoA::member_types,
+      typename Space::memory_space,
+      SrcAoSoA::vector_length>
+create_mirror_aosoa(
+    const Space&,
+    const SrcAoSoA& src,
+    typename std::enable_if<(is_aosoa<SrcAoSoA>::value &&
+                             !std::is_same<typename SrcAoSoA::memory_space,
+                             typename Space::memory_space>::value)>::type* = 0 )
+{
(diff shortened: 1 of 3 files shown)


core/src/Cabana_DeepCopy.hpp
```diff
@@ -126,6 +126,119 @@ inline void deep_copy(
     }
 }
 
+//---------------------------------------------------------------------------//
+/*!
+  \brief Create a mirror of the given AoSoA in the given memory
+  space. Same space specialization returns the input AoSoA.
+ */
```
dalg24: Kokkos provides both `create_mirror` and `create_mirror_view` functions.  The difference between the two is that `create_mirror` always allocate a new view while `create_mirror_view ` only creates a new view if the original one has a different memory space.  I am not a big fan of this naming convention.  `_view` does make sense in a way because it does give you access from the given execution space into the underlying multidimensional array.  But `View` is also what Kokkos call its md array but and the name does not suggest is that it retains shared ownership 
over that array...  Anyway here you replaced `_view` with `_aosoa` but I think it is wrong, you should still call it `_view` because that is what the semantics of your functions are.

core/src/Cabana_DeepCopy.hpp
```diff
@@ -126,6 +126,119 @@ inline void deep_copy(
...
+template<class Space, class SrcAoSoA>
+SrcAoSoA
+create_mirror_aosoa(
```
dalg24: inline all of these.

sslattery:
per discussion with @dalg24 I have reduced the number of functions implemented to cover what is needed for #88 - we can add more in the future as needed. Also we discussed the semantics of the word "Mirror" and "Mirror View" in Kokkos and how it applies here. The name I have chosen for our function represents those semantics but may be subject to change - hence inclusion in the `Experimental` namespace.

core/src/Cabana_DeepCopy.hpp
sslattery: done

sslattery: note that I used regular `inline` instead of `KOKKOS_INLINE_FUNCTION`

core/src/Cabana_DeepCopy.hpp
sslattery: per discussion with @dalg24 I have reduced the number of functions implemented to cover what is needed for #88 - we can add more in the future as needed. Also we discussed the semantics of the word "Mirror" and "Mirror View" in Kokkos and how it applies here. The name I have chosen for our function represents those semantics but may be subject to change - hence inclusion in the Experimental namespace.

core/src/Cabana_DeepCopy.hpp
```diff
@@ -126,6 +126,66 @@ inline void deep_copy(
...
+//---------------------------------------------------------------------------//
+namespace Experimental
+{
+//---------------------------------------------------------------------------//
+/*!
+  \brief Create a mirror view of the given AoSoA in the given memory
+  space. Same space specialization returns the input AoSoA.
+
+  \note The semantics of using the word view in the name of this function
+  indicate that memory allocation will only occur if the requested mirror
```
dalg24: \note ~~The semantics of using the word view in the name of this function
	  indicate that~~ (M)emory allocation will only occur if the requested mirror

core/unit_test/tstDeepCopy.hpp
```diff
@@ -118,6 +118,89 @@ void testDeepCopy()
...
+    Cabana::AoSoA<DataTypes,TEST_MEMSPACE> aosoa( num_data );
+
+    // Initialize data.
+    float fval = 3.4;
+    double dval = 1.23;
+    int ival = 1;
+    auto slice_0 = aosoa.template slice<0>();
+    auto slice_1 = aosoa.template slice<1>();
+    auto slice_2 = aosoa.template slice<2>();
+    auto slice_3 = aosoa.template slice<3>();
```
dalg24: I really dislike this syntax.  We should consider implementing this as a free function:
```C++
auto slice_0 = slice<0>(aosoa);
auto slice_1 = slice<1>(aosoa);
```

core/unit_test/tstDeepCopy.hpp
```diff
@@ -118,6 +118,89 @@ void testDeepCopy()
...
+        std::is_same<TEST_MEMSPACE,
+                     decltype(same_space_copy)::memory_space>::value;
+    EXPECT_TRUE( ssc_same_ms );
+    bool ssc_same_mt =
+        std::is_same<DataTypes,
+                     decltype(same_space_copy)::member_types>::value;
+    EXPECT_TRUE( ssc_same_mt );
+
+    // Check values.
+    checkDataMembers( same_space_copy, fval, dval, ival, dim_1, dim_2, dim_3 );
```
dalg24: You need to test that there was no memory allocation.  Checking that `aosoa` and `same_space_copy` are the same type and have the same address would do.

core/src/Cabana_DeepCopy.hpp
sslattery: done

core/unit_test/tstDeepCopy.hpp
sslattery: done

core/unit_test/tstDeepCopy.hpp
sslattery: we had something like this before. I'd be fine adding one back in as an alternative to the AoSoA member function

core/unit_test/tstDeepCopy.hpp
dalg24: Sure `EXPECT_EQ( aosoa.ptr(), same_space_copy.ptr() );` does check that `AoSoA` that is being returned has the same underlying data but `create_mirror_view_and_copy` states that the `AoSoA` argument is being returned which you would test with `EXPECT_EQ( &aosoa, &same_space_copy );`.

PR #94, opened 2019-03-18
add doxygen coverage

junghans:


junghans:
@sslattery @dalg24 we broke doxygen's xml output:
```
error: error while parsing xml file xml/classCabana_1_1AoSoA.xml : not well-formed (invalid token): line 25, column 33
```

junghans:
apparently there was a bug in doxygen, so this PR won't be able to merged soon. Alternatively we need to build our own doxygen, which might not be worth it.

junghans:
Or we could patch the xml files using `sed`.

sslattery:
@junghans do you still want a review here if we are unable to merge due to the bug?

junghans:
@sslattery, I am working on workaround...

junghans:
@dalg24, 20.4% doxygen coverage isn't great, but at least we have the numbers now!

junghans:
@dalg24 please review

sslattery:
@junghans what's up with the coverage changes?

sslattery:
Is it that doxygen is now included in the coverage?

.travis.yml
```diff
@@ -53,6 +53,10 @@ before_script:
       popd;
     fi
   - for i in ${BACKENDS}; do CMAKE_OPTS+=( -DCabana_ENABLE_${i}=ON ); done
+  - if [[ ${COVERAGE} ]]; then
+       pip install --user coverxygen &&
+       mkdir -p $HOME/.local/bin && wget -O $HOME/.local/bin/codecov https://codecov.io/bash && chmod +x $HOME/.local/bin/codecov;
```
sslattery: OK so here it looks like you are adding doxygen to the coverage - correct?

.travis.yml
```diff
@@ -82,12 +87,17 @@ after_success:
  - ccache -s
  - test "${TRAVIS_BUILD_STAGE_NAME}" = "Test" || travis_terminate 0
  - if [[ ${COVERAGE} ]]; then
-     pushd build && wget -O codecov https://codecov.io/bash && chmod +x codecov &&
+     pushd build &&
      if [[ ${CC} = clang ]]; then
-       ./codecov -x "llvm-cov gcov" -F "${CC}";
+       codecov -x "llvm-cov gcov" -F "${CC}";
```
sslattery: What is different about coverage with LLVM?

.travis.yml
```diff
@@ -82,12 +87,17 @@ after_success:
...
      else
-       ./codecov -F "${CC}";
+       codecov -F "${CC}";
      fi &&
+     if [[ $(doxygen --version) != 1.8.16 ]]; then
+       echo "Patching Doxygen files (see doxygen/doxygen#6892)" &&
+       sed -e '/enable_if</s/&/\&amp;/g' -e '/enable_if/s/<\([^a-z/]\)/\&lt;\1/g' -e '/enable_if/s/ >/ \&gt;/g' -i xml/*.xml;
```
sslattery: Here it looks like you are removing the offending code in the XML files? Is 1.8.16 the latest doxygen version that has the bug fix?

CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ endif()
 
 find_package(Doxygen)
 if(Doxygen_FOUND)
+  set(DOXYGEN_GENERATE_XML YES)
```
sslattery: So ultimately coverage with Doxygen requires coverage to assess the doxygen-generated XML files?

junghans: Yes, it needs the xml outputs!

.travis.yml
junghans: Yes, the fix is part of the next release, 1.8.16.

.travis.yml
junghans: For `clang` you need to use `llvm-cov gcov` instead of `gcov`.

junghans:
Yes, see the percentage for `#doxygen` Flag above.

.travis.yml
junghans: `coverxygen` is the new tool. I download codecov here as it seems more like a `before_script:` step.

junghans:
And coverage check is failing now as doxygen pulls the overall coverage down.

dalg24:
Please post a link to the Doxygen coverage report.  I browsed a bit on codecov and could not make any sense of what I saw.

junghans:
> Please post a link to the Doxygen coverage report. I browsed a bit on codecov and could not make any sense of what I saw.

Yes, it is confusing to find, you need to go to the file view and then click on “doxygen” at the very top. Unfortunately  there is no summary view per tags, only per file.

.travis.yml
sslattery: should we use `<` instead of `!=` for the version check assuming that future versions will also not have this bug?

junghans:
@sslattery version comparison is hard:
```
$ [[ 1.8.16 < 1.8.2 ]] && echo error
error
```
I did `[[ $(doxygen --version) != 1.8.16 ]]` as Doxygen release are slow and we can delete the whole block anyhow once doxygen is up to date.

junghans:
@sslattery #112

PR #96, opened 2019-03-19
.travis: add canary build, fix clang coverage reporting and minimize scafacos builds

junghans:


dalg24:
Please describe what these changes are intended for?

junghans:
@dalg24 @sslattery I want to one build, the _canery_ build, frist before kicking of all the CI, so basically as kind of sniff test.

junghans:
@dalg24 I fixed one more issue for coverage reporting using clang, but otherwise ready to merge.

junghans:
@dalg24 please review

junghans:
Well, Travis has no early fail feature, hence you have sequentially do the builds to get early abort.

I will add some comment re: explanation.

.travis.yml
```diff
@@ -69,7 +80,16 @@ script:
 
 after_success:
  - ccache -s
- - if [[ ${COVERAGE} ]]; then pushd build && bash <(curl -s https://codecov.io/bash) && popd; fi
+ - test "${TRAVIS_BUILD_STAGE_NAME}" = "Test" || travis_terminate 0
```
dalg24: Isn't this always going to terminate early?

.travis.yml
```diff
@@ -69,7 +80,16 @@ script:
...
+ - if [[ ${COVERAGE} ]]; then
+     pushd build && wget -O codecov https://codecov.io/bash && chmod +x codecov &&
+     if [[ ${CC} = clang ]]; then
+       ./codecov -x "llvm-cov gcov" -F "${CC}";
+     else
+       ./codecov -F "${CC}";
+     fi &&
```
dalg24: You did not advertise you were changing logic to upload the code coverage in this PR.

.travis.yml
```diff
@@ -34,13 +43,15 @@ before_script:
...
     popd &&
     popd
-  - travis_retry wget https://github.com/scafacos/scafacos/releases/download/v1.0.1/scafacos-1.0.1.tar.gz &&
-    tar -xzf scafacos-*.tar.gz &&
-    pushd scafacos* &&
-    ./configure --enable-fcs-solvers=direct,p2nfft,p3m,ewald,fmm --prefix=$HOME/scafacos CC=mpicc CXX=mpicxx FC=mpif90 &&
-    make -j2 &&
-    make install &&
-    popd
+  - if [[ ${SCAFACOS} ]]; then
```
dalg24: Same here.

To be clear I have nothing against the changes but I think you should either capture that in the description, or (better) do that in a separate PR.

.travis.yml
junghans: No, the logic is: is it the test stage or terminate.

.travis.yml
junghans: Done!

.travis.yml
junghans: Done!

dalg24:
> Well, Travis has no early fail feature, hence you have sequentially do the builds to get early abort.
> 
> I will add some comment re: explanation.

I expect this PR would come with a description that states why we want these changes.

Yes I did figure that if the "Canary" build fails, most likely all will, therefore we are wasting computing resources on Travis CI.  But you still need to capture this in the description attached to the PR and justify why we care and that the complexity added is ok.

PR #97, opened 2019-03-20
.ecp-gitlab-ci.yml: update tags

junghans:
Fix ECP-CI.


junghans:
Yeah, I will fix the status reporting later!

PR #99, opened 2019-03-22
Remove test source files that can be generated at configuration time

dalg24:
When reviewing #88 I realized we were repeating ourselves and unnecessarily creating and maintaining a number of test source files by hand.  This PR remove these files and uses CMake to generate them at configuration time.


core/unit_test/CMakeLists.txt
```diff
@@ -52,18 +59,25 @@ endforeach()
 if(${Cabana_ENABLE_MPI})
   foreach(_device ${CABANA_SUPPORTED_DEVICES})
     if(Cabana_ENABLE_${_device})
+      set(_dir ${CMAKE_CURRENT_BINARY_DIR}/${_device})
+      file(MAKE_DIRECTORY ${_dir})
```
sslattery: What happens if this directory has already been created? Does this fail or throw a warning?

core/unit_test/CMakeLists.txt
```diff
@@ -52,18 +59,25 @@ endforeach()
...
       foreach(_test CommunicationPlan Distributor Halo)
-        add_executable(${_test}_test_${_device} ${_device}/tst${_test}_${_device}.cpp mpi_unit_test_main.cpp)
-        target_link_libraries(${_test}_test_${_device} cabanacore cabana_core_gtest)
+        set(_file ${_dir}/tst${_test}_${_device}.cpp)
```
sslattery: This looks like the same code as above - could we make a function out of this? Might be useful elsewhere as we add new packages.

core/unit_test/CMakeLists.txt
dalg24: Neither.

core/unit_test/CMakeLists.txt
dalg24: We could.  Won't fix it here though.

sslattery: works for me

core/unit_test/CMakeLists.txt
junghans: CMake works like `mkdir -p`:
```
$ cmake -E make_directory 1; echo $?
0
$ cmake -E make_directory 1; echo $?
0
```

sslattery: got it

core/unit_test/CMakeLists.txt
junghans: Let's do the refactor in another PR.

PR #106, opened 2019-03-22
Use Kokkos ScopeGuard and deprecate Cabana initialize/finalize 

streeve:
Per discussion with @rfbird and @sslattery this uses Kokkos to init/finalize rather than wrapping with Cabana. 

Eventually, C++14 [[deprecate]] could replace the current and/or the functions be removed


junghans:
OK to test

core/src/Cabana_Core.hpp
```diff
@@ -56,6 +56,7 @@ bool cabana_initialized_kokkos = false;
 //---------------------------------------------------------------------------//
 // Initialize Kokkos, if it needs initialization.
 template <typename... Args>
+__attribute__ ((deprecated))
```
dalg24: Not sure about that one.  This is a GCC thing so it might not be portable across all compilers we want to support.  We might need to define yet another macro...

core/example/benchmark/Cabana_peakflops.cpp
```diff
@@ -320,11 +320,11 @@ void run()
 //---------------------------------------------------------------------------//
 int main( int argc, char **argv )
 {
-    Cabana::initialize(argc,argv);
+    Kokkos::initialize(argc,argv);
```
dalg24: We should consider teaching to use `Kokkos::ScopeGuard` instead.  It doesn't really matter here but it is a safer alternative because `Kokkos::finalize()` always gets called.  For instance if someone change the code and add an if-clause with early return (e.g. parsing user command line arguments and finding the flag `--help`)

core/src/Cabana_Core.hpp
dalg24: I recommend for know you define an empty macro `#define CABANA_DEPRECATED` and you decorate deprecated stuff with it.

dalg24: There has been some talk about requiring C++14.  It would make our lives easier here because we would then just use the `[[deprecated]]` attribute.

dalg24: Of course I missed your comment in the description :)
Anyway IMO you should not use `__attribute__ ((deprecated))`

streeve: So for now I could mark the code with the empty macro, without a warning?

sslattery: Yeah that's not a bad idea - if we move to C++14 soon then we can replace the empty macros with the C++14 attribute as @dalg24 indicated

streeve: Or just wait until the C++14 decision

core/example/benchmark/Cabana_peakflops.cpp
sslattery: Good call - maybe do that in the tutorials?

core/example/tutorial/01_hello_world/hello_world.cpp
```diff
@@ -18,35 +18,15 @@
...
-
-           1) Call Cabana::initialize() with argc and argv as below. This will
-           pass any command line argument to the Kokkos runtime.
-
-           2) Call Cabana::initialize() with no arguments. Defaults and system
-           variables will be used to select things like number of threads.
-    */
-    Cabana::initialize(argc,argv);
+    // The Kokkos runtime must first be initialized.
+    Kokkos::initialize(argc,argv);
```
sslattery: Yeah let's use the `Kokkos::ScopeGuard` here

core/example/benchmark/Cabana_peakflops.cpp
streeve: Yeah, I'll add that in the next commit

core/example/tutorial/01_hello_world/hello_world.cpp
streeve: Macro and ScopeGuard added

dalg24: @streeve you misunderstood how the scope guard works.  It goes like this
```C++
namespace Kokkos {
struct ScopeGuard {
  ScopeGuard(int &argc, char * argv[]) { initialize(argc, argv); }
  ~ScopeGuard() { finalize(); }
};
} // namespace Kokkos
```
So you would actually do:
```diff
- Kokkos::initialize(argc, argv);
+ Kokkos::ScopeGuard scope_guard(argc, argv);
  // do stuff
- Kokkos::finalize();
```
Kokkos is initialized when `scope_guard` is constructed and when it goes out of scope its destructor is called which guarantees that Kokkos gets finalized.
In practice `Kokkos::ScopeGuard` implementation is a bit smarter that what I showed you but you get the gist of it.

core/src/Cabana_Core.hpp
```diff
@@ -12,6 +12,8 @@
 #ifndef CABANA_CORE_HPP
 #define CABANA_CORE_HPP
 
+#define CABANA_DEPRECATED
```
dalg24: Please move this definition to `<Cabana_Macros.hpp>`.

PR #115, opened 2019-04-04
Permute options for slice and linked cell slice

### LINKED ISSUE #44: Add member `permute`
The implementations of `permute` currently only work on an entire `AoSoA`. They should also work on `Slice` as some algorithms only sort some particle members and not others.

Per further discussion with @streeve we would also like a permute function which applies to a subset of the data structure.

streeve:
This extends sorting to ~~part of an AoSoA,~~ slices ~~and part of a slice~~. Also available for sorting from a linked cell list. 

Closes #44


dalg24:
ok to test

dalg24:
add to whitelist

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -319,6 +319,87 @@ void permute(
...
+*/
+template<class LinkedCellListType, class AoSoA_t>
+void permute(
+    const LinkedCellListType& linked_cell_list,
+    AoSoA_t& aosoa,
+    const std::size_t begin,
+    const std::size_t end,
+    typename std::enable_if<(is_linked_cell_list<LinkedCellListType>::value &&
+                             is_aosoa<AoSoA_t>::value),
+    int>::type * = 0 )
```
dalg24: @sslattery  You guys are overusing/misusing SFINAE.  This slows down compile times and does not bring you much.  The error is still going to be about the compiler not finding a matching function to call just as without the template parameters.  Consider using static assertions instead.

sslattery: yeah this is a good point here. in some cases this is necessary to get the compiler to pick the right function implementation but in this case a static assert is more appropriate.

sslattery: @streeve In these functions you can just delete the `enable_if` lines and replace them with `static_assert` in the function implementation using the same boolean check in the `enable_if`

sslattery: Other than that this API is good

core/src/Cabana_Sort.hpp
```diff
@@ -589,6 +589,195 @@ void permute( const BinningDataType& binning_data,
...
+  \param binning_data The binning data.
+
+  \param aosoa The AoSoA to permute.
+
+  \param begin The beginning index of the AoSoA range to sort.
+
+  \param end The end index of the AoSoA range to sort.
+*/
+template<class BinningDataType, class AoSoA_t>
+void permute( const BinningDataType& binning_data,
```
sslattery: We can improve code reuse here. This should be the core implementation. Change the implementation of `permute` that permutes the entire AoSoA to just call this function with `aosoa.begin()` and `aosoa.end()`. That way we only have a single implementation. You can drop this call right in at line 566 and delete the rest.

core/src/Cabana_Sort.hpp
```diff
@@ -589,6 +589,195 @@ void permute( const BinningDataType& binning_data,
...
+  \param slice The slice to permute.
+ */
+template<class BinningDataType, class SliceType>
+void permute( const BinningDataType& binning_data,
+              SliceType& slice,
+              typename std::enable_if<(is_binning_data<BinningDataType>::value &&
+                                       is_slice<SliceType>::value),
+              int>::type * = 0)
+{
+    auto begin = binning_data.rangeBegin();
```
sslattery: Same here - just call the slice version that permutes a range using `begin` and `end` and you can delete the rest of this function implementation.

core/src/Cabana_Sort.hpp
```diff
@@ -589,6 +589,195 @@ void permute( const BinningDataType& binning_data,
...
+  \param begin The beginning index of the AoSoA range to sort.
+
+  \param end The end index of the AoSoA range to sort.
+*/
+template<class BinningDataType, class AoSoA_t>
+void permute( const BinningDataType& binning_data,
+              AoSoA_t& aosoa,
+              const std::size_t begin,
+              const std::size_t end,
+              typename std::enable_if<(is_binning_data<BinningDataType>::value &&
```
sslattery: Per @dalg24 comment before we can replace these with `static_assert` in the function body.

core/src/Cabana_Sort.hpp
```diff
@@ -589,6 +589,195 @@ void permute( const BinningDataType& binning_data,
...
+    auto permute_to_scratch =
+        KOKKOS_LAMBDA( const std::size_t i )
+        {
+            auto permute_i = binning_data.permutation(i);
+            auto s = SliceType::index_type::s( permute_i );
+            auto a = SliceType::index_type::a( permute_i );
+            std::size_t slice_offset = s*slice.stride(0) + a;
+            for ( int n = 0; n < num_comp; ++n )
+                scratch_array( i-begin, n ) =
+                    slice_data[ slice_offset + SliceType::vector_length * n ];
```
sslattery: Nice - you must have looked at the communication functions for this.

core/unit_test/tstLinkedCellList.hpp
```diff
@@ -237,6 +236,437 @@ void testLinkedList()
     }
 }
 
+//---------------------------------------------------------------------------//
+void testLinkedListRange()
+{
+    // Make an AoSoA with positions and ijk cell ids.
```
sslattery: For the construction part of the test it looks like this is reused from the previous test - maybe we can break this into a separate function for code reuse.

core/unit_test/tstLinkedCellList.hpp
```diff
@@ -237,6 +236,437 @@ void testLinkedList()
...
+                }
+            }
+        }
+    }
+}
+
+//---------------------------------------------------------------------------//
+void testLinkedListSlice()
+{
+    // Make an AoSoA with positions and ijk cell ids.
```
sslattery: Same here in terms of code reuse for initialization

core/src/Cabana_Sort.hpp
streeve: The original idea for this was to bin all particles with the `LinkedCellList`, permute only local or ghost, then reuse the binning with the `VerletList`. Since then we've decided against that, but that part was already done

core/src/Cabana_Sort.hpp
streeve: Yeah, stealing from `Halo` made this a lot easier

core/unit_test/tstLinkedCellList.hpp
streeve: Agreed. I'll make one function to set up the problem, then reuse throughtout

sslattery:
@streeve also note the CUDA errors in the Jenkins build

core/src/Cabana_LinkedCellList.hpp
streeve: I've left the ``enable_if`` since the compiler wasn't able to distinguish the aosoa and slice variants

core/src/Cabana_Sort.hpp
streeve: Removed

core/unit_test/tstLinkedCellList.hpp
streeve: Right now code is duplicated (especially throughout tests for Sort); what do you think about adding back test fixtures where it makes sense?

core/src/Cabana_Sort.hpp
```diff
@@ -589,6 +589,75 @@ void permute( const BinningDataType& binning_data,
...
+  \tparm SliceType The slice type.
+
+  \param binning_data The binning data.
+
+  \param slice The slice to permute.
+ */
+template<class BinningDataType, class SliceType>
+void permute( const BinningDataType& binning_data,
+              SliceType& slice,
+              typename std::enable_if<(is_binning_data<BinningDataType>::value &&
```
sslattery: @dalg we need the `is_slice` for the compiler to pick the correct implementation. If we move `is_binning_data` to a `static_assert` will compile times improve?

sslattery: talked with @dalg24 on this - we will resolve elsewhere

PR #116, opened 2019-04-04
CabanaCore_config: expose backend defines

junghans:


core/src/CabanaCore_config.hpp.cmakein
```diff
@@ -7,4 +7,12 @@
 
 #cmakedefine01 Cabana_ENABLE_MPI
 
+#cmakedefine01 Cabana_ENABLE_Serial
+
+#cmakedefine01 Cabana_ENABLE_Pthread
+
+#cmakedefine01 Cabana_ENABLE_OpenMP
+
+#cmakedefine01 Cabana_ENABLE_Cuda
```
dalg24: I actually wonder if we should do
```diff
-  #cmakedefine01 Cabana_ENABLE_Whatever
+  #cmakedefine Cabana_ENABLE_Whatever
```
instead because there is change users will do
```C++
#if defined( Cabana_ENABLE_Whatever )
```

sslattery: good point

junghans: Good point, @sslattery what do you think?

junghans: It seems we already use it this way anyhow: https://github.com/ECP-copa/Cabana/blob/master/core/src/Cabana_Core.hpp#L32

junghans: Anyhow I prefer `#ifdef FOO` over `#if ( FOO )`.

junghans: So any preferences?

dalg24: I would change everything to `#cmakedefine Foo` and `#if defined( FOO )`

junghans: Done!

PR #122, opened 2019-05-07
Mirror View

sslattery:
Adding mirror view construction for AoSoA. Allows one to create a mirror in the specified memory space. This was needed in the decorators to efficiently mirror data between host and device for MPI communication when GPU-aware is not used. In this case (and many others) we want a mirror but don't want to immediately do a copy. I use this in the upcoming communication performance benchmark PR (#123).

### DIFF
--- a/core/src/Cabana_DeepCopy.hpp
+++ b/core/src/Cabana_DeepCopy.hpp
@@ -141,6 +141,58 @@ namespace Experimental
 template<class Space, class SrcAoSoA>
 inline
 SrcAoSoA
+create_mirror_view(
+    const Space&,
+    const SrcAoSoA& src,
+    typename std::enable_if<(std::is_same<typename SrcAoSoA::memory_space,
+                             typename Space::memory_space>::value)>::type* = 0 )
+{
+    static_assert( is_aosoa<SrcAoSoA>::value,
+                   "create_mirror_view() requires an AoSoA" );
+    return src;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief Create a mirror on the host of the given AoSoA in the given memory
+  space. Different space specialization allocates a new AoSoA.
+
+  \note Memory allocation will only occur if the requested mirror
+  memory space is different from that of the input AoSoA. If they are the
+  same, the original AoSoA (e.g. a view of that AoSoA) is returned.
+ */
+template<class Space, class SrcAoSoA>
+inline
+AoSoA<typename SrcAoSoA::member_types,
+      typename Space::memory_space,
+      SrcAoSoA::vector_length>
+create_mirror_view(
+    const Space&,
+    const SrcAoSoA& src,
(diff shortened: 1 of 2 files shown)


core/src/Cabana_DeepCopy.hpp
```diff
@@ -141,6 +141,56 @@ namespace Experimental
 template<class Space, class SrcAoSoA>
 inline
 SrcAoSoA
+create_mirror_view(
+    const Space&,
+    const SrcAoSoA& src,
+    typename std::enable_if<(is_aosoa<SrcAoSoA>::value &&
```
dalg24: `is_aosoa<SrcAoSoA>::value` is superfluous.  I would drop it.  You may add a static assert if you like.

core/unit_test/tstDeepCopy.hpp
```diff
@@ -176,41 +176,83 @@ void testMirror()
...
+        Cabana::deep_copy( same_space_copy, aosoa );
+        EXPECT_EQ( same_space_copy.size(), aosoa.size() );
+        bool ssc_same_ms =
+            std::is_same<TEST_MEMSPACE,
+                         decltype(same_space_copy)::memory_space>::value;
+        EXPECT_TRUE( ssc_same_ms );
+        bool ssc_same_mt =
+            std::is_same<DataTypes,
+                         decltype(same_space_copy)::member_types>::value;
+        EXPECT_TRUE( ssc_same_mt );
```
dalg24: The last two assertions can be checked at compile-time.

core/unit_test/tstDeepCopy.hpp
```diff
@@ -176,41 +176,83 @@ void testMirror()
...
-    EXPECT_TRUE( hsc_same_ms );
-    bool hsc_same_mt =
-        std::is_same<DataTypes,
-                     decltype(host_space_copy)::member_types>::value;
-    EXPECT_TRUE( hsc_same_mt );
+    // Create a mirror with the same memory space and copy separately.
+    {
+        auto same_space_copy = Cabana::Experimental::create_mirror_view(
+            TEST_MEMSPACE(), aosoa );
+        Cabana::deep_copy( same_space_copy, aosoa );
```
dalg24: Avoid duplicating the blocks.  Encapsulate the "create mirror view and copy" and call the test for the two variants.

core/src/Cabana_DeepCopy.hpp
sslattery: fixed

core/unit_test/tstDeepCopy.hpp
sslattery: fixed

core/unit_test/tstDeepCopy.hpp
sslattery: fixed

sslattery: well - I did a variant of what you suggested that reused code

PR #123, opened 2019-05-07
Comm Performance Benchmark

sslattery:
This benchmark allows for broad study of Cabana communication algorithms on machines like Summit and will give an easy platform to test performance optimizations. Some features:

- Both halo exchange and migration via the distributor are tested
- A cartesian communication space with periodic boundaries is generated. Each rank communicates with its 26 neighbor ranks in the cartesian space
- Each computation is done a specified number of times so a min, max, and average timing can be generated to monitor jitter
- The data communicated is arbitrary (i.e. not necessarily particle or mesh data) and therefore the conclusions from this can be applicable to a wide variety of applications
- A parameter sweep through the amount of data communicated to each neighbor is done to understand the effect of message size on performance.
- Both fast and slow construction methods for the distributor and halo are tested
- Decorators allow one to do GPU-GPU, CPU-CPU, or GPU-CPU-CPU-GPU as the communication pathway
- This code has been run on 1/4 of Summit 

This PR depends on #122 being merged first.

### DIFF
--- a/core/example/benchmark/CMakeLists.txt
+++ b/core/example/benchmark/CMakeLists.txt
@@ -8,3 +8,9 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "([xX]86|[aA][mM][dD]64)")
   add_executable(PeakFlops Cabana_peakflops.cpp)
   target_link_libraries(PeakFlops cabanacore)
 endif()
+
+if(Cabana_ENABLE_MPI AND Cabana_ENABLE_Cuda AND (Cabana_ENABLE_Serial OR Cabana_ENABLE_OpenMP OR Cabana_ENABLE_Pthread) )
+  add_executable(CommPerformance Cabana_CommPerformance.cpp)
+  target_link_libraries(CommPerformance cabanacore)
+
+endif()
--- a/core/example/benchmark/Cabana_CommPerformance.cpp
+++ b/core/example/benchmark/Cabana_CommPerformance.cpp
@@ -1,0 +1,592 @@
+/****************************************************************************
+ * Copyright (c) 2018-2019 by the Cabana authors                            *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#include <Cabana_Core.hpp>
+
+#include <Kokkos_Core.hpp>
+
+#include <iostream>
+#include <ratio>
+#include <chrono>
+#include <vector>
+#include <string>
+#include <algorithm>
+#include <numeric>
+#include <fstream>
+
+#include <mpi.h>
+
+//---------------------------------------------------------------------------//
+// Parallel timer. Carries multiple data points (the independent variable in
+// the parameter sweep) for each timer to allow for parametric sweeps. Each
+// timer can do multiple runs over each data point in the parameter sweep. The
+// name of the data point and its values can then be injected into the output
+// table.
+class ParallelTimer
+{
+  public:
+
+    // Create the timer.
+    ParallelTimer( MPI_Comm comm,
+                   const std::string& name,
+                   const int num_data )
+        : _comm( comm )
+        , _name( name )
+        , _starts( num_data )
+        , _data( num_data )
+        , _is_stopped( num_data, true )
+    {}
+
+    // Start the timer for the given data point. It is assumed that this
+    // function is called collectively across all ranks in the communicator.
+    void start( const int data_point )
+    {
+        if ( !_is_stopped[data_point] )
+            throw std::logic_error( "attempted to start a running timer" );
+        auto now = std::chrono::high_resolution_clock::now();
+        _starts[data_point].push_back( now );
+        _is_stopped[data_point] = false;
+    }
+
+    // Stop the timer at the given data point. It is assumed that this
+    // function is called collectively across all ranks in the communicator.
+    void stop( const int data_point )
+    {
+        if ( _is_stopped[data_point] )
+            throw std::logic_error( "attempted to stop a stopped timer" );
+        auto now = std::chrono::high_resolution_clock::now();
+        std::chrono::duration<double,std::micro> fp_micro =
+            now - _starts[data_point].back();
+        _data[data_point].push_back( fp_micro.count() );
+        _is_stopped[data_point] = true;
+    }
+
+    // Write timer results on rank 0. Provide the values of the data points so
+    // they can be injected into the table. This function does collective
+    // communication.
+    template<typename Scalar>
+    void outputResults( std::ostream& stream,
+                        const std::string& data_point_name,
+                        const std::vector<Scalar>& data_point_vals ) const
+    {
+        // Get comm rank;
+        int comm_rank;
+        MPI_Comm_rank( _comm, &comm_rank );
+
+        // Get comm size;
+        int comm_size;
+        MPI_Comm_size( _comm, &comm_size );
+
+        // Write the data header.
+        if ( 0 == comm_rank )
+        {
+            stream << "\n";
+            stream << _name << "\n";
+            stream << "num_rank " << data_point_name << " min max ave" << "\n";
+        }
+
+        // Write out each data point
+        for ( std::size_t n = 0; n < _data.size(); ++n )
+        {
+            if ( !_is_stopped[n] )
+                throw std::logic_error( "attempted to output from a running timer" );
+
+            // Compute the minimum.
+            double local_min = *std::min_element( _data[n].begin(), _data[n].end() );
+            double global_min = 0.0;
+            MPI_Reduce( &local_min, &global_min, 1, MPI_DOUBLE, MPI_MIN, 0, _comm );
+
+            // Compute the maximum.
+            double local_max = *std::max_element( _data[n].begin(), _data[n].end() );
+            double global_max = 0.0;
+            MPI_Reduce( &local_max, &global_max, 1, MPI_DOUBLE, MPI_MAX, 0, _comm );
+
+            // Compute the average.
+            double local_sum = std::accumulate( _data[n].begin(), _data[n].end(),  0.0 );
+            double average = 0.0;
+            MPI_Reduce( &local_sum, &average, 1, MPI_DOUBLE, MPI_SUM, 0, _comm );
+            average /= _data[n].size() * comm_size;
+
+            // Output on rank 0.
+            if ( 0 == comm_rank )
+            {
+                stream << comm_size << " "
+                       << data_point_vals[n] << " "
+                       << global_min << " "
+                       << global_max << " "
+                       << average << "\n";
+            }
+        }
+    }
+
+  private:
+
+    MPI_Comm _comm;
+    std::string _name;
+    std::vector<std::size_t> _byte_sizes;
+    std::vector<std::vector<std::chrono::high_resolution_clock::time_point> > _starts;
+    std::vector<std::vector<double> > _data;
+    std::vector<bool> _is_stopped;
+};
+
+//---------------------------------------------------------------------------//
+// Performance test.
+// Data device type is where the data to be communicated lives.
+// Comm device type is the device we want to use for communication.
+template<class DataDevice, class CommDevice>
+void performanceTest( std::ostream& stream,
+                      const std::size_t num_particle,
+                      const std::string& test_prefix )
+{
+    // PROBLEM SETUP
+    // -------------
+
+    // Get comm world.
+    auto comm = MPI_COMM_WORLD;
+
+    // Get comm size;
+    int comm_size;
+    MPI_Comm_size( comm, &comm_size );
+
+    // Partition the problem in 3 dimensions.
+    const int space_dim = 3;
+    std::vector<int> ranks_per_dim( space_dim );
+    MPI_Dims_create( comm_size, space_dim, ranks_per_dim.data() );
+
+    // Generate a communicator with a cartesian topology and periodic
+    // boundaries.
+    std::vector<int> periodic_dims( space_dim, 1 );
+    MPI_Comm cart_comm;
+    int reorder_cart_ranks = 1;
+    MPI_Cart_create( comm,
+                     space_dim,
+                     ranks_per_dim.data(),
+                     periodic_dims.data(),
+                     reorder_cart_ranks,
+                     &cart_comm );
+
+    // Get the Cartesian topology index of this rank.
+    int linear_rank;
+    MPI_Comm_rank( cart_comm, &linear_rank );
+    std::vector<int> cart_rank( space_dim );
+    MPI_Cart_coords( cart_comm, linear_rank, space_dim, cart_rank.data() );
+
+    // Compute the 27 ranks this rank communicates with. This includes self
+    // communication. Put self communication first.
+    std::vector<int> neighbor_ranks;
+    neighbor_ranks.push_back( linear_rank );
+    for ( int k = -1; k < 2; ++k )
+        for ( int j = -1; j < 2; ++j )
+            for ( int i = -1; i < 2; ++i )
+                if ( !(i==0 && j==0 && k==0) )
+                {
+                    std::vector<int> ncr = { cart_rank[0] + i,
+                                             cart_rank[1] + j,
+                                             cart_rank[2] + k };
+                    int nr;
+                    MPI_Cart_rank( cart_comm, ncr.data(), &nr );
+                    neighbor_ranks.push_back( nr );
+                }
+
+    // Make a unique list of neighbors that will serve as the communication
+    // topology.
+    std::vector<int> unique_neighbors( neighbor_ranks );
+    std::sort( unique_neighbors.begin(), unique_neighbors.end() );
+    auto unique_end =
+        std::unique( unique_neighbors.begin(), unique_neighbors.end() );
+    unique_neighbors.resize( std::distance(unique_neighbors.begin(),unique_end) );
+
+    // Define the aosoa.
+    using comm_memory_space = typename CommDevice::memory_space;
+    using data_memory_space = typename DataDevice::memory_space;
+    using member_types = Cabana::MemberTypes<double[3],
+                                             double[3],
+                                             double,
+                                             int>;
+    using aosoa_type = Cabana::AoSoA<member_types,data_memory_space>;
+
+    // Get the byte size of a particle
+    using tuple_type = Cabana::Tuple<member_types>;
+    int bytes_per_particle = sizeof( tuple_type );
+
+    // Number of runs in the test loops.
+    int num_run = 10;
+
+    // Fraction of particles on each rank that will be communicated to the
+    // neighbors. We will sweep through these fractions to get an indicator of
+    // performance as a function of message size.
+    std::vector<double> comm_fraction = { 0.0001,
+                                          0.001,
+                                          0.005,
+                                          0.01,
+                                          0.05,
+                                          0.10,
+                                          0.25,
+                                          0.5 };
+    int num_fraction = comm_fraction.size();
+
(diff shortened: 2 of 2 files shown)


sslattery:
I have a bunch of results from Summit but am still processing/understanding them. Will post some here when I am ready.

sslattery:
This is ready for review.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,573 @@
...
+    std::vector<int> ranks_per_dim( 3 );
+    MPI_Dims_create( comm_size, 3, ranks_per_dim.data() );
+
+    // Generate a communicator with a cartesian topology and periodic
+    // boundaries.
+    std::vector<int> periodic_dims( 3, 1 );
+    MPI_Comm cart_comm;
+    int reorder_cart_ranks = 1;
+    MPI_Cart_create( comm,
+                     3,
```
rfbird: Shall we dry up some of these hard coded `dim=3`'s?

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,573 @@
...
+                if ( !(i==0 && j==0 && k==0) )
+                {
+                    std::vector<int> ncr = { cart_rank[0] + i,
+                                             cart_rank[1] + j,
+                                             cart_rank[2] + k };
+                    int nr;
+                    MPI_Cart_rank( cart_comm, ncr.data(), &nr );
+                    neighbor_ranks.push_back( nr );
+                }
+    std::sort( neighbor_ranks.begin(), neighbor_ranks.end() );
```
rfbird: Is it odd to seemingly do these in a fairly specific order and then call sort? On my first pass I thought when you said `//Put self communication first.` I assumed it was convention, but it's just convenience?

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+    // Fraction of particles on each rank that will be communicated to the
+    // neighbors. We will sweep through these fractions to get an indicator of
+    // performance as a function of message size.
+    std::vector<double> comm_fraction = { 0.0001,
+                                          0.001,
+                                          0.005,
+                                          0.01,
+                                          0.05,
+                                          0.10,
```
rfbird: Should we add something between `0.1` and `0.5`? For PIC it's not unbelievable to get something a little higher than `0.1` (although between ~3-10% if typical)

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,573 @@
...
+        int send_per_neighbor = num_send / 26;
+        num_send = send_per_neighbor * 26;
+        int num_stay = num_particle - num_send;
+        comm_bytes[fraction] = send_per_neighbor * bytes_per_particle;
+        Kokkos::View<int*,Kokkos::HostSpace>
+            export_ranks_host( "export_ranks", num_particle );
+        for ( int p = 0; p < num_stay; ++p )
+            export_ranks_host(p) = neighbor_ranks[0];
+        for ( int n = 0; n < 26; ++n )
+            for ( int p = 0; p < send_per_neighbor; ++p )
```
rfbird: Do you think it would be cleaner if we had some `{}` around these blocks? I feel like especially for liens that are broken because of length (as below) it gets a bit scary

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,573 @@
...
+    ParallelTimer distributor_slice_migrate(
+        comm, test_prefix+"distributor_slice_migrate", num_fraction );
+
+    // Loop over comm fractions.
+    for ( int fraction = 0; fraction < num_fraction; ++fraction )
+    {
+        // Create the migrate distribution in the data memory space. This is
+        // where it would likely be created.
+        int num_send = num_particle * comm_fraction[fraction];
+        int send_per_neighbor = num_send / 26;
```
rfbird: This `/26` then `*26` strikes me as a little odd ? :p

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,573 @@
...
+            Cabana::Distributor<comm_memory_space> distributor_fast(
+                comm, comm_export_ranks, neighbor_ranks );
+            distributor_fast_create.stop(fraction);
+
+            // Create a distributor using the slow construction method.
+            distributor_slow_create.start(fraction);
+            comm_export_ranks =
+                Kokkos::create_mirror_view_and_copy(
+                    comm_memory_space(), export_ranks );
+            Cabana::Distributor<comm_memory_space> distributor_slow(
```
rfbird: In an ideal world this wouldn't be `slow`, right? More.. `simple` or something?

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: this is intended to be a 3d test I guess - would you rather see this as an integer defined externally from the function?

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: oooooh - you found a bug!

core/example/benchmark/Cabana_CommPerformance.cpp
rfbird: `const int DIMENSION  = 3;` or whatever would be fine for me

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: we could add 0.25 here. You really want to have a realistic number of particles per GPU for this test. You need to test both message size as well as performance of the other kernels.

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: I can add those

core/example/benchmark/Cabana_CommPerformance.cpp
rfbird: 👐 👐

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: That was a way to cheat and get a number that was evenly divisible by 26

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: done

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: this should be fixed - I made a separate list of unique neighbor ranks that define the pre-computed communication topology.

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: done

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: done

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: no it's actually slower because you need to compute the communication topology rather than having received it from the user. maybe the better terminology is "general"

sslattery: changed "slow" -> "general"

core/example/benchmark/Cabana_CommPerformance.cpp
dalg24: Add a comment to capture the intent

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+            comm_src_particles =
+                Cabana::Experimental::create_mirror_view_and_copy(
+                    comm_memory_space(), src_particles );
+            comm_dst_particles =
+                Cabana::Experimental::create_mirror_view(
+                    comm_memory_space(), dst_particles );
+
+            auto s0 = comm_src_particles.template slice<0>();
+            auto d0 = comm_dst_particles.template slice<0>();
```
dalg24: Probably should be using the free function here.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+}
+
+//---------------------------------------------------------------------------//
+// main
+int main( int argc, char* argv[] )
+{
+    // Initialize environment
+    MPI_Init( &argc, &argv );
+    Kokkos::initialize( argc, argv );
+
```
dalg24: Exit here if `argc < 3` and consider printing usage

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+    // Do everything on the GPU with UVM GPU memory.
+    performanceTest<Kokkos::CudaUVMSpace,Kokkos::CudaUVMSpace>(
+        file, num_particle, "cudauvm_cudauvm_" );
+
+    // Close the output file on rank 0.
+    if ( 0 == comm_rank )
+        file.close();
+
+    // Barier before finalizing.
+    MPI_Barrier( MPI_COMM_WORLD );
```
dalg24: Why so?

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: I have added a comment discussing intent

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+    // Get the number of timer runs.
+    int numRun() const
+    {
+        return _data[0].size();
+    }
+
+    // Write timer results on rank 0. Provide the data points so they can be
+    // injected into the table.
+    template<typename Scalar>
+    void outputResults( std::fstream& file,
```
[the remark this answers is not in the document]
sslattery: replaces `fstream` with `ostream`

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+    // Get the number of timer data points.
+    int numDataPoint() const
+    {
+        return _data.size();
+    }
+
+    // Get the number of timer runs.
+    int numRun() const
+    {
+        return _data[0].size();
```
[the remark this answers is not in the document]
sslattery: good point. I changed this function to return the number of runs for a given data point as it may be arbitrary.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+        MPI_Comm_rank( _comm, &comm_rank );
+
+        // Get comm size;
+        int comm_size;
+        MPI_Comm_size( _comm, &comm_size );
+
+        // Write the data header.
+        if ( 0 == comm_rank )
+        {
+            file << std::endl;
```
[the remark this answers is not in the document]
sslattery: replaced use of `std::endl` with `"\n"`

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+        // Write out each data point
+        for ( std::size_t n = 0; n < _data.size(); ++n )
+        {
+            if ( !_is_stopped[n] )
+                throw std::logic_error( "attempted to output from a running timer" );
+
+            // Compute the minimum.
+            double local_min = (double) std::min_element(
+                _data[n].begin(), _data[n].end() )->count();
```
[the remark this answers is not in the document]
sslattery: The min and max values should actually be integers. I wanted the results to be in terms of microseconds. I have changed min and max to integers.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+            double global_min = 0.0;
+            MPI_Reduce( &local_min, &global_min, 1, MPI_DOUBLE, MPI_MIN, 0, _comm );
```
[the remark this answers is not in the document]
sslattery: I was going to do in-place but then would have to have some extra logic. The MPI standard says that `MPI_IN_PLACE` can only be used on the root rank for the send buffer argument. So either I make a local value as I have done here or have some Boolean logic based on root rank.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+            double global_max = 0.0;
+            MPI_Reduce( &local_max, &global_max, 1, MPI_DOUBLE, MPI_MAX, 0, _comm );
+
+            // Compute the average.
+            double local_sum = 0.0;
+            for ( auto d : _data[n] )
+                local_sum += (double) d.count();
+            double average = 0.0;
+            MPI_Reduce( &local_sum, &average, 1, MPI_DOUBLE, MPI_SUM, 0, _comm );
+            average /= _data[n].size() * comm_size;
```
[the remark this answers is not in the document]
sslattery: Yes - it assumes the timer is called collectively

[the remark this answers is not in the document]
sslattery: I will make a note of this.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+    // Partition the problem in 3 dimensions.
+    const int space_dim = 3;
+    std::vector<int> ranks_per_dim( space_dim );
+    MPI_Dims_create( comm_size, space_dim, ranks_per_dim.data() );
+
+    // Generate a communicator with a cartesian topology and periodic
+    // boundaries.
+    std::vector<int> periodic_dims( space_dim, 1 );
+    MPI_Comm cart_comm;
+    int reorder_cart_ranks = 1;
```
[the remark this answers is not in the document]
sslattery: I was thinking that it could possibly result in a more optimal layout of processors for the type of communication I am doing in this test (essentially Cartesian grid halo exchange). It would be interesting to turn this off though and see if/how performance changes.

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: Yes - I will use that now that it is in `master`

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: done

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: Seemed like a good idea but probably not necessary. I will remove.

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,586 @@
...
+    {
+        return _data[0].size();
+    }
+
+    // Write timer results on rank 0. Provide the data points so they can be
+    // injected into the table.
+    template<typename Scalar>
+    void outputResults( std::fstream& file,
+                        const std::string& data_point_name,
+                        const std::vector<Scalar>& data_point_vals ) const
```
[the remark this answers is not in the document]
sslattery: Yes - this allows the user to essentially inject another column in the output table corresponding to the independent variable in the parameter sweep. In this case it is the fraction of particles communicated to the neighbors.

core/example/benchmark/Cabana_CommPerformance.cpp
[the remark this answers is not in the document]
dalg24: the C-style cast is superfluous

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,604 @@
...
+            int local_max = (int) std::max_element(
+                _data[n].begin(), _data[n].end() )->count();
+            int global_max = 0.0;
+            MPI_Reduce( &local_max, &global_max, 1, MPI_INT, MPI_MAX, 0, _comm );
+
+            // Compute the average.
+            double local_sum = 0.0;
+            for ( auto d : _data[n] )
+                local_sum += (double) d.count();
+            double average = 0.0;
```
dalg24: Not changing this one to `int`?

core/example/benchmark/Cabana_CommPerformance.cpp
```diff
@@ -0,0 +1,604 @@
...
+    int numDataPoint() const
+    {
+        return _data.size();
+    }
+
+    // Get the number of timer runs for a given data point.
+    int numRun( const int data_point ) const
+    {
+        return _data[data_point].size();
+    }
```
dalg24: I would have dropped it since it is not call anywhere.  Anyway no big deal

core/example/benchmark/CMakeLists.txt
```diff
@@ -8,3 +8,9 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "([xX]86|[aA][mM][dD]64)")
   add_executable(PeakFlops Cabana_peakflops.cpp)
   target_link_libraries(PeakFlops cabanacore)
 endif()
+
+if(Cabana_ENABLE_MPI AND Cabana_ENABLE_Cuda AND (Cabana_ENABLE_Serial OR Cabana_ENABLE_OpenMP OR Cabana_ENABLE_Pthread) )
```
dalg24: I don't get the `AND (Cabana_ENABLE_Serial OR Cabana_ENABLE_OpenMP OR Cabana_ENABLE_Pthread)`

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: I guess here I was thinking an average could/should be represented as floating point because fractions of seconds are possible.

core/example/benchmark/CMakeLists.txt
sslattery: Essentially this should only build if Cuda is on and an execution space is enabled that supports `Kokkos::HostSpace`. Are there configurations of Kokkos possible where `HostSpace` is not supported?

core/example/benchmark/Cabana_CommPerformance.cpp
dalg24: To be clear, I would have done everything floating point durations all the way.  See my previous comment.

core/example/benchmark/CMakeLists.txt
dalg24: You only use the memory space and it will be there even when CUDA only is enabled

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: OK guess I was confused there. I wanted my output units to be in microseconds. Would you have just done the duration cast to microseconds in the implementation but then stored in the class data as double?

core/example/benchmark/Cabana_CommPerformance.cpp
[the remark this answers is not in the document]
sslattery: yes it is - removed

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: removed

core/example/benchmark/Cabana_CommPerformance.cpp
sslattery: fixed per phone discussion

core/example/benchmark/CMakeLists.txt
sslattery: Per phone call we need to do a follow-on PR which templates our classes on device types instead of memory spaces. Then we would refactor this to explicitly need `Cuda` and `OpenMP`. Once we do that then this test will specify device types that have both memory and execution space definitions.

sslattery:
@dalg24 sorry my push made your review go stale - could you re-approve?

PR #124, opened 2019-05-07
User ScatterView

### LINKED ISSUE #82: Use Kokkos::ScatterView where Possible
A number of algorithms including `LinkedCellList` and `VerletList` make extensive use of atomics. In many places it should be possible to replace those atomic operations on a regular `Kokkos::View` to standard operations on a `Kokkos::ScatterView`

sslattery:
There were several instances in the code where atomics were used. Almost all instances of atomics have been replaced with `Kokkos::ScatterView` where possible to enhance performance on non-GPU architectures. There are still several instances of `Kokkos::atomic_fetch_add` in the code but these are needed to implement the current algorithms and therefore they cannot be changed without some algorithmic research.

Closes #82

### DIFF
--- a/core/src/Cabana_CommunicationPlan.hpp
+++ b/core/src/Cabana_CommunicationPlan.hpp
@@ -13,6 +13,7 @@
 #define CABANA_COMMUNICATIONPLAN_HPP
 
 #include <Kokkos_Core.hpp>
+#include <Kokkos_ScatterView.hpp>
 
 #include <mpi.h>
 
@@ -250,12 +251,15 @@ class CommunicationPlan
             num_export_host( _num_export.data(), num_n );
         auto export_counts = Kokkos::create_mirror_view_and_copy(
                 memory_space(), num_export_host );
+        auto export_counts_sv =
+            Kokkos::Experimental::create_scatter_view( export_counts );
         auto count_neighbor_func =
             KOKKOS_LAMBDA( const int i )
             {
+                auto export_counts_data = export_counts_sv.access();
                 for ( int n = 0; n < num_n; ++n )
                     if ( topology(n) == element_export_ranks(i) )
-                        Kokkos::atomic_increment( &export_counts(n) );
+                        export_counts_data(n) += 1;
             };
         Kokkos::RangePolicy<execution_space> count_neighbor_policy(
             0, _num_export_element );
@@ -263,6 +267,7 @@ class CommunicationPlan
                               count_neighbor_policy,
                               count_neighbor_func );
         Kokkos::fence();
+        Kokkos::Experimental::contribute( export_counts, export_counts_sv );
 
         // Copy counts back to the host.
         Kokkos::deep_copy( num_export_host, export_counts );
@@ -357,12 +362,16 @@ class CommunicationPlan
         // Count the number of sends this rank will do to other ranks.
         Kokkos::View<int*,memory_space> neighbor_counts(
             "neighbor_counts", comm_size );
+        auto neighbor_counts_sv =
+            Kokkos::Experimental::create_scatter_view( neighbor_counts );
         auto count_sends_func =
             KOKKOS_LAMBDA( const int i )
             {
                 if ( element_export_ranks(i) >= 0 )
-                    Kokkos::atomic_increment(
-                        &neighbor_counts(element_export_ranks(i)) );
+                {
+                    auto neighbor_counts_data = neighbor_counts_sv.access();
+                    neighbor_counts_data(element_export_ranks(i)) += 1;
+                }
             };
         Kokkos::RangePolicy<execution_space> count_sends_policy(
             0, _num_export_element );
@@ -370,6 +379,7 @@ class CommunicationPlan
                               count_sends_policy,
                               count_sends_func );
         Kokkos::fence();
+        Kokkos::Experimental::contribute( neighbor_counts, neighbor_counts_sv );
 
         // Copy the counts to the host.
         auto neighbor_counts_host = Kokkos::create_mirror_view_and_copy(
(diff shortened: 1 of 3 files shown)


core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -250,19 +251,23 @@ class CommunicationPlan
...
+        auto export_counts_sv =
+            Kokkos::Experimental::create_scatter_view( export_counts );
         auto count_neighbor_func =
             KOKKOS_LAMBDA( const int i )
             {
+                auto export_counts_data = export_counts_sv.access();
                 for ( int n = 0; n < num_n; ++n )
                     if ( topology(n) == element_export_ranks(i) )
-                        Kokkos::atomic_increment( &export_counts(n) );
+                        export_counts_data(n) += 1;
```
dalg24: Does it support prefix increment?

sslattery: yes

sslattery:
I didn't profile but for past codes I have profiled I have never had a case where the duplication from `ScatterView` didn't perform better than using atomics on a CPU. On a GPU it defaults to atomics which is our preferred behavior.

sslattery:
Ok here are some speedups on Summit on communication plan construction where I expected them. The speedup listed is the fraction over 1 that was improved. So `0.53` means a 53% speedup and `-0.013` would mean a 1.3% slow down:

rfbird:
> I didn't profile but for past codes I have profiled I have never had a case where the duplication from `ScatterView` didn't perform better than using atomics on a CPU. On a GPU it defaults to atomics which is our preferred behavior.

FWIW I've seen cases where our hand written implementation is definitely faster than `ScatterView`, but I agree I fully expect `ScatterView` to be faster than atomics on CPU for any sane use case

sslattery:
yeah so actually after investigating the duplication from scatter view *was* slower in the halo scatter - in that case there is often so few collisions that the extra duplication became an overhead. therefore I removed it.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -250,19 +251,23 @@ class CommunicationPlan
             num_export_host( _num_export.data(), num_n );
         auto export_counts = Kokkos::create_mirror_view_and_copy(
                 memory_space(), num_export_host );
+        auto export_counts_sv =
+            Kokkos::Experimental::create_scatter_view( export_counts );
```
rfbird: Are we worried about always having to create the views? Is it worth considering providing a fixed size allocation for us by experts with consistent/regular data patterns/sends?

rfbird: I guess then we have to be more careful about resetting the data too

sslattery: I guess I'm not sure what you mean here.

sslattery:
Yes - if we find places where, for example, the Halo scatter operation has tons of collisions and could use duplication on the CPU then we could add back in `ScatterView` and develop some collision metric based on the communication plan (because we know all collisions once the plan is generated) to pick the implementation at run time.

core/src/Cabana_CommunicationPlan.hpp
rfbird: Sorry, I should have been clearer.  This makes a `ScatterView` per function invocation, right? And then throws it away at the end? 

For regular/structured apps, does this imply memory allocation/deallocation of the same size every timestep? Or is this somehow only ever done once and I'm missing a detail?

guangyechen:
@rfbird when you said that handwritten one faster than scatter_view, is it for cpu or gpu, or both?

rfbird:
> @rfbird when you said that handwritten one faster than scatter_view, is it for cpu or gpu, or both?

Only CPU (there basically no overhead to ScatterView on GPU)

core/src/Cabana_CommunicationPlan.hpp
sslattery: I believe it would imply allocation/deallocation of the replicated views if replication is the scatter method of choice. Because we don't have information about the slice the user wants to scatter a priori I think it would be difficult to save the state of a scatter view somewhere unless we developed some type of container approach where the user would register the slices they are scattering.

guangyechen:
May the overhead of scatter_view be creating/destroying temporary arrays on the fly?

core/src/Cabana_CommunicationPlan.hpp
rfbird: Yea, that's what I'm thinking. A second advance API where they provide more information. Something for down the road perhaps.

sslattery:
Yes the overhead would be from that as well as the recombination at the end.

rfbird:
> May the overhead of scatter_view be creating/destroying temporary arrays on the fly?

I think it's nothing to be overly concerned about right now. Two things:

1) My comparison wasn't apple to apples, so the data is only so-so
2) Kokkos actually has a performance test for this, which means they're aware and happy with the status of the implementation: https://github.com/kokkos/kokkos/blob/master/containers/performance_tests/TestScatterView.hpp

stanmoore1:
I've tested ScatterView in ExaMiniMD, see https://github.com/ECP-copa/ExaMiniMD/pull/21. There was little difference between creating temporary ScatterViews vs using a persistent allocation. We also compared performance to a hand-coded version, and for 1D arrays it was close. There may be more optimizations needed for 2D arrays though.

stanmoore1:
FYI, see Christian's comment at the end of https://github.com/kokkos/kokkos/issues/1390.

rfbird:
> FYI, see Christian's comment at the end of [kokkos/kokkos#1390](https://github.com/kokkos/kokkos/issues/1390).

Good point. overall I'm happy to rely on this (as I also do for a LANL production app..), hopefully it will only get better! If we see it really causing an issue I'll do something similar to Christians' comments in #1390 and file a PR..

sslattery:
Agreed - and we have our own performance tests to track these changes (although not automated yet). I'm very happy with a 2x speedup for the communication stuff.

PR #125, opened 2019-05-08
Deep Copy Additions

sslattery:
This PR adds a number of user-requested features for Deep Copy:

- The ability to deep copy between two slices. The slices must have the same element types for this to work. This is checked at compile time.
- The ability to assign a scalar value to all elements of a slice. Good for initialization or reinitialization. This is identical to the Kokkos view scalar assignment capability.
- The ability to assign a tuple to each element of an AoSoA. This is the analog to slice scalar assignment.

The tutorial has also been updated to include these features. I did some slight cleaning so the new create mirror view code can be used to simplify the deep copy implementations.

### DIFF
--- a/core/unit_test/tstDeepCopy.hpp
+++ b/core/unit_test/tstDeepCopy.hpp
@@ -122,6 +122,20 @@ void testDeepCopy()
 
     // Check values.
     checkDataMembers( dst_aosoa, fval, dval, ival, dim_1, dim_2, dim_3 );
+
+    // Create a second AoSoA and deep copy by slice.
+    DstAoSoA_t dst_aosoa_2( num_data );
+    auto dst_slice_0 = dst_aosoa_2.template slice<0>();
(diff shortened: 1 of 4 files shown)


core/unit_test/tstDeepCopy.hpp
```diff
@@ -229,6 +243,57 @@ void testMirror()
...
+    // Declare data types.
+    using DataTypes = Cabana::MemberTypes<float[2],int>;
+
+    // Create an AoSoA in the test memory space.
+    int num_data = 423;
+    Cabana::AoSoA<DataTypes,TEST_MEMSPACE> aosoa( num_data );
+
+    // Assign every tuple in the AoSoA to the same value.
+    float fval = 3.2;
+    float ival = 1;
```
streeve: int ival

sslattery: good catch - fixed

PR #127, opened 2019-05-09
Adding labels to the aosoa and slice

### LINKED ISSUE #92: Add a Label to AoSoA and Slice Objects

sslattery:
Adds labels to the AoSoA and Slice for profiling and debugging purposes. The AoSoA label will also end up in the Kokkos View managing that memory and therefore be present in the Kokkos allocation tracker. The slice label is only assigned to the slice because there is no allocation tracker associated with unmanaged views. However, these labels can still be used to compose output messages and other information. The mirror creation is also updated to add a mirror suffix to labels of newly create AoSoA data structures.

Closes #92

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -17,6 +17,7 @@
 #include <Cabana_Tuple.hpp>
 #include <Cabana_Types.hpp>
 #include <Cabana_SoA.hpp>
+#include <Cabana_Macros.hpp>
 #include <impl/Cabana_Index.hpp>
 #include <impl/Cabana_PerformanceTraits.hpp>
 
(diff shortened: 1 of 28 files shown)


core/src/Cabana_AoSoA.hpp
```diff
@@ -126,27 +126,44 @@ class AoSoA
     /*!
       \brief Default constructor.
 
+      \param label An optional label for the data structure.
+
       The container size is zero and no memory is allocated.
     */
-    AoSoA()
+    explicit AoSoA( const std::string& label = "" )
```
sslattery: remove explicit

core/src/Cabana_AoSoA.hpp
```diff
@@ -126,27 +126,44 @@ class AoSoA
...
 
     /*!
       \brief Allocate a container with n tuples.
 
       \param n The number of tuples in the container.
+
+      \param label An optional label for the data structure.
     */
-    explicit AoSoA( const int n )
+    AoSoA( const int n, const std::string label = "" )
```
sslattery: Change the order of the parameters here to follow kokkos conventions. Add back the original single size constructor for backwards compatibility and deprecate it.

streeve:
Do you want to wait to update examples with labels or add them now?

sslattery:
@streeve good call - I'll update the tutorials to not use the deprecated functions.

sslattery:
@streeve I updated the tests and examples so the deprecated constructor is no longer used.

PR #128, opened 2019-05-09
Unmanaged AoSoA

### LINKED ISSUE #114: Add Unmanaged AoSoA
In our Fortran work it seems that the ability to use Fortran-allocated memory with Cabana would greatly simplify the programming model. We can do this by adding an unmanaged constructor to the AoSoA where the user provides a pointer to an array of SoAs, the number of SoAs, and the size. 

In the implementation we will need to do 2 things:
1. Add state to the AoSoA to indicate whether or not it is unmanaged and to disallow unmanaged containers to do memory operations
2. Update the container such that the Kokkos View containing the AoSoA data knows that this data is unmanaged. Because this requires a template change to the Kokkos View it may be that  an additional memory traits template parameter is needed to trigger this.

sslattery:
Adding a memory traits template parameter to the AoSoA to allow for the usage of unmanaged memory. Although this could be useful in many places I see this as a critical piece of infrastructure for Fortran support. Some notes:

- The memory traits parameter is the last parameter in the template list and has a default argument equivalent the the Kokkos default argument (basic managed memory).
- These traits are propagated down to the Kokkos view that manages the AoSoA data
- If a user wants unmanaged memory they need to list all 4 template parameters. I like this because they have to specify vector length which means they will need to sync up their native data structure definition with the AoSoA definition explicitly.
- Operations that modify memory (e.g. resize and reserve) are explicitly not allowed with unmanaged memory via `static_assert`
- Use of slices is unaffected - they already used unmanaged memory anyway

Closes #114

### DIFF
--- a/core/unit_test/tstAoSoA.hpp
+++ b/core/unit_test/tstAoSoA.hpp
@@ -475,6 +475,94 @@ void testAccess()
 }
 
 //---------------------------------------------------------------------------//
+// Manually defined SoA.
+template<int VLEN,int D1, int D2, int D3>
+struct MySoA
+{
+    float m0[D1][D2][D3][VLEN];
+    int m1[VLEN];
+    double m2[D1][VLEN];
+    double m3[D1][D2][VLEN];
+};
+
+// Test an unmanaged AoSoA.
+void testUnmanaged()
+{
+    // Manually set the inner array size.
+    const int vector_length = 16;
+
+    // Data dimensions.
+    const int dim_1 = 3;
+    const int dim_2 = 2;
+    const int dim_3 = 4;
+
+    // Declare data types that are equivalent to the user defined struct.
+    using DataTypes =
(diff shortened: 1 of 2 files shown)


core/unit_test/tstAoSoA.hpp
```diff
@@ -474,6 +474,94 @@ void testAccess()
...
+                            int,
+                            double[dim_1],
+                            double[dim_1][dim_2]
+                            >;
+
+    // Allocate an AoSoA manually.
+    using soa_type = MySoA<vector_length,dim_1,dim_2,dim_3>;
+    int num_soa = 3;
+    int size = 35;
+    Kokkos::View<soa_type*,TEST_MEMSPACE> user_data( "user_aosoa", 3 );
```
rfbird: Would this be a better example if it came from totally outside of Kokkos (a user made array)?

PR #133, opened 2019-05-16
Add default constructor for VerletList

streeve:
This will now match LinkedCellList

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -530,8 +530,14 @@ class VerletList
     VerletListData<MemorySpace,LayoutTag> _data;
 
     /*!
-      \brief Given a list of particle positions and a neighborhood radius calculate
-      the neighbor list.
+      \brief Default constructor.
+    */
+    VerletList()
+    {}
+
+    /*!
+      \brief VerletList constructor. Given a list of particle positions and
+      a neighborhood radius calculate the neighbor list.
 
       \param x The slice containing the particle positions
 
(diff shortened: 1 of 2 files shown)


core/unit_test/tstNeighborList.hpp
```diff
@@ -420,8 +420,11 @@ void testVerletListFull()
     double grid_min[3] = { box_min, box_min, box_min };
     double grid_max[3] = { box_max, box_max, box_max };
     Cabana::VerletList<TEST_MEMSPACE,Cabana::FullNeighborTag,LayoutTag>
-        nlist( Cabana::slice<0>(aosoa), 0, aosoa.size(),
-               test_radius, cell_size_ratio, grid_min, grid_max );
+        nlist_full( Cabana::slice<0>(aosoa), 0, aosoa.size(),
+                    test_radius, cell_size_ratio, grid_min, grid_max );
+
+    Cabana::VerletList<TEST_MEMSPACE,Cabana::FullNeighborTag,LayoutTag>
+        nlist = nlist_full;
```
dalg24: What is the point of making a copy?

core/src/Cabana_VerletList.hpp
```diff
@@ -530,8 +530,14 @@ class VerletList
     VerletListData<MemorySpace,LayoutTag> _data;
 
     /*!
-      \brief Given a list of particle positions and a neighborhood radius calculate
-      the neighbor list.
+      \brief Default constructor.
+    */
+    VerletList()
+    {}
```
dalg24: Not called anywhere

core/unit_test/tstNeighborList.hpp
sslattery: To check the assignment operator and ensure the default function was behaving as expected.

core/src/Cabana_VerletList.hpp
sslattery: True - we should technically update the code on line 427 to only create the neighbor list and then do the assignment on a separate line.

core/unit_test/tstNeighborList.hpp
dalg24: Default function?

PR #136, opened 2019-05-22
cmake: use FFTW3 imported target

junghans:
Use FFTW3 a little more like we use KOKKOS.


cmake/FindFFTW3.cmake
```diff
@@ -37,10 +34,23 @@ find_package_handle_standard_args(FFTW3 DEFAULT_MSG FFTW3_LIBRARY FFTW3_INCLUDE_
...
   #adding MATH_LIBRARIES here to allow static libs, this does not harm us as we are anyway using it
-  check_library_exists("${FFTW3_LIBRARIES};${MATH_LIBRARIES}" fftw_plan_r2r_1d "" FOUND_FFTW_PLAN)
+  check_library_exists("${FFTW3_LIBRARY};${MATH_LIBRARIES}" fftw_plan_r2r_1d "" FOUND_FFTW_PLAN)
   if(NOT FOUND_FFTW_PLAN)
     message(FATAL_ERROR "Could not find fftw_plan_r2r_1d in ${FFTW3_LIBRARY}, take a look at the error message in ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log to find out what was going wrong. If you are using a static lib (.a) make sure you have specified all dependencies of fftw3 in FFTW3_LIBRARY by hand (i.e. -DFFTW3_LIBRARY='/path/to/libfftw3.so;/path/to/libm.so') !")
   endif(NOT FOUND_FFTW_PLAN)
 endif (FFTW3_FOUND)
 
+# Copy the results to the output variables and target.
+if(FFTW3_FOUND)
```
dalg24: Why do we put this in a separate if-statement?

sslattery: Yes this would also be my only comment here.

junghans: Fixed.

PR #143, opened 2019-05-29
added simple example of unmanaged view 

### LINKED ISSUE #137: Unmanaged AoSoA Tutorial
At a minimum we need a C-struct based unmanaged AoSoA tutorial to demonstrate using the capability merged in #114. This should likely be a separate tutorial and include calling algorithms and/or `parallel_for`.

rfbird:
Adds a simple file that:

1) Makes an [unmanaged] AoSoA from a new'd array
2) Checks the values are correct
3) Sorts the AoSoA to demonstrate it works in our standard algorithms

Closes #137

### DIFF
--- a/core/example/tutorial/14_unmanaged_view/CMakeLists.txt
+++ b/core/example/tutorial/14_unmanaged_view/CMakeLists.txt
@@ -1,0 +1,3 @@
+  add_executable(UnmanagedParallelFor parallel_for_example.cpp)
+  target_link_libraries(UnmanagedParallelFor cabanacore)
+  add_test(NAME Core_tutorial_14 COMMAND UnmanagedParallelFor)
--- a/core/example/tutorial/14_unmanaged_view/CMakeLists.txt
+++ b/core/example/tutorial/14_unmanaged_view/CMakeLists.txt
@@ -1,3 +1,3 @@
-  add_executable(UnmanagedParallelFor parallel_for_example.cpp)
-  target_link_libraries(UnmanagedParallelFor cabanacore)
-  add_test(NAME Core_tutorial_14 COMMAND UnmanagedParallelFor)
+add_executable(UnmanagedParallelFor parallel_for_example.cpp)
+target_link_libraries(UnmanagedParallelFor cabanacore)
+add_test(NAME Core_tutorial_14 COMMAND UnmanagedParallelFor)
--- a/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
+++ b/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
@@ -59,6 +59,9 @@ void parallelForExample()
     const int num_tuple = 128;
     const int num_soa = 128/8; // Be careful with the int division..
 
+    // We use `new` here to represent an arbitrary pointer, in a typical
+    // example this will be passed in from a users existing code/memory
+    // allocation (possibly from Fortran)
     Data* local_data = new Data[num_tuple];
 
     // This is equivalent to a Cabana AoSoA of:
--- a/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
+++ b/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
@@ -109,6 +109,9 @@ void parallelForExample()
             // Unexpected value
             printf("%d: Unexpected %d vs %d \n", __LINE__, slice_b(i-1) , slice_b(i) );
         }
+
+        // We can also set values
+        slice_a(i) = i / 2;
     }
 
     // Clean up local data
--- a/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
+++ b/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
@@ -27,15 +27,16 @@ void parallelForExample()
      * Declare Underlying data type
     */
 
+    const int INNER_SIZE = 2;
     // Regular class form
     class Data {
         public:
-            double a[VectorLength];
+            double a[INNER_SIZE][VectorLength];
             int    b[VectorLength];
     };
 
     // Cabana MemberTypes form:
-    using DataTypes = Cabana::MemberTypes<double, int>;
+    using DataTypes = Cabana::MemberTypes<double[INNER_SIZE], int>;
 
     /*
      * Create the AoSoA.
@@ -57,7 +58,7 @@ void parallelForExample()
     {
         for (int j = 0; j < VectorLength; j++)
         {
-            local_data[i].a[j] = i;
+            local_data[i].a[1][j] = i;
             local_data[i].b[j] = j;
         }
     }
@@ -111,7 +112,7 @@ void parallelForExample()
         }
 
         // We can also set values
-        slice_a(i) = i / 2;
+        slice_a(i, 0) = i / 2;
     }
 
     // Clean up local data
--- a/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
+++ b/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
@@ -12,7 +12,7 @@
 #include <Cabana_Core.hpp>
 #include <Cabana_Sort.hpp>
 
-#include <iostream>
+#include <cstdio>
 
 //---------------------------------------------------------------------------//
 // parallel for using an unmanaged view.
--- a/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
+++ b/core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
@@ -48,7 +48,7 @@ void parallelForExample()
 
     Data* local_data = new Data[num_tuple];
 
-    // This is equivalent to a Cabana AoAoS, which looks like:
+    // This is equivalent to a Cabana AoSoA, which looks like:
     //    Cabana::AoSoA<DataTypes,DeviceType,VectorLength> aosoa( "my_aosoa", num_tuple );
 
     /*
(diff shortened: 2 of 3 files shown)


dalg24:
Add to whitelist

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,137 @@
...
+    using DataTypes = Cabana::MemberTypes<double,double>;
+
+
+    /*
+     * Create the AoSoA.
+    */
+    const int num_tuple = 128;
+    const int num_soa = 128/8; // Be careful with the int division..
+
+    Data* local_data = new Data[num_tuple];
```
dalg24: Do you mean to allocate on the free store?  Since the size is known at compile time, why not having `local_data` on the stack?  This way, we don't encourage our users to do `new` and `delete` :)

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,137 @@
...
+    auto slice_a = aosoa.slice<0>(); // a = 0
+    auto slice_b = aosoa.slice<1>(); // b = 1
+
+    // Look at the data in the AosoA
+    for (int i = 0; i < num_tuple; i++)
+    {
+        if ( slice_a(i) != i / VectorLength )
+        {
+            // Unexpected Value
+            printf("%d: Unexpected %e != %e \n", __LINE__, slice_a(i), i / VectorLength);
```
dalg24: You want to go ahead and assert it?  (Since you included the header `<cassert>` and don't actually use it)

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
sslattery: In general a user that wants to wrap an unmanaged view around their data likely does not know the sizes at compile time. Therefore `new` or `malloc` is the correct choice for this example in terms of memory allocation.

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    /*
+     * Create the AoSoA.
+    */
+    const int num_tuple = 128;
+    const int num_soa = 128/8; // Be careful with the int division..
+
+    Data* local_data = new Data[num_tuple];
+
+    // This is equivalent to a Cabana AoAoS, which looks like:
```
sslattery: `AoAoS` -> `AoSoA`

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,137 @@
...
+    >;
+
+    // Do convert
+    auto ptr = reinterpret_cast<typename AoSoA_t::soa_type*>(local_data);
+    AoSoA_t aosoa( ptr, num_soa, num_tuple );
+
+    /*
+    * Inspect Data In the Unmanaged AoSoA
+    */
+    auto slice_a = aosoa.slice<0>(); // a = 0
```
sslattery: use `Cabana::slice<0>(aosoa)`

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
sslattery: Yeah I would `assert` or `throw` here

sslattery: Well maybe not actually - this seems to be more like a check in a unit test - would we want to halt the program here?

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    /*
+     * Declare Underlying data type
+    */
+
+    // Regular class form
+    class Data {
+        public:
+            double a[VectorLength];
+            double b[VectorLength];
```
sslattery: Maybe make on of these multidimensional to make clear the ordering of the data.

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
rfbird: Yea thats exactly the cycle I went through. I wrote it as asserts, but then ran it in Release mode and it didn't tell me when I screwed up, so decided that a manual `if` was more robust

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
rfbird: Agreed. I can (optionally) pull it from `argv` if we really don't want it to be known at compile time, but I think dynamic is the more useful example

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
rfbird: Is there anything more complicated we want to do? That could always come in a second example?

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
rfbird: fixed

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
rfbird: fixed

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
sslattery: yeah for now lets keep this basic

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
dalg24: I don't have a strong opinion.  We have debug builds in the CI so we would catch bugs.  If you decide not to have assertions then remove `#include <cassert>`.

rfbird: Let's stick with `printf`. I cut them down a bit, now they're a bit less offensive too

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -12,7 +12,7 @@
 #include <Cabana_Core.hpp>
 #include <Cabana_Sort.hpp>
 
-#include <cassert>
+#include <iostream>
```
dalg24: You want `<cstdio>`

rfbird: thanks!

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include <Cabana_Sort.hpp>
+
+#include <cstdio>
+
+//---------------------------------------------------------------------------//
+// parallel for using an unmanaged view.
+//---------------------------------------------------------------------------//
+void parallelForExample()
+{
+    /* Declare general run parameters */
```
streeve: Can you add a general description comment here at the beginning?

rfbird: I thought about this... but I really didn't know what to say. @sslattery do you have a good feel of what to say about this ?

sslattery: I would describe the premise of an unmanaged `AoSoA` and examples of why one would want to use it. The main point is someone has allocated data elsewhere that is directly mappable to a `Cabana::AoSoA` and that person would like to use Cabana algorithms in their codes. I would also comment on how you composed the user C struct and why the data layout maps to an AoSoA. Adding the extra dimension to `a` was a good idea here because it exposes the user to how the inner array layouts must be setup - with the coalesced index defined last.

rfbird: Added an attempt to address this

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+//---------------------------------------------------------------------------//
+// Example of a parallel-for using an unmanaged view.
+//---------------------------------------------------------------------------//
+void parallelForExample()
+{
+    /*
+     * Summary: An unmanaged AoSoA allows the user to wrap manually allocated
+     * data (if compatible) inside of Cabana structures, without paying the
+     * cost of duplicating, or reallocating, the underlying data structure.
+     * Once wrapper, the resulting AoSoAs can be treated like regular Cabana
```
sslattery: `wrapper` -> `wrapped`

core/example/tutorial/14_unmanaged_view/parallel_for_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+     * containers, including be invoked in standard algorithms (such as sort).
```
sslattery: `be` -> `being`

core/example/tutorial/14_unmanaged_view/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+  add_executable(UnmanagedParallelFor parallel_for_example.cpp)
+  target_link_libraries(UnmanagedParallelFor cabanacore)
+  add_test(NAME Core_tutorial_14 COMMAND UnmanagedParallelFor)
```
junghans: Indentation?

rfbird: nice catch. fixed.

sslattery:
@junghans do you approve?

guangyechen:
Does it work with `using MemorySpace = Kokkos::Cuda;`?

rfbird:
I just realized this doesn't actually call parallel for, I should probably update that..

PR #145, opened 2019-06-04
.travis.yml: bump CI to gcc-6

junghans:


junghans:
@sslattery can you look at this warning:
```
/home/travis/build/ECP-copa/Cabana/core/src/Cabana_CommunicationPlan.hpp:512:19: error: unused variable 'ec' [-Werror,-Wunused-variable]
        const int ec =
                  ^
```
I found this originally in #52, but I wanted to break it out into a separate PR.

sslattery:
This line uses an assert:
https://github.com/ECP-copa/Cabana/blob/7b856aaed8bf863af2ec11c6c153aac030a94ec6/core/src/Cabana_CommunicationPlan.hpp#L514
This always gets checked in debug so gives a warning on some release builds I guess. We should replace this with an exception anyway because we always want to check these.

junghans:
@dalg24 please review

junghans:
@dalg24 you are right, I could have done that to, but I found it more readable to add it directly.

PR #146, opened 2019-06-06
Specify jsrun as the MPIEXEC_EXECUTABLE on ECP CI builds

dalg24:


dalg24:
@junghans A cheap

> `MPIEXEC_EXECUTABLE` got only add in cmake-3.10, so we need to bump the minimal cmake version then.

Yep I figured that out in the meantime...  Fixed :)

dalg24:
@junghans my last push did not seem to trigger a pipeline on ECP CI :/

junghans:
It pulls every 5mins only.

dalg24:
It looks lie the CUDA builds do not have `nvcc_wrapper` selected as the compiler.

dalg24:
Of course... Reverting 32fafb404bc81cc5e80b0f7ec881743233417a92

core/unit_test/CMakeLists.txt
```diff
@@ -65,10 +66,10 @@ macro(Cabana_add_tests)
           if(_device STREQUAL Pthread OR _device STREQUAL OpenMP)
             foreach(_thread ${CABANA_UNIT_TEST_NUMTHREADS})
               add_test(NAME ${_target}_${_thread} COMMAND
-                ${_target} ${gtest_args} --kokkos-threads=${_thread})
+                ${NONMPI_PRECOMMAND} ${_target} ${gtest_args} --kokkos-threads=${_thread})
```
dalg24: @junghans admittedly it is probably not the right way to set the number of threads with `jsrun` but this will do for now

junghans: ok

PR #147, opened 2019-06-07
Advanced fortran examples

guangyechen:
In the advanced fortran examples:
01: dispatch parallel_for on a fortran array
02-03: extracted from the XGC-Cabana code, which set up some examples of Cabana for Fortran that can be readily adopted for users wanting to build a PIC code or incorporate Cabana kernels into an existing Fortran PIC code. 
04: dispatch simd_parallel_for and kokoos::parallel_for on a fortran array (which is sent to C++ for creating an unmanaged AoSoA).
The common feature here is that Fortran main programs (compiled with mpi) call the Cabana library, which dispatches Fortran/C++ kernels.


guangyechen:
I ran into some problems with the previous branch. This is a new branch which is basically the same as the previous one but with an additional example on unmanaged aosoa.

junghans:
> To up the coverage, can you that the examples as test (see https://github.com/ECP-copa/Cabana/blob/master/core/fortran-example/tutorial/01_hello_world/CMakeLists.txt#L3)

I saw you already did this, however none of the code gets build by the CI as all of the new examples require PGI, so that explains by the coverage goes down bu 2%.

sslattery:
Lets try to get this working on the Travis CI so we have test coverage for this. This won't get us GPU coverage due to the PGI requirement but we will at least have some testing.

guangyechen:
@junghans Could you take a look at why CI failed, and suggest a way to fix it? Thanks.

rfbird:
> @junghans Could you take a look at why CI failed, and suggest a way to fix it? Thanks.

It looks like it's failing because you are failing on the clang-format section. The error log (eg https://travis-ci.org/ECP-copa/Cabana/jobs/572451734#L1400) tells you how it wants it's fixed

I'm not sure if we have a way to say "Accept the automated fixes" or not

guangyechen:
Close it per our last discussion, and put the content to a new repo https://github.com/ECP-copa/Advanced-fortran-examples .

PR #150, opened 2019-06-14
cmake: fix mixing of compilers

### LINKED ISSUE #140: Mixing different compilers
In #111, when mixing gfortran with clang, clang flags get inherited, which gfortran won't understand.

junghans:
Fix #140 (hopefully)


cmake/FindKOKKOS.cmake
```diff
@@ -38,6 +38,7 @@ if(KOKKOS_SETTINGS_DIR AND KOKKOS_INCLUDE_DIR AND KOKKOS_LIBRARY)
     IMPORTED_LOCATION ${KOKKOS_LIBRARY}
     INTERFACE_INCLUDE_DIRECTORIES ${KOKKOS_INCLUDE_DIR}
     INTERFACE_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CXX>:${KOKKOS_CXX_FLAGS_WITHOUT_INCLUDES}>"
+    IMPORTED_LINK_INTERFACE_LANGUAGES "CXX"
```
dalg24: Did you mean to quote `CXX`?

junghans: I think it doesn't matter.

junghans:
> Does it fix the issue or not?

I tested `clang++` together with `gfortran` and it worked.

junghans:
@guangyechen can you try this with PGI?

guangyechen:
It does not seem to work with clang+pgfortran.
[error_clang_pgfortran.txt](https://github.com/ECP-copa/Cabana/files/3307939/error_clang_pgfortran.txt)

junghans:
Did you compile kokkos with clang or pgi?

guangyechen:
Clang.

junghans:
@guangyechen we could just put exclude mixing PGI with anything ;-)

dalg24:
I suspect it is mixing libstdc++ that PGI is taking from the GCC 4.8.5 on the system

PR #152, opened 2019-07-11
Recovered peak flops/cycle using gcc

guangyechen:
Recovered the peak flops/cycle of performance tests using the gcc compiler.
Note:
1. gcc 6.4.0 tested, compiled with latest kokkos develop branch.
2. broadwell (avx-256), skylake (avx-512) tested.
3. gcc seems to require the kernel function to be "noinline" to reach the peak. Not sure why. Maybe a compiler bug. 
4. flops/cycle seems to be badly measured by rdtsp() on the newer cpus. May want to change to flops/sec. 
5. clang and intel tested, but they are not able to reach the peak flops yet.


junghans:
I think the test just runs longer than 10min now, which could be related to increasing `n` by a factor 100.

sslattery:
of course.... @guangyechen can you knock off a factor of 10? I was wondering how the runtimes increased but forgot I saw the problem size increase when I reviewed

guangyechen:
Of course.

core/performance_test/peakflops/01_cpp_simple.cpp
```diff
@@ -118,7 +118,7 @@ struct data * axpy_10(struct data *__restrict__ a, struct data *__restrict__ x0,
 
 
 TEST(cpp, simple) {
-  long n = static_cast<long>(2e6);
+  long n = static_cast<long>(2e6); //use a higher number e.g. 2e8 for skylake
```
junghans: Why are we hard-coding this? Could we make this into a CMake option? Have it be `2e8` by default and use `2e6` for the CI.

guangyechen: @junghans Sounds like a good idea. I am not sure how to do that. Could you help? Thanks.

junghans: You can just copy what @rfbird did for `Cabana_PERFORMANCE_ERROR_MARGIN`:
https://github.com/ECP-copa/Cabana/blob/master/core/performance_test/peakflops/CMakeLists.txt#L17-L22

rfbird: I can do this for you Gy

rfbird: I pushed something to your branch that does it (https://github.com/guangyechen/Cabana/commit/b15345aadac0082eae3e3171abb69112b82d8cda), please take a look

I think we should also consider adding prints to feed back to the user so they can see if their attempt at changing it stuck -- what do you think? (print n and flops etc)

PR #154, opened 2019-07-11
fixing #153

### LINKED ISSUE #153: Passing Slice to Distributor causes a seg fault
The distributor accepts a `Cabana::slice` in the constructor (which is good), but it then expects the `.extents` call to work like a `Kokkos::view` and ends up creating arrays that are too small, ultimately leading to an out of bounds.

https://github.com/ECP-copa/Cabana/blob/master/core/src/Cabana_CommunicationPlan.hpp#L122

sslattery:
Adds a new unit test that recreates the bug reported in #153 - the updated code passes this test.

Fix #153

### DIFF
--- a/core/src/Cabana_DeepCopy.hpp
+++ b/core/src/Cabana_DeepCopy.hpp
@@ -241,7 +241,7 @@ inline void deep_copy(
 /*!
   \brief Fill an AoSoA with a tuple.
 
-  \param slice The AoSoA to fill.
+  \param aosoa The AoSoA to fill.
 
   \param src The tuple to assign. All AoSoA elements will be assigned this
   value.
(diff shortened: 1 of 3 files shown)


core/src/Cabana_DeepCopy.hpp
```diff
@@ -241,7 +241,7 @@ inline void deep_copy(
 /*!
   \brief Fill an AoSoA with a tuple.
 
-  \param slice The AoSoA to fill.
+  \param aosoa The AoSoA to fill.
```
sslattery: Yes I know - unrelated. But I saw it and it needed to be fixed.

rfbird: looks good

PR #156, opened 2019-07-25
Spme buildfixes

sfogerty:
The CMakeLists for longrange examples did not build SPME on CPUs, and so some simple compile errors in spme.cpp were masked. This should now build SPME on CPU and GPU.


dalg24:
ok to test

core/example/longrange/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if(Cabana_ENABLE_Cuda)
   find_package(CUDA REQUIRED)
   add_executable(SPME spme_example.cpp)
   target_link_libraries(SPME cabanacore ${CUDA_CUFFT_LIBRARIES})
-elseif()
+else()
```
dalg24: :(
Was it evaluating to false?

core/example/longrange/spme.cpp
```diff
@@ -358,15 +359,27 @@ double TPME::compute( ParticleList& particles, ParticleList& mesh, double lx, do
...
               }
               else
               {
+                  #ifdef Cabana_ENABLE_Cuda
                   BC[idx].x = 0.0;
                   BC[idx].y = 0.0;//set origin element to zero
+                  #else
+                  BC[idx][0] = 0.0;
+                  BC[idx][1] = 0.0;//set origin element to zero
+                  #endif
```
dalg24: Did you guys consider encapsulating the element access to avoid the proliferation of preprocessor directives?

dalg24:
Also how comes the tests were passing before?  Is that code not running as part of the CI??

core/example/longrange/CMakeLists.txt
sfogerty: Yes, apparently CMake treats this as `elseif(FALSE)`. This is why the build tests passed - it never built SPME.

It looks like similar issues were encountered by others, and even within CMake source itself? [https://gitlab.kitware.com/cmake/cmake/merge_requests/3515](url)

core/example/longrange/spme.cpp
sfogerty: Yes, our to-do list includes wrapping all FFT stuff to mostly remove those directives cluttering the code.

core/example/longrange/CMakeLists.txt
dalg24: Haha.  Thanks for sharing the reference!

core/example/longrange/spme.cpp
dalg24:
```c++
#ifdef Cabana_ENABLE_Cuda
inline double& real(cufftDoubleComplex& c) { return c.x; }
inline double const& real(cufftDoubleComplex const& c) { return c.x; }
inline double& imag(cufftDoubleComplex& c) { return c.y; }
inline double const& imag(cufftDoubleComplex const& c) { return c.y; }
#else
inline double& real(fftw_complex& c) { return c[0]; }
inline double const& real(fftw_complex const& c) { return c[0] }
inline double& imag(fftw_complex& c) { return c[1]; }
inline double const& imag(fftw_complex const& c) { return c[1]; }
#endif
```

dalg24: Anyway this belong to another PR.

dalg24:
@junghans On a side note I just can't make sense of the coverage reports.

PR #157, opened 2019-07-25
Fix #155

### LINKED ISSUE #155: Indefinite Blocking in Migrate
In some cases the communication plan is created incorrectly leading to an indefinite wait. This is caused when a recv anticipates a message from the incorrect MPI rank. The problem seems to lie in the indexing that specifies the number of imports from a neighbor. The line should be changed to use iterators for the index.
```_num_import[found_neighbor-_neighbors.begin()] = import_sizes[i];```
https://github.com/ECP-copa/Cabana/blob/master/core/src/Cabana_CommunicationPlan.hpp#L687

### PR DESCRIPTION
**Pull request by sslattery**
Thanks to @LucasBondDavis for finding the bug and submitting the unit test. I modified the unit test slightly to make it more general and to work on an arbitrary number of processors.

The way the test is written the bug only shows up when we use 3 or more processors. I could probably refactor such that the bug happens with 2 processors but I think this is sufficient. 

Basically the communication plan was wrong when the user didn't specify a topology *and* there were ranks that received data from ranks that they didn't send to. See the essentially one line fix in the code.

Fixed #155

### CODE REVIEW

File: core/src/Cabana_CommunicationPlan.hpp [hunk 1]
```diff
@@ -670,21 +670,26 @@ class CommunicationPlan
...
+            // to) record this.
             if ( found_neighbor == std::end(_neighbors) )
             {
                 _neighbors.push_back( source );
                 _num_import.push_back( import_sizes[i] );
                 _num_export.push_back( 0 );
             }
+
+            // Otherwise if we already sent something to this neighbor that
+            // meqns we already have a neighbor/export entry. Just assign the
```
dalg24:
```suggestion
            // means we already have a neighbor/export entry. Just assign the
```

core/unit_test/tstDistributor.hpp
```diff
@@ -625,6 +625,89 @@ void test7( const bool use_topology )
...
+
+    // 2 items to the ranks before and after it
+    int num_data = 2;
+    Kokkos::View<int*,TEST_MEMSPACE>
+        export_ranks( "export_ranks", num_data );
+    auto fill_ranks =
+        KOKKOS_LAMBDA( const int )
+        {
+            export_ranks(0) = ( my_rank == 0 ) ? 0 : my_rank - 1;
+            export_ranks(1) = ( my_rank == my_size - 1 ) ? 0 : my_rank + 1;
```
dalg24: Is that a typo?
```suggestion
            export_ranks(1) = ( my_rank == my_size - 1 ) ? my_size - 1 : my_rank + 1;
```

I have a hard time following your comment `"2 items to the ranks before and after it"`

sslattery: This zero actually has to be here to trigger the bug - it creates a condition where rank 0 receives data from a rank it does not send data to.

sslattery: I'll update the comment to make this more clear.

PR #158, opened 2019-07-26
Clang Format

sslattery:
Adds a clang format file with hooks to run this automatically.


sslattery:
@junghans Clang format file has been added to this PR

sslattery:
I have reduced the clang format file to only make a few changes to the default LLVM format

junghans:
Ok, I added a `make format` rule and added it to CI, this should fail now.

.travis.yml
```diff
@@ -97,6 +97,7 @@ script:
           ${COVERAGE:+-DCabana_ENABLE_COVERAGE_BUILD=ON -DCOMPILER_SUPPORTS_MARCH=OFF} .. &&
     make -k -j2 VERBOSE=1 &&
     make test CTEST_OUTPUT_ON_FAILURE=1 &&
+    make format && git diff --exit-code &&
```
dalg24: Nice

CMakeLists.txt
```diff
@@ -133,3 +133,11 @@ add_subdirectory(core)
 
 configure_file(cmake/Cabana.pc.cmakein Cabana.pc @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/Cabana.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
+
+find_package(CLANG_FORMAT 7.0.1)
+if(CLANG_FORMAT_FOUND)
+  file(GLOB_RECURSE FORMAT_SOURCES core/*.cpp core/*.hpp)
```
dalg24: I would have gone for something like `git ls-files | grep -E "\.hpp$|\.cpp$"` but that works.

junghans: Yeah, I just wanted to avoid the usage of more external tools like git and grep.

junghans:
> Find module for clang-format feels a bit overkill but this is good overall.

It is, but at least now we can require a minimum version if needed.

junghans:
@sslattery if the formatting is fine with you, let’s merge this.

junghans:
ping @sslattery

core/src/Cabana_Parallel.hpp
```diff
@@ -20,33 +20,27 @@
 #include <cstdlib>
 #include <type_traits>
 
-namespace Cabana
-{
+namespace Cabana {
```
sslattery: This is probably the main thing I don't like - having an open `{` or `(` not start on the next line.

.clang-format
```diff
@@ -0,0 +1,8 @@
+BasedOnStyle:  LLVM
+---
+Language:        Cpp
+AlwaysBreakTemplateDeclarations: true
+BreakBeforeBraces: Custom
```
sslattery: Would this get changed to true to allow for the break I mentioned below?

junghans: `true` don't work, but `Allman` does.

core/src/Cabana_Parallel.hpp
junghans: Fixed!

junghans:
The remaining differences seem to be different defaults in `clang-format-7` and `clang-format-8`. What style do you prefer?

sslattery:
Nice! For me I have no preference of 7 vs. 8. If the difference is not substantial I would say go with the latest (8).

PR #160, opened 2019-08-02
Remove deprecated Kokkos code

dalg24:


sslattery:
Looks like the Jenkins might have Kokkos 2.8 but why is the travis failing? I can't seem to find the error in the output log.

dalg24:
I had pushed a couple new images to Docker Hub but forgot to update the Compose file...
For the record `dalg24/cabana-base:19.08.0` is with deprecated code enabled, and it is disabled in `19.08.1`.

PR #163, opened 2019-09-05
Use Kokkos after its CMake overhaul

dalg24:
@junghans I left out Travis CI for now


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ option(CMAKE_VERBOSE_MAKEFILE "Generate verbose Makefiles" OFF)
 include(GNUInstallDirs)
 
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
-find_package(KOKKOS REQUIRED)
+find_package(Kokkos REQUIRED)
```
junghans: Is the imported target still called `Kokkos::kokkos`?

dalg24: Yessir

junghans:
Can you modify https://github.com/ECP-copa/Cabana/blob/master/.travis.yml#L39 to test travis as well?

.travis.yml
```diff
@@ -88,6 +87,7 @@ script:
   - export FFLAGS="-Wall -Wextra -pedantic ${WERROR:+-Werror}" 
   - mkdir build && pushd build &&
     cmake -DCMAKE_PREFIX_PATH="$HOME/kokkos;$HOME/scafacos"
+          -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
```
junghans: I don't think that is needed as we use symlinked compilers.

dalg24: I removed the symlink

junghans: Ok, this makes it more obvious.

sfogerty:
Let's remember to change the wiki build instructions when this goes through.

Building Kokkos' using its develop branch using CMake

junghans:
Hmm, I have no idea by AMPI segfaults with 4 tasks, @evan-charmworks any idea?

matthiasdiener:
@junghans: The segfault is actually an unhandled C++ exception in the application code. Charm++/AMPI do not use exceptions, but we have a default exception handler that aborts on all uncaught exceptions: https://github.com/UIUC-PPL/charm/commit/becc7731feef74b4001fbe31a85658d054411fb6

junghans:
@dalg24 now we are back in kokkos space for the exception.

dalg24:
@matthiasdiener Thanks for looking into it.
I installed Charm++ this morning and wasn't able to reproduce the error.  Do you remember what line was throwing?
It really puzzles me because there is no AMPI specific code on Cabana as far as I can tell and no exception is thrown with other MPI implementations.

matthiasdiener:
@dalg24: I haven't build this myself, but this is what I understood from the build logs in Travis. I don't know which of the lines in `Kokkos::SharedAllocationRecord::decrement` actually throws the exceptions (I think there are 3 lines in that function where exceptions could be thrown). Do you have a script to build Kokkos/Cabana?

sslattery:
Retest this once #171 is merged

sslattery:
If we want to depend on a Kokkos release then the merge of this PR will be blocked by the release of Kokkos 3.0

junghans:
@dalg24 are we ready to pull the trigger here?

scripts/jenkins/Dockerfile
```diff
@@ -35,21 +35,22 @@ RUN export OPENMPI_VERSION=4.0.0 && \
 
 # Install Kokkos
 ENV KOKKOS_DIR=/opt/kokkos
-RUN export KOKKOS_VERSION=2.9.00 && \
-    export KOKKOS_URL=https://github.com/kokkos/kokkos/archive/${KOKKOS_VERSION}.tar.gz && \
-    export KOKKOS_ARCHIVE=kokkos-${KOKKOS_VERSION}.tar.gz && \
-    export KOKKOS_SOURCE_DIR=kokkos-${KOKKOS_VERSION} && \
+RUN KOKKOS_VERSION=master && \
```
sslattery: should we fix to 3.0?

junghans: There is no master tag ;-)

junghans: Neither will there is a 3.0 one.

dalg24: You can use the revision umber if you want to be explicit and replace with a tag later.

PR #165, opened 2019-09-05
cmake: export cabanacore target

### LINKED ISSUE #164: CMake: export library target for Cabana{MD,PIC}

junghans:
Fix #164


core/src/CabanaConfig.cmakein
```diff
@@ -0,0 +1,6 @@
+include(CMakeFindDependencyMacro)
+find_dependency(KOKKOS REQUIRED)
```
junghans: @dalg24 the crux is here, if want to make this work before #163 is merge, we would need to install our `FindKOKKOS.cmake`, which is just ugly. So wait?

junghans:
@sslattery @dalg24 looking `CabanaMD`, I think we need to export `CabanaMD_ENABLE_Serial` etc. in the exported targets as well. Any objections?

sslattery:
Do you mean `Cabana_ENABLE_Serial`? If so I have no objections. One thing to note is that in Cajita I have moved to just automatically using the backends on which Kokkos is built rather than having the user explicitly specify. We then just propagate/use the `Kokkos_ENABLE_Serial` instead. What do you think about doing it this way instead? It seems a user would be less likely to mismatch Kokkos/Cabana backends. They likely want the ones they built Kokkos with anyway.

junghans:
> Do you mean `Cabana_ENABLE_Serial`? If so I have no objections. One thing to note is that in Cajita I have moved to just automatically using the backends on which Kokkos is built rather than having the user explicitly specify. We then just propagate/use the `Kokkos_ENABLE_Serial` instead. What do you think about doing it this way instead? It seems a user would be less likely to mismatch Kokkos/Cabana backends. They likely want the ones they built Kokkos with anyway.

Actually I like that better, we should do the same in Cabana and just go with the kokkos backends.

sslattery:
> > Do you mean `Cabana_ENABLE_Serial`? If so I have no objections. One thing to note is that in Cajita I have moved to just automatically using the backends on which Kokkos is built rather than having the user explicitly specify. We then just propagate/use the `Kokkos_ENABLE_Serial` instead. What do you think about doing it this way instead? It seems a user would be less likely to mismatch Kokkos/Cabana backends. They likely want the ones they built Kokkos with anyway.
> 
> Actually I like that better, we should do the same in Cabana and just go with the kokkos backends.

OK - should we do that in this PR then?

junghans:
> OK - should we do that in this PR then?

No, let's do a separate one, the export target in this PR just don't set the `Cabana_ENABLE_*` at all for now.

sslattery:
> No, let's do a separate one, the export target in this PR just don't set the `Cabana_ENABLE_*` at all for now.

OK - I'll make a new PR for this

junghans:
@dalg24 please review.

.travis.yml
```diff
@@ -92,7 +93,13 @@ script:
     make test CTEST_OUTPUT_ON_FAILURE=1 &&
     make format && git diff --exit-code &&
     make doxygen &&
-    make install DESTDIR=${PWD}/install && rm -rf ${PWD}/install/usr/local && rmdir ${PWD}/install/usr &&
+    make install DESTDIR=${PWD}/install && rm -rf ${PWD}/install/$HOME/cabana && rmdir ${PWD}/install/$HOME &&
```
dalg24: Why do we have this line at all?

core/example/tutorial/01_hello_world/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
-  add_executable(HelloWorld hello_world.cpp)
-  target_link_libraries(HelloWorld cabanacore)
-  add_test(NAME Core_tutorial_01 COMMAND ${NONMPI_PRECOMMAND} HelloWorld)
+# the next for lines are just, so that we can use this as
+# a standalone CMakeLists.txt for testing of the Target export
+cmake_minimum_required(VERSION 3.9)
```
dalg24: Should we be setting the project name as well?

dalg24: I was surprised we didn't get a warning from CMake when I looked at the logs on TravisCI.  I guess you can ignore then.

.travis.yml
sslattery: I guess I am confused what this line does.

junghans: This line tests if `make install` installs any files outside of `CMAKE_INSTALL_PREFIX`.

core/example/tutorial/01_hello_world/CMakeLists.txt
sslattery: Do we need to up the minimum required?

junghans: We could set the project name, but it is not necessary, `cmake_minimum_required()` is only thing needed.

.travis.yml
dalg24: paths do not look right

dalg24: I take it back no that I looked at the [doc](https://www.gnu.org/prep/standards/html_node/DESTDIR.html)

junghans: Huh? When using `DESTDIR` everything gets prefixed with it value.

core/example/tutorial/01_hello_world/CMakeLists.txt
junghans: @sslattery this is the minimum required version for the dependent package.

PR #169, opened 2019-09-19
Updating AoSoA to not always initialize data

### LINKED ISSUE #89: Support Kokkos::ViewAllocateWithoutInitializing in AoSoA
The current AoSoA semantics always zero-initialize data. We should support using `Kokkos::AllocateViewWithoutInitializing` to allow for performance improvements in many cases.

sslattery:
We have always said this was the expected behavior but upon further inspection this was not what was actually happening and hence causing a performance hit. Thanks to @weinbe2 for measuring the performance issues

Closes #89

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -190,8 +190,11 @@ class AoSoA
         : _size( 0 )
         , _capacity( 0 )
         , _num_soa( 0 )
-        , _data( label, 0 )
+        , _data( Kokkos::ViewAllocateWithoutInitializing( label ), 0 )
     {
+        static_assert(
+            !memory_traits::is_unmanaged,
+            "Construction by allocation cannot use unmanaged memory" );
     }
 
     /*!
@@ -226,8 +229,11 @@ class AoSoA
         : _size( n )
         , _capacity( 0 )
         , _num_soa( 0 )
-        , _data( label, 0 )
+        , _data( Kokkos::ViewAllocateWithoutInitializing( label ), 0 )
     {
+        static_assert(
+            !memory_traits::is_unmanaged,
+            "Construction by allocation cannot use unmanaged memory" );
         resize( _size );
     }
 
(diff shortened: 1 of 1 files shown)


sslattery:
@weinbe2 can you run this through the profiler and see if this improves things? If it does can you put some before and after results in the PR?

core/src/Cabana_AoSoA.hpp
```diff
@@ -358,8 +365,21 @@ class AoSoA
...
+            num_soa_alloc );
+        if ( _num_soa > 0 )
+        {
+            // Note that we make an unmanaged view here. The subview approach
+            // was creating a view rank mismatch. This allows us to have two
+            // views of the same size for the call to deep_copy.
+            Kokkos::View<soa_type *, device_type, Kokkos::MemoryUnmanaged>
+                resized_data_um( resized_data.data(), _num_soa );
+            Kokkos::deep_copy( resized_data_um, _data );
+        }
```
dalg24: FYI https://github.com/kokkos/kokkos/pull/2285

Also
```C++
Kokkos::deep_copy(Kokkos::subview(resized_data, std::make_pair(0, _num_soa)), _data);
```

sslattery: that would have been nice - and yeah good call on the syntax

sslattery:
From @weinbe2

weinbe2:
Further expanding on @sslattery 's data, this was on a Quadro GV100, timings within `Cabana_CommPerformance.cpp`, running using the Cuda memory space for both the particles and communications buffers. Problem size 10240000 particles.

First line: `aosoa_type src_particles( num_particle );` --- now doesn't initialize data, 12 us is just overheads
Second line: `dst_particles.resize( distributor_fast.totalNumImport() );` --- again, now doesn't initialize data but does have to run a `cudaMalloc`, ~2.2 ms reasonable.

sslattery:
Yes - the established/documented behavior had been that it will not initialize. However, the implementation didn't necessarily do that.

PR #173, opened 2019-10-09
Adding empty() and shrinkToFit() to AoSoA

sslattery:
These two additional functions further align `Cabana::AoSoA` with a more `std::vector`-like memory management interface. The `empty()` function is convenience for size checks and `shrinkToFit()` allows users to clean up excess capacity in problems with dynamic particle populations. The latter will be particularly important in a variety of ongoing ECP use cases.

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -278,6 +278,17 @@ class AoSoA
     size_type size() const { return _size; }
 
     /*!
+      \brief Returns if the container is empty or not.
+
+      \return True if the number of tuples in the container is zero.
+
+      This is the number of actual objects held in the container, which is not
+      necessarily equal to its storage capacity.
+    */
+    KOKKOS_FUNCTION
+    bool empty() const { return ( _size == 0 ); }
+
+    /*!
       \brief Returns the size of the storage space currently allocated for the
       container, expressed in terms of tuples.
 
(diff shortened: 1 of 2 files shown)


core/src/Cabana_AoSoA.hpp
```diff
@@ -277,6 +277,17 @@ class AoSoA
...
+    /*!
+      \brief Returns if the container is empty or not.
+
+      \return True if the number of tuples in the container is zero.
+
+      This is the number of actual objects held in the container, which is not
+      necessarily equal to its storage capacity.
+    */
+    KOKKOS_FUNCTION
+    bool empty() const { return ( _size == 0 ); }
```
dalg24:
```suggestion
    bool empty() const { return ( size() == 0 ); }
```

core/src/Cabana_AoSoA.hpp
```diff
@@ -379,6 +396,42 @@ class AoSoA
...
+        soa_view resized_data(
+            Kokkos::ViewAllocateWithoutInitializing( _data.label() ),
+            _num_soa );
+        if ( _num_soa > 0 )
+            Kokkos::deep_copy(
+                resized_data,
+                Kokkos::subview( _data, Kokkos::pair<size_type, size_type>(
+                                            0, _num_soa ) ) );
+        _data = resized_data;
+    }
```
rfbird: Per an offline discussion with Stuart, this will obviously cost ~2x in memory (possibly blowing out and causing an OOM), as well as have some performance penalty to copying the data over. Users may expect something like a shrink to be cheap and get shot in the foot. As things stands (wrt the kokkos resize initialization stuff), there is no good way to avoid this, but it might be nice to somehow communicate it to potential users?

Eventually it might be nice to rely on `Kokkos::resize` of our underlying views, but this is good enough for now

PR #175, opened 2019-10-10
Halo Owner Ranks

### LINKED ISSUE #77: Add MPI support for Periodic Boundaries and Cartesian Decompositions
Many applications in PIC and MD operate on a Cartesian Domain decomposition and often employ periodic boundary conditions. Our communication structures currently do not consider a Cartesian case and are instead more general. Because of this, we may miss opportunities for optimization. In addition, it makes support for periodic boundaries difficult.

sslattery:
After working with @streeve on the issue of handling periodic boundaries with the halo it became clear that we could resolve this issue by providing a user with the rank that owns each ghost value in a halo. Once the owner is known, a user can then perform any operation needed on the ghost data after gathering or before scattering based on the rank the data came from (e.g. performing a periodic shift of particle coordinates for particles that were communicated across a periodic boundary).

This information was already available in the halo and could have ostensibly been generated by the user, but the new function `Cabana::Halo::ghostOwningRanks()` provides a simple output with this information for the user.

Fixes #77

### DIFF
--- a/core/src/Cabana_Halo.hpp
+++ b/core/src/Cabana_Halo.hpp
@@ -115,6 +115,7 @@ class Halo : public CommunicationPlan<DeviceType>
             element_export_ranks, neighbor_ranks, mpi_tag );
         this->createExportSteering( neighbor_ids, element_export_ranks,
                                     element_export_ids );
+        createGhostOwningRanks();
     }
 
     /*!
@@ -167,6 +168,7 @@ class Halo : public CommunicationPlan<DeviceType>
             this->createFromExportsOnly( element_export_ranks, mpi_tag );
         this->createExportSteering( neighbor_ids, element_export_ranks,
                                     element_export_ids );
+        createGhostOwningRanks();
     }
 
     /*!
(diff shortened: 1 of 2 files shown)


core/unit_test/tstHalo.hpp
```diff
@@ -73,6 +73,27 @@ void test1( const bool use_topology )
     EXPECT_EQ( halo->numLocal(), num_local );
     EXPECT_EQ( halo->numGhost(), my_size );
 
+    // Check the owning ranks of the ghost data.
+    auto ghost_own_ranks = halo->ghostOwningRanks();
+    EXPECT_EQ( halo->numGhost(), ghost_own_ranks.size() );
+    auto own_ranks_host = Kokkos::create_mirror_view_and_copy(
+        Kokkos::HostSpace(), ghost_own_ranks );
+    for ( int i = 0; i < my_size; ++i )
```
dalg24: `my_size == halo->numGhost()`?

core/src/Cabana_Halo.hpp
```diff
@@ -185,8 +187,45 @@ class Halo : public CommunicationPlan<DeviceType>
...
+            Kokkos::HostSpace(), _ghost_owning_ranks );
+
+        std::size_t ghost_id = 0;
+        for ( int n = 0; n < this->numNeighbor(); ++n )
+        {
+            auto nrank = this->neighborRank( n );
+            for ( std::size_t i = 0; i < this->numImport( n ); ++i, ++ghost_id )
+            {
+                owning_ranks_mirror( ghost_id ) = nrank;
+            }
```
dalg24: Can't remember the implementation of `Halo` but it feels like this assumes the layout of the ghosts.

core/unit_test/tstHalo.hpp
sslattery: Yes - will update the test to use this value.

core/src/Cabana_Halo.hpp
sslattery: Yes - the halo has a layout specified by the incoming data and is chunked by neighbor rank in the order specified by the underlying data in `Cajita::CommunicationPlan`. I will make a comment in the test to clarify.

sslattery: Actually I will comment in the implementation.

sslattery:
Turns out this doesn't give quite enough information to fix the periodicity problem so we won't merge this one and go for another solution instead.

PR #178, opened 2019-10-17
Fixing missing header and deprecated constructors in benchmarks

sslattery:
This PR fixes a missing header and use of deprecated constructors in one of the benchmarks. This indicates that in either Travis or Jenkins these tests weren't getting built or run so we need to update the CI settings to ensure this.


sslattery:
@junghans this one is clean now - I had to revert a previous accidental merge so there is a new PR for this one.

Issue #172, opened 2019-10-03
VerletList optimization 


streeve:
The VerletList could be improved with an initial allocation guess of neighbors per particle, count during the neighbor fill, and a second fill only if reallocation is necessary.

Currently, both implementations count and fill every time.

PR #181, opened 2019-10-24
Verlet AllocateWithoutInit

streeve:
In looking into #172, the most significant speedup came from allocating the neighbor lists without initialization, not from removing the count.

The unit test was modified to properly test #133

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -374,7 +374,8 @@ struct VerletListBuilder
     {
         // Allocate offsets.
         _data.offsets = Kokkos::View<int *, memory_space>(
-            "neighbor_offsets", _data.counts.size() );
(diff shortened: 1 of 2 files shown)


streeve:
Here are a few timings on 1 V100. I included the total runtime and neighbor construction time for the default test in CabanaMD (where total shows how much slower accessing the CSR list is, while construction is slightly faster)

```
Old
Atoms 2D CSR
256000   0.642932(0.118848)  0.814226(0.097267) Total(Neighbor)
2048000  2.373256(0.795977)  3.910342(0.626304)
16384000 17.434220(6.281761) 30.798254(5.009196)

New
Atoms 2D CSR
256000   0.611078(0.084963)  0.772870(0.064450)
2048000  2.055771(0.475183)  3.641815(0.402845)
16384000 14.709009(3.543217) 28.804732(2.982732)
```

sslattery:
Nice! So is that translating to a 2x speedup in the neighbor list construction for the larger problem?

PR #182, opened 2019-10-28
VerletList 2D fill/refill

streeve:
Optimization for 2D neighbor layout where count and fill is replaced with fill (and refill only if the preallocated array is exceeded). The CSR layout count and fill is unchanged.

A `Cabana::NeighborList::maxNeighbor` function is added to easily keep the array size for the next neighbor list build.

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -392,11 +392,14 @@ struct VerletListBuilder
 
     void initCounts( VerletLayout2D )
     {
-        count = false;
+        if ( max_n > 0 )
+        {
+            count = false;
 
-        _data.neighbors = Kokkos::View<int **, memory_space>(
-            Kokkos::ViewAllocateWithoutInitializing( "neighbors" ),
-            _data.counts.size(), max_n );
+            _data.neighbors = Kokkos::View<int **, memory_space>(
+                Kokkos::ViewAllocateWithoutInitializing( "neighbors" ),
+                _data.counts.size(), max_n );
+        }
     }
 
     void processCounts( VerletLayoutCSR )
@@ -445,7 +448,8 @@ struct VerletListBuilder
         Kokkos::fence();
 
         // Reallocate the neighbor list if previous size is exceeded.
-        if ( (std::size_t)max_num_neighbor > _data.neighbors.extent( 1 ) )
+        if ( count or (std::size_t)max_num_neighbor >
+             _data.neighbors.extent( 1 ) )
         {
             refill = true;
             Kokkos::deep_copy( _data.counts, 0 );
@@ -639,7 +643,7 @@ class VerletList
                 const typename PositionSlice::value_type cell_size_ratio,
                 const typename PositionSlice::value_type grid_min[3],
                 const typename PositionSlice::value_type grid_max[3],
-                const std::size_t max_neigh = 32,
+                const std::size_t max_neigh = 0,
                 typename std::enable_if<( is_slice<PositionSlice>::value ),
                                         int>::type * = 0 )
     {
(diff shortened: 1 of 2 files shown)


streeve:
Timings on 1 V100 below. Only about 3% overall speedup for the larger systems. And as you would expect, not passing a large enough max_neighbors array size ends up being twice as slow (bottom), where a reallocation and refill is required every time.

```
Old
Atoms 2D CSR
256000   0.625873(0.084249)  0.870528(0.065755) Total(Neighbor)
2048000  2.391449(0.461928)  4.514206(0.397901)
16384000 20.196130(3.587699) 37.433971(2.998192)

New
Atoms 2D CSR
256000   0.616126(0.076382)  0.870992(0.065048)
2048000  2.323710(0.395443)  4.565645(0.406780)
16384000 19.550911(3.013080) 37.501739(3.008820)

New, refill every time
Atoms 2D
256000  0.676724(0.143572)
2048000 2.899219(0.956383)
```

sslattery:
I guess I am confused with the timing results - how does the old code with 2D get the same performance as the new code with a good max neighbors guess while refilling everything is twice as slow. Isn't refilling every time equivalent to the old code?

streeve:
With a bad guess we're replacing count+fill with fill+fill, or with a good guess replacing count+fill with fill. So the count is just so much less expensive we don't save much time, or double it if we're not careful

sslattery:
OK I have been through the new logic in the code a few times and I think it seems correct in terms of the behavior we want. It's interesting that `fill` is so much more expensive that `count` but now that I look at it there could be a significant reduction in atomic operations in `count` because it is formulated as a reduction rather than a parallel for.

core/src/Cabana_VerletList.hpp
```diff
@@ -213,22 +215,37 @@ struct VerletListBuilder
...
+
     // Constructor.
     VerletListBuilder( PositionSlice slice, const std::size_t begin,
                        const std::size_t end,
                        const PositionValueType neighborhood_radius,
                        const PositionValueType cell_size_ratio,
                        const PositionValueType grid_min[3],
-                       const PositionValueType grid_max[3] )
+                       const PositionValueType grid_max[3],
+                       const std::size_t max_neigh )
```
sslattery: Let's have a sensible default for this. Better yet, lets default to zero and if we get zero, fall back on the old implementation. That way if the user doesn't provide a value they get decent performance and if they do, they will know it needs to be big enough to get the performance improvement.

sslattery:
Also a side note: in your results the CSR neighbor list construction is faster but it does indeed seem like we are getting a performance improvement still from 2D. Can we still say this improvement comes from data access patterns then?

core/src/Cabana_VerletList.hpp
streeve: That's a good point. I switched the default (in the VerletList constructor) to zero

streeve:
> Can we still say this improvement comes from data access patterns then?
Yes, I think that is the case

streeve:
Here are the updated timings. I fixed one mistake (missing another AllocateWIthoutInit) which only affected cases which refill, then changed the behavior when not passing the max allocation size to count+fill. 

```
Old (f320f26e44d54e794cdb45d752456a647e23bb52)
Atoms 2D CSR
256000   0.625873(0.084249)  0.870528(0.065755) Total(Neighbor)
2048000  2.391449(0.461928)  4.514206(0.397901)
16384000 20.196130(3.587699) 37.433971(2.998192)

New (431a8ab3d8cb0b09d7384916d332d89fcae62168)
Atoms 2D CSR
256000   0.572058(0.072108)  0.870992(0.065048)
2048000  2.131263(0.394467)  4.565645(0.406780)
16384000 18.943566(2.932039) 37.501739(3.008820)

New, refill every time, added missing ViewWithoutInit (e85a5444c3b25af7f971d82a3d21f59a4\
f139402)
Atoms 2D
256000   0.591669(0.100976)
2048000  2.290826(0.561492)
16384000 20.410003(4.362833)

New, refill every time, revert back to count (19c906b1a64200b09b1e9dce1727e152343d3c96)
Atoms 2D
256000   0.567128(0.078500)
2048000  2.204067(0.464687)
16384000 19.577390(3.524220)
```

streeve:
Forgot to close #172 and link to the previous #75, a related optimization for reusing linked cell lists within neighbor lists that we have decided not to pursue for now

PR #186, opened 2019-12-02
Remove MPI Comm Tags

### LINKED ISSUE #179: Use MPI_Comm_Dup in favor of exposing tags
The previous strategy for parallel communication in the library was to use an MPI communicator provided by the user and then expose a tag in communication calls to ensure that we wouldn't interfere with other communication happening on the given communicator. By using `MPI_Comm_Dup` instead to create a new duplicate communicator for MPI-based classes, we do not need to expose this tag as we create a new, separate communication space. This also makes asynchronous operations much easier. `MPI_Comm_Dup` should be relatively cheap and require no communication.

sslattery:
Use `MPI_Comm_dup` to create a new communication space for all library objects which use a communicator. Duplication creates a new communication space so MPI tags for messages are no longer needed to be provided by the user.

Fixes #179 

Note that this change breaks backwards compatibility. As the previous interface exposed tags as an optional argument you won't need to change anything unless you explicitly set a tag.

### DIFF
--- a/core/src/Cabana_CommunicationPlan.hpp
+++ b/core/src/Cabana_CommunicationPlan.hpp
@@ -287,8 +287,10 @@ class CommunicationPlan
       \return The MPI communicator for this plan.
     */
     CommunicationPlan( MPI_Comm comm )
-        : _comm( comm )
     {
+        // Duplicate the communicator so this object has it's own
+        // communication space.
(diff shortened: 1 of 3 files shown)


sslattery:
Good call - added a destructor

sslattery:
retest this please

PR #189, opened 2020-01-03
AoSoA resize without deprecated code

streeve:
When resizing an AoSoA, the underlying deep_copy can fail without deprecated code because of mismatched sizes. This only comes up with multiple subsequent resize operations with differing numbers of SoA.

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -371,7 +371,8 @@ class AoSoA
                 Kokkos::subview(
                     resized_data,
                     Kokkos::pair<size_type, size_type>( 0, _num_soa ) ),
-                _data );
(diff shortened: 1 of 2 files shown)


core/src/Cabana_AoSoA.hpp
```diff
@@ -405,7 +406,9 @@ class AoSoA
             _num_soa );
         if ( _num_soa > 0 )
             Kokkos::deep_copy(
-                resized_data,
+                Kokkos::subview(
+                    resized_data,
+                    Kokkos::pair<size_type, size_type>( 0, _num_soa ) ),
```
dalg24: This is unnecessary. `resized_data` already has the right size.
https://github.com/ECP-copa/Cabana/blob/08ae3750bf9743511600c4ceef8e3cd67e2b252f/core/src/Cabana_AoSoA.hpp#L404-L406

PR #190, opened 2020-01-08
Add host_mirror_type for explicit use with host data

rfbird:
This adds a comparable type helper to `Kokkos::HostMirror` is useful if you want to store a class level pointer to the data

- Also removes misleading comment that says deep copy occurs to the host, when it really occurs to the specified (memory) space

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -113,12 +113,8 @@ class AoSoA
     using aosoa_type = AoSoA<DataTypes, DeviceType, VectorLength, MemoryTraits>;
 
     // Host mirror type.
-    using host_mirror_type = AoSoA<
-        DataTypes,
-        Kokkos::HostSpace, // mirrors return AoSoA build with spaces not device
-        VectorLength,
-        MemoryTraits
-    >;
+    using host_mirror_type =
+        AoSoA<DataTypes, Kokkos::HostSpace, VectorLength, MemoryTraits>;
 
     // Member data types.
     static_assert( is_member_types<DataTypes>::value,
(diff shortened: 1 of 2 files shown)


core/src/Cabana_AoSoA.hpp
```diff
@@ -112,6 +112,10 @@ class AoSoA
     // AoSoA type.
     using aosoa_type = AoSoA<DataTypes, DeviceType, VectorLength, MemoryTraits>;
 
+    // Host mirror type.
+    using host_mirror_type =
+        AoSoA<DataTypes, Kokkos::HostSpace, VectorLength, MemoryTraits>;
```
dalg24: Where did you mean to use that public member type?

rfbird: I was going to add it in for use with the mirror return types but it turns out we don't need to use it (unless we want to add a host specific specialization).

Where I actually think it's useful is so I can store a class pointer to a mirror, without manually constructing the type (in user code).

In VPIC I now do:

```
class foo {
  // set this later using Cabana::create_mirror_view
  particle_t::host_mirror_type particles_host;
};
```

dalg24: Not sure it does what you want for unmanaged memory though.

rfbird: Hmm. I guess I'm not sure either..

rfbird: I agree with it not working well under unmanaged memory -- good catch, thanks. 

I had a chat with Stuart and we decided it might be better to avoid inheriting any traits at all (which I just pushed) -- what are your thoughts on how this should look?

We can also just not merge this, and I can do some declytyping / hard coding in user space

sslattery: For reference, not inheriting traits would then provide the exact same behavior as `Kokkos::View<>::HostMirror` hence why we chose to remove them.

PR #198, opened 2020-02-06
cmake: fix non-mpi build

### LINKED ISSUE #193: Long Range Example Requires MPI even when MPI is disabled
The spme example depends on particles.h, which includes mpi. It's not conditional on MPI like the `directsum_example`

https://github.com/ECP-copa/Cabana/blob/500326520268bcaa05e77ca8c9a6a7cb7320230b/core/example/longrange/CMakeLists.txt#L10

junghans:
Fix #193

### DIFF
--- a/core/example/longrange/CMakeLists.txt
+++ b/core/example/longrange/CMakeLists.txt
@@ -1,9 +1,11 @@
-if(Cabana_ENABLE_MPI)
-  add_executable(Direct directsum_example.cpp)
-  target_link_libraries(Direct cabanacore)
-  add_executable(Ewald ewaldsum_example.cpp)
-  target_link_libraries(Ewald cabanacore)
+if(NOT Cabana_ENABLE_MPI)
+  return()
 endif()
+  
+add_executable(Direct directsum_example.cpp)
+target_link_libraries(Direct cabanacore)
+add_executable(Ewald ewaldsum_example.cpp)
+target_link_libraries(Ewald cabanacore)
 
 if(Cabana_ENABLE_Cuda)
   find_package(CUDA REQUIRED)


core/example/longrange/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
-if(Cabana_ENABLE_MPI)
-  add_executable(Direct directsum_example.cpp)
-  target_link_libraries(Direct cabanacore)
-  add_executable(Ewald ewaldsum_example.cpp)
-  target_link_libraries(Ewald cabanacore)
+if(NOT Cabana_ENABLE_MPI)
+  return()
```
dalg24: Come on :@

dalg24: Don't include the subdirectory then

junghans: :-)

junghans: We do the same in a couple of places:
```
git grep 'return()' -- '*CMakeLists.txt'
core/example/CMakeLists.txt:  return()
core/fortran-example/tutorial/05_parallel_for/CMakeLists.txt:    return()
core/fortran-example/tutorial/06_kokkos_parallel_for/CMakeLists.txt:  return()
core/performance_test/peakflops/CMakeLists.txt:  return()
core/performance_test/peakflops/CMakeLists.txt:    return()
core/unit_test/CMakeLists.txt:  return()
```

junghans: I prefer this way as it localizes the restriction more, but I am happy to exclude the subdir.

dalg24: Fine

PR #199, opened 2020-02-10
Always build with existing kokkos devices

sslattery:
Updates the build to always use existing Kokkos devices rather than a user specified subset:

- The macros `Cabana_ENABLE_DEVICETYPE` where `DEVICETYPE` is `Cuda`, `OpenMP`, etc. no longer exist. Use `KOKKOS_ENABLE_DEVICETYPE` instead to use macros in an application which delineate by a device type
- New macros `Cabana_REQUIRE_DEVICETYPE` can be set in the configuration by an application to ensure that their required device types are available - the configure will fail if they are not
- Code and tests have been updated to use the new macros as well as some new Kokkos CMake utilities to check for a give device type


CMakeLists.txt
```diff
@@ -20,59 +20,48 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
...
 include(GNUInstallDirs)
 
+# allow users to directly point to a kokkos install
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
-find_package(Kokkos REQUIRED)
-
-option(Cabana_ENABLE_Serial "Build Cabana with Serial support" ON)
+set(Kokkos_INSTALL_PATH "" CACHE PATH "Kokkos installation location" )
+if(Kokkos_INSTALL_PATH)
+  find_package(Kokkos 3 PATHS ${Kokkos_INSTALL_PATH} NO_DEFAULT_PATH)
```
dalg24: This change is not advertised in your description.  Why do we want this?

CMakeLists.txt
```diff
@@ -20,59 +20,48 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
...
+endif()
+if(NOT Kokkos_FOUND)
+  find_package(Kokkos 3 REQUIRED)
+endif()
 
-option(Cabana_ENABLE_Pthread "Build Cabana with Pthread support" OFF)
-if( Cabana_ENABLE_Pthread )
-  find_package(Threads)
+# check user requirements
```
dalg24: This can/should be a loop.

CMakeLists.txt
```diff
@@ -20,59 +20,48 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
...
 
 option(Cabana_ENABLE_TESTING "Build tests" OFF)
 if(Cabana_ENABLE_TESTING)
   enable_testing()
 endif()
 
-find_package(Doxygen)
-if(Doxygen_FOUND)
-  doxygen_add_docs(doxygen core/src)
+option(Cabana_ENABLE_DOXYGEN "Build doxygen" OFF)
```
dalg24:
```suggestion
option(Cabana_ENABLE_DOXYGEN "Build doxygen" ON)
```
Must default to `ON` if you want to prevent the previous behavior.  Also not advertised in the description.  This option feels unnecessary to me.

core/src/CabanaCore_config.hpp.cmakein
```diff
@@ -7,12 +7,4 @@
 
 #cmakedefine Cabana_ENABLE_MPI
 
-#cmakedefine Cabana_ENABLE_Serial
-
-#cmakedefine Cabana_ENABLE_Pthread
-
-#cmakedefine Cabana_ENABLE_OpenMP
-
-#cmakedefine Cabana_ENABLE_Cuda
```
dalg24: This might break code downstream silently.  Just making sure you realize that.

CMakeLists.txt
sslattery: oh sorry will document - a user requested this so they could specify a Kokkos install outside of the prefix path in case they had multiple Kokkos variants installed. this would allow them to select the correct install in that case

CMakeLists.txt
sslattery: will do

sslattery: fixed

CMakeLists.txt
sslattery: Another user request to reduce build times by making this optional - I can default this to `ON` though. Also documented in the PR description now.

core/src/CabanaCore_config.hpp.cmakein
sslattery: Yes I agree this is a backwards compatibility change for the upcoming 1.0 release - documenting this in the PR notes as well.

CMakeLists.txt
dalg24: I would argue the user can prepend to the CMake search path and that option is not absolutely necessary.

sslattery: I have pinged the user who made the request to comment here on their needs

CMakeLists.txt
dalg24: This would only affect configuration and I doubt the user can tell the difference.  The documentation is only built if one run `make doxygen`, it is not part of the default build target.

CMakeLists.txt
junghans: Yeah, I would just use `CMAKE_PREFIX_PATH` for that. Also, isn't there `Kokkos_ROOT` exactly for that purpose?

core/example/longrange/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
-  
+
 add_executable(Direct directsum_example.cpp)
 target_link_libraries(Direct cabanacore)
 add_executable(Ewald ewaldsum_example.cpp)
 target_link_libraries(Ewald cabanacore)
 
-if(Cabana_ENABLE_Cuda)
+if(Kokkos_ENABLE_CUDA)
   find_package(CUDA REQUIRED)
```
junghans: Do we still need to detect `CUDA` or is `find_package(Kokkos)` pull that in?

CMakeLists.txt
```diff
@@ -20,61 +20,43 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
...
 option(Cabana_ENABLE_TESTING "Build tests" OFF)
 if(Cabana_ENABLE_TESTING)
   enable_testing()
 endif()
 
-find_package(Doxygen)
-if(Doxygen_FOUND)
-  doxygen_add_docs(doxygen core/src)
+# enable doxygen
+option(Cabana_ENABLE_DOXYGEN "Build doxygen" ON)
```
junghans: Why? One can just use `CMAKE_DISABLE_FIND_PACKAGE_Doxygen=ON` if you really really don't want to detect doxygen.

CMakeLists.txt
```diff
@@ -20,61 +20,43 @@ set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11
...
-endif()
-message(STATUS "Enable Devices: ${ENABLE_DEVICES}")
-
-# if we are using Cuda, also add a "CudaUVM" device to use that memory space
-# in the test suite. Also enable a Cabana UVM option so the tests get triggered.
-foreach(_device ${ENABLE_DEVICES})
-  if(_device STREQUAL Cuda)
-    list(APPEND CABANA_SUPPORTED_DEVICES CudaUVM)
-    set(Cabana_ENABLE_CudaUVM ON)
+  option(Cabana_REQUIRE_${_device} "Build Cabana with required Kokkos ${device} support" OFF)
```
junghans: Do we really need the `Cabana_REQUIRE_*` options, why aren't we just using whatever kokkos has.

CMakeLists.txt
junghans: Agreed.

core/example/longrange/CMakeLists.txt
dalg24: We do.  Kokkos does not link to Cuda FFT.

CMakeLists.txt
sslattery: Agreed - I am removing this

CMakeLists.txt
sslattery: OK - removing as well

CMakeLists.txt
sslattery: See the discussion above - I have removed this change.

CMakeLists.txt
sslattery: I actually thought this was a good idea. If we are going to automatically build Cabana functionality with all the support the existing Kokkos backends this allows a user to verify at configure time that those backends are actually available. This would especially be nice if someone else installed a Kokkos version and you weren't 100% sure what it supported. Note that this option is optional

sslattery:
@dalg24 @junghans anything here not yet resolved to your satisfaction?

sslattery:
> Do we need to modify `.gitlab-ci.yml` as well?

Already updated!

sslattery:
> > Do we need to modify `.gitlab-ci.yml` as well?
> 
> Already updated!

For reference here is the current version:
```
test:default:
  image: ubuntu
  script:
    - apt-get update
    - apt-get install -y gcovr gfortran cmake g++ git bc
    - git clone --depth=1 https://github.com/kokkos/kokkos.git && 
      pushd kokkos &&
      mkdir build &&
      pushd build &&
      cmake -DCMAKE_INSTALL_PREFIX=$HOME/kokkos -DKokkos_ENABLE_SERIAL=ON -DKokkos_ENABLE_OPENMP=ON .. &&
      make -j2 &&
      make install &&
      popd &&
      popd
    - mkdir build && cd build &&
      cmake -DCMAKE_PREFIX_PATH=$HOME/kokkos -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_EXAMPLES=ON .. &&
      make -k -O -j2 VERBOSE=1 &&
      make test
```
So it looks like the new uppercase device types are being used.

scripts/jenkins/build.sh
```diff
@@ -9,9 +9,9 @@ cmake \
   -D CMAKE_PREFIX_PATH=$KOKKOS_DIR \
   -D MPIEXEC_PREFLAGS="--allow-run-as-root;--mca;btl_smcuda_use_cuda_ipc;0" \
   -D Cabana_ENABLE_MPI=ON \
-  -D Cabana_ENABLE_Cuda=ON \
-  -D Cabana_ENABLE_Serial=OFF \
-  -D Cabana_ENABLE_OpenMP=OFF \
+  -D Cabana_REQUIRE_CUDA=ON \
+  -D Cabana_REQUIRE_SERIAL=OFF \
+  -D Cabana_REQUIRE_OPENMP=OFF \
```
dalg24: `Cabana_REQUIRE_<BACKEND>=OFF` has no effect.

sslattery: good catch - fixed

PR #202, opened 2020-02-20
Update Jenkins CI

dalg24:
Cleaning up the Dockerfile and changing the way CI is setup to use a Jenkinsfile.


dalg24:
Ready for review

dalg24:
I disabled the old CI build so we need to drop the requirement on `continuous-integration/jenkin-ci/pr` check passing.

scripts/jenkins/Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM nvidia/cuda:10.1-devel-ubuntu18.04
+ARG BASE=nvidia/cuda:10.2-devel
```
sslattery: Above says `stage('CUDA-10.1-NVCC-DEBUG')` but this is 10.2?

dalg24: Typo

PR #208, opened 2020-03-05
[WIP] Use ArborX to compute the particle neighbor list

dalg24:


core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -0,0 +1,207 @@
...
+    Kokkos::View<int *, MemorySpace> col_ind;
+    Kokkos::View<int *, MemorySpace> row_ptr;
+    typename MemorySpace::size_type shift;
+    typename MemorySpace::size_type total;
+};
+
+} // namespace Impl
+
+template <typename DeviceType, typename Slice, typename Tag>
+auto makeNeighborList( Tag, Slice const &coordinate_slice,
```
streeve: I think we need something to differentiate this from other neighbor construction

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -0,0 +1,207 @@
...
+    {
+        if ( getData( predicate ) > i )
+        {
+            out( i );
+        }
+    }
+};
+
+template <typename MemorySpace, typename Tag>
+struct CrsGraph
```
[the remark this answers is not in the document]
dalg24: Done

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -0,0 +1,207 @@
...
+namespace Cabana
+{
+namespace Experimental
+{
+namespace Impl
+{
+
+// Custom callback for ArborX::BVH::query()
+template <typename Tag>
+struct NeighborDiscriminator;
```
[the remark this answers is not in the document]
dalg24: I appended `Callback` to it.

core/src/Cabana_Experimental_NeighborList.hpp
dalg24: What name do you suggest?

streeve: `makeArborXList` or `makeTreeList`?

dalg24: It constructs a neighbor list so the name is appropriate in my opinion.

We could
* either append something to denote what algo/library (but don't see the point because it would not be helpful in a generic context),
* or pass yet another tag so it can be overloaded for the Verlet list.

streeve: After discussion, I agree with merging this as is and improving/combining the neighbor list storage/generation in a later PR

PR #216, opened 2020-03-31
Remove duplicate definition of CABANA_SUPPORTED_DEVICES

dalg24:
https://github.com/ECP-copa/Cabana/blob/3d340ddabecfed2c56d59a3352f08b5f462ba086/CMakeLists.txt#L33-L34


junghans:
@sslattery here: https://github.com/ECP-copa/Cabana/blob/master/CMakeLists.txt#L23

PR #218, opened 2020-03-31
Only pass non-empty parallel labels

streeve:
~~Deprecate~~ Only forward labels with `simd_parallel_for` and `neighbor_parallel_for` if not empty ~~and add required label versions~~. This was discussed in #183

### DIFF
--- a/core/src/Cabana_Parallel.hpp
+++ b/core/src/Cabana_Parallel.hpp
@@ -104,17 +104,23 @@ inline void simd_parallel_for(
 
     using index_type = typename team_policy::index_type;
 
-    Kokkos::parallel_for(
-        str, dynamic_cast<const team_policy &>( exec_policy ),
-        KOKKOS_LAMBDA( const typename team_policy::member_type &team ) {
-            index_type s = team.league_rank() + exec_policy.structBegin();
-            Kokkos::parallel_for(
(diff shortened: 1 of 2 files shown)


sslattery:
> Did you consider only passing the label if it is not empty instead of requiring it?

What is the current Kokkos strategy for labels? Are they always required now?

dalg24:
> > Did you consider only passing the label if it is not empty instead of requiring it?
> 
> What is the current Kokkos strategy for labels? Are they always required now?

No

streeve:
> Did you consider only passing the label if it is not empty instead of requiring it?

Good point. I can do that - is it worth keeping the interface with name first?

core/src/Cabana_Parallel.hpp
```diff
@@ -270,23 +281,27 @@ inline void neighbor_parallel_for(
...
+            for ( index_type a = n + 1; a < nn; ++a )
+            {
+                const index_type k =
+                    neighbor_list_traits::getNeighbor( list, i, a );
+                Impl::functorTagDispatch<work_tag>( functor, i, j, k );
             }
-        } );
+        }
+    };
+    if ( str == "" )
```
sslattery, line 301: Use `str.empty()`

streeve:
Added `str.empty()` and rebased on master

PR #220, opened 2020-04-03
Add neighbor_parallel_reduce

streeve:
This adds `neighbor_parallel_reduce` options that mirror the `neighbor_parallel_for`. I'll use these to give energy calculation the same flexibility as current force calculation (particularly where force and energy computes are fused). 

I didn't add a `simd_parallel_reduce`, but can if there's a use case

### DIFF
--- a/core/src/Cabana_Parallel.hpp
+++ b/core/src/Cabana_Parallel.hpp
@@ -609,8 +609,8 @@ inline void neighbor_parallel_for(
   the Kokkos::parallel_reduce called by this code and can be used for
   identification and profiling purposes.
 */
-template <class FunctorType, class NeighborListType, class... ExecParameters,
-          class ReduceType>
+template <class FunctorType, class NeighborListType, class ReduceType,
+          class... ExecParameters>
 inline void neighbor_parallel_reduce(
     const Kokkos::RangePolicy<ExecParameters...> &exec_policy,
     const FunctorType &functor, const NeighborListType &list,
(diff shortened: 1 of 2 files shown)


core/src/Cabana_Parallel.hpp
```diff
@@ -554,6 +576,422 @@ inline void neighbor_parallel_for(
...
+            for ( index_type a = n + 1; a < nn; ++a )
+            {
+                const index_type k =
+                    neighbor_list_traits::getNeighbor( list, i, a );
+                Impl::functorTagDispatch<work_tag>( functor, i, j, k, ival );
+            }
+        }
+    };
+    if ( str.empty() )
+        Kokkos::parallel_reduce( exec_policy, neigh_reduce, reduce_val );
```
sslattery, line 722: This seems to reduce a single value over all particles over all neighbors correct? I guess I though the MD use case was a per particle reduction over all neighbors. Hence if there are `N` particles you would get `N` reductions where the reduction is a force or something. What you have will almost work in that case - just replace the `parallel_reduce` with a `parallel_for` and have the user input a view/slice of values for each particle into which to write the reductions. Also note that the reducers can be multidimensional so you could calculate a force in 3 space dimensions or something in this way.

sslattery: OK after talking with Sam this functionality is needed for total reductions over neighbor operations (e.g. computing total energy in an MD calculculations). We will do what I described here in another PR which does something like `neighbor_parallel_for_reduce` or similar name with a custom result functor to allow the user to extract the particle-wise reductions.

PR #221, opened 2020-04-09
Update examples

streeve:
This PR does a few things - 

- Minor comment fixes
- Reordered a few tests to match the wiki, with advanced topics numbered with the main example
- Add print outs to tests without any
- Add `neighbor_parallel_for` example


core/example/tutorial/02_tuple/tuple_example.cpp
```diff
@@ -24,24 +24,24 @@ void tupleExample()
       of Cabana::Tuple that are different from std::tuple:
 
       *: Cabana tuples must consist of data of trivial types and classes. A
-         trivial type or class is one who's
+         trivial type or class is one:
```
rfbird, line 27: Just to throw a 3rd phrasing in the mix:

> is one which: 
> 
> 1) has contagious storage 
> 2) only supports..

rfbird:
How do we feel about having examples with the same numbers?

I find `05_slice_advanced_cuda` and `05_slice_advanced_openmp` a bit confusing, but at the same time I see why you did it...

core/example/tutorial/07_sorting/sorting_example.cpp
```diff
@@ -22,8 +22,8 @@ void sortingExample()
       In many algorithms we will want to bin or sort AoSoA data to improve
       computational performance. Binning and sorting can be achieved by using
       a user-defined comparator function or by using an auxiliary set of key
-      values (such as a slice) who's sorted/binned order will define the order
-      of the AoSoA. In this example we will demonstrate sorting an AoSoA a
+      values (such as a slice) whose sorted/binned order will define the order
```
sslattery, line 25: man my grammar is terrible

PR #228, opened 2020-05-05
Use C++14

sslattery:
C++14 is now required. As a test we now use the `index_sequence` as an implementation detail which is a feature available from C++14 onward.


sslattery:
I welcome debate/discussion on how we should enable C++14. I took the CXX_STANDARD route for a first attempt.

PR #229, opened 2020-05-05
Clean example/performance test directory structure

sslattery:
`example/` now sits in the top level directory. Performance tests existed in multiple locations so these are now in the same location.


rfbird:
Looks good to me. Seems like you need to update travis to reflect the new path too: https://github.com/sslattery/Cabana/blob/clean_example_dirs/.travis.yml#L91

sslattery:
> Looks good to me. Seems like you need to update travis to reflect the new path too: https://github.com/sslattery/Cabana/blob/clean_example_dirs/.travis.yml#L91

Yeah just caught that one - pushed a fix

PR #230, opened 2020-05-06
Team only VerletList build

streeve:
This adds an option for counting during the VerletList build with team threading only, in addition to the existing team vector. 

I started this to enable HIP VerletLists, but it should be a worthwhile optimization for some cases. With this, tests are only disabled with HIP for `neighbor_parallel_reduce` with team/vector parallelism and `VerletList` builds with vector reduces. 

I set `TeamVector` as the default to match the previous code, but then ended up setting `Team` explicitly for most examples/tests to work with HIP. 

The same thing could be done for the list fill portion of the build, but I thought I would get feedback first.


streeve:
Quick performance check on one V100 with `NeighborListMDPerfTest`, 10 iterations each:
```
Cutoff distance to minimum distance ratio: 4 and 10

Particles Neighbors TeamVector Team      Change
10000     100       124.5ms    123.9ms   0.48%
100000    100       1378.5ms   1359.4ms  1.4%
10000     1000      1440.7ms   1443ms    -0.15%
```
This only affects the count, not fill, but still a small differences for these few cases; there was even less difference on the CPU.

sslattery:
Some warnings in the HIP build:

```
In file included from /var/jenkins/workspace/Cabana_PR-230/build/core/unit_test/HIP/tstNeighborList_HIP.cpp:2:
/var/jenkins/workspace/Cabana_PR-230/core/unit_test/tstNeighborList.hpp:808:12: warning: unused variable 'team_sum' [-Wunused-variable]
    double team_sum = 0;
           ^
/var/jenkins/workspace/Cabana_PR-230/core/unit_test/tstNeighborList.hpp:870:12: warning: unused variable 'vector_sum' [-Wunused-variable]
    double vector_sum = 0;
           ^
/var/jenkins/workspace/Cabana_PR-230/core/unit_test/tstNeighborList.hpp:869:12: warning: unused variable 'team_sum' [-Wunused-variable]
    double team_sum = 0;
           ^
In file included from /var/jenkins/workspace/Cabana_PR-230/build/core/unit_test/SERIAL/tstNeighborList_SERIAL.cpp:2:
/var/jenkins/workspace/Cabana_PR-230/core/unit_test/tstNeighborList.hpp:808:12: warning: unused variable 'team_sum' [-Wunused-variable]
    double team_sum = 0;
           ^
/var/jenkins/workspace/Cabana_PR-230/core/unit_test/tstNeighborList.hpp:869:12: warning: unused variable 'team_sum' [-Wunused-variable]
    double team_sum = 0;
           ^
/var/jenkins/workspace/Cabana_PR-230/core/unit_test/tstNeighborList.hpp:870:12: warning: unused variable 'vector_sum' [-Wunused-variable]
    double vector_sum = 0;
```

core/src/Cabana_VerletList.hpp
```diff
@@ -375,6 +382,108 @@ struct VerletListBuilder
...
+        // working on.
+        int cell = team.league_rank();
+
+        // Get the stencil for this cell.
+        int imin, imax, jmin, jmax, kmin, kmax;
+        cell_stencil.getCells( cell, imin, imax, jmin, jmax, kmin, kmax );
+
+        // Operate on the particles in the bin.
+        std::size_t b_offset = bin_data_1d.binOffset( cell );
+        Kokkos::parallel_for(
```
sslattery, line 401: So 1 extra level of parallelism works with HIP but the second does not?

core/src/Cabana_VerletList.hpp
```diff
@@ -375,6 +382,108 @@ struct VerletListBuilder
...
+                                        linked_cell_list.binOffset( i, j, k );
+
+                                    for ( int n = 0;
+                                          n <
+                                          linked_cell_list.binSize( i, j, k );
+                                          n++ )
+                                    {
+                                        //  Get the true id of the candidate
+                                        //  neighbor.
+                                        std::size_t nid =
```
sslattery, line 440: It would be nice to abstract this part so we don't have to repeat all of this code. The loop body is mostly the same - just different how the loop is called (serial vs. vector). Can we separate the loop body into another function and abstract how we call the loop? That should let us combine a lot of code. This function is then templated on the type tag and the loop function will have explicit instantiations.

core/src/Cabana_VerletList.hpp
```diff
@@ -190,8 +187,20 @@ struct LinkedCellStencil
 //---------------------------------------------------------------------------//
 // Verlet List Builder
 //---------------------------------------------------------------------------//
+template <class Tag>
+struct CountNeighborsTag;
+
+template <>
+struct CountNeighborsTag<TeamVectorOpTag>
```
sslattery, line 194: Per my comments on looping below you might need an alias in here like `using tag = TeamVectorOpTag` to facilitate meta programming.

core/unit_test/tstNeighborList.hpp
```diff
@@ -903,15 +920,25 @@ TEST( TEST_CATEGORY, linked_cell_stencil_test ) { testLinkedCellStencil(); }
 //---------------------------------------------------------------------------//
 TEST( TEST_CATEGORY, verlet_list_full_test )
 {
-    testVerletListFull<Cabana::VerletLayoutCSR>();
-    testVerletListFull<Cabana::VerletLayout2D>();
+    testVerletListFull<Cabana::VerletLayoutCSR, Cabana::TeamOpTag>();
+    testVerletListFull<Cabana::VerletLayout2D, Cabana::TeamOpTag>();
+
+#ifndef KOKKOS_ENABLE_HIP
```
sslattery, line 926: Good - we will want this for now. add a `fixme` note so we can remove this when HIP is ready to go.

core/src/Cabana_VerletList.hpp
streeve: Hierarchical reductions don't work with HIP yet

core/src/Cabana_VerletList.hpp
streeve: Agreed - fixed this

core/unit_test/tstNeighborList.hpp
streeve: Done

streeve:
Ok, I rebased this, removed the duplicate code, and it should be ready for review

core/src/Cabana_VerletList.hpp
```diff
@@ -14,14 +14,11 @@
 
 #include <Cabana_LinkedCellList.hpp>
 #include <Cabana_NeighborList.hpp>
+#include <Cabana_Parallel.hpp> // TeamOpTag, TeamVectorOpTag
```
sslattery, line 17: Probably don't need these comments here

core/src/Cabana_VerletList.hpp
```diff
@@ -319,52 +317,8 @@ struct VerletListBuilder
...
-
-                                                // If within the cutoff add to
-                                                // the count.
-                                                if ( dist_sqr <= rsqr )
-                                                    local_count += 1;
-                                            }
-                                        },
-                                        cell_count );
+                                    neighbor_reduce( team, pid, i, j, k,
+                                                     cell_count, BuildOpTag() );
```
sslattery, line 321: nice

core/src/Cabana_VerletList.hpp
```diff
@@ -375,6 +329,91 @@ struct VerletListBuilder
...
+    {
+        // Cache the particle coordinates (again).
+        double x_p = position( pid, 0 );
+        double y_p = position( pid, 1 );
+        double z_p = position( pid, 2 );
+
+        std::size_t a_offset = linked_cell_list.binOffset( i, j, k );
+
+        Kokkos::parallel_reduce(
+            Kokkos::ThreadVectorRange( team,
```
sslattery, line 346: I think you can have another reduce abstraction here because this inner. A lot of code is still repeated between these two functions. Perhaps at least abstract the loop body to take in a `local_count` by reference - you can then either just pass in which ever value it reduces to.

core/unit_test/tstNeighborList.hpp
```diff
@@ -786,31 +796,33 @@ void testFirstNeighborParallelReduce()
...
     // Test the list parallel operation by adding a value from each neighbor
     // to the particle and compare to counts.
     auto sum_op = KOKKOS_LAMBDA( const int i, const int n, double &sum )
     {
         sum += positions( i, 0 ) + positions( n, 0 );
     };
     Kokkos::RangePolicy<TEST_EXECSPACE> policy( 0, aosoa.size() );
+
+    // Do the reductions.
+    double test_sum = 0;
```
dalg24, line 815: I would move the definition closer to its point of use in the block starting line 828

core/src/Cabana_VerletList.hpp
streeve: Took care of this in a similar way

streeve:
@sslattery can you do another review?

Issue #231, opened 2020-05-06
Support MPI for CUDA and HIP when not GPU-aware


sslattery:
We should allow for versions of MPI to be used that aren't GPU-aware. My proposal to do this is to add an extra optional template parameter to the communication classes which describes the communication memory space. We then simply mirror the user data into the communication memory space prior to calling MPI.

Some sensible defaults will be assuming CUDA-aware MPI and using the host space for HIP. A user can always change this in their implementation depending on their dependencies. We could then add a CMake option which would change these defaults for the user so they could easily configure for the given system they are using.

Once a HIP-aware MPI becomes available we can use this as the HIP default.

rfbird:
bump on this, I just ran into it when making sure everything passed but I wasn't using an GPU aware MPI

streeve:
Very stale and I don't think necessary - a manual copy to the host will work if really necessary (as in the comm benchmarks)

PR #232, opened 2020-05-06
Require MPI

sslattery:
Cajita requires MPI which means a large portion of the library will need MPI to function when it is merged. Would like thoughts/feedback here on the approach. Note that HIP is still disabled with MPI until #231 is resolved.


sslattery:
> I am against requiring MPI

Can you elaborate on your objection?

dalg24:
I feel like it is still reasonable to want to use Cabana w/o MPI.  We had discussed it at long before we implemented it this way.  We can just disable Cajita if MPI is not found for now and re-evaluate later if necessary.

sslattery:
That is not unreasonable - should be easy to control the Cajita build. As a counter point I could see people wanting to use Cajita without MPI as well but that would require some signficant refactoring. As a result any Cajita+Cabana algorithms would require MPI as well. 

@rfbird @junghans thoughts?

rfbird:
In an ideal world we would make it possible to build Cajita without MPI, but I've seen that's a none trivial code effort with minimal upside. I'm OK with No MPI => No Cajita.

sslattery:
OK - we will proceed with the merger without an MPI requirement

PR #233, opened 2020-05-06
Add HIP MPI

streeve:
Use the same strategy with HIP as CUDA. 

Worked with rocm/3.3 and openmpi/4.0.0 on MI25 at LLNL

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -50,6 +50,7 @@ pipeline {
                               -D CMAKE_BUILD_TYPE=Debug \
                               -D CMAKE_CXX_COMPILER=hipcc \
                               -D CMAKE_CXX_FLAGS="-Wall -Wextra -Wpedantic" \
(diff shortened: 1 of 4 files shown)


sslattery:
@dalg24 is MPI explicitly not included in the HIP Jenkins build stage?

sslattery:
> @dalg24 is MPI explicitly not included in the HIP Jenkins build stage?

We are going to try adding it to the build stage and see what happens - @streeve has had some luck with it locally

streeve:
Ok, this passed. The `MPIEXEC_MAX_NUMPROCS` is getting populated with what I assume is the CPU count, but otherwise it looks fine

sslattery:
Looks like Jenkins passed - just not reporting the status

dalg24:
> Looks like Jenkins passed - just not reporting the status

It did report success.

PR #234, opened 2020-05-07
Enable ArborX in CUDA and HIP builds on Jenkins

dalg24:


dalg24:
I just remembered we did not merge the HIP support in ArborX's `master` yet.
We are waiting for the next ROCm release (should be later this month or early next one) so we don't have to introduce a bunch of workarounds.
I will re-add later.

dalg24:
OK this is ready.  I fixed all the warnings.

core/src/Cabana_Tuple.hpp
```diff
@@ -144,7 +144,7 @@ struct Tuple<MemberTypes<Types...>> : SoA<MemberTypes<Types...>, 1>
 {
     using base = SoA<MemberTypes<Types...>, 1>;
 
-    KOKKOS_FORCEINLINE_FUNCTION Tuple() = default;
+    KOKKOS_DEFAULTED_FUNCTION Tuple() = default;
```
sslattery, line 147: Nice - I had seen these warnings but didn't know about this macro

core/unit_test/tstNeighborList.hpp
```diff
@@ -84,19 +84,19 @@ convert_crs_graph_to_verlet_list(
             auto const last = crs_graph.row_ptr( i + 1 );
             counts( shift + i ) = last - first;
         } );
-    int const max_entries_per_row = ArborX::max( counts );
+    using ExecutionSpace = typename DeviceType::execution_space;
+    int const max_entries_per_row = ArborX::max( ExecutionSpace{}, counts );
```
sslattery, line 88: is this your own portable implementation of `std::max`?

core/unit_test/tstNeighborList.hpp
```diff
@@ -586,7 +586,7 @@ void testFirstNeighborParallelFor()
     ListType nlist( Cabana::slice<0>( aosoa ), 0, aosoa.size(), test_radius,
                     cell_size_ratio, grid_min, grid_max );
 #else
-    (void)cell_size_ratio;
+    std::ignore = cell_size_ratio;
     using device_type = TEST_MEMSPACE; // sigh...
```
sslattery, line 590: Yeah we need to fix the use of `Device` in our classes. This stuff was written when we thought that was the best design pattern.

core/unit_test/tstNeighborList.hpp
```diff
@@ -662,7 +662,7 @@ void testVerletListFullPartialRange()
         Cabana::slice<0>( aosoa ), 0, num_ignore, test_radius, cell_size_ratio,
         grid_min, grid_max );
 #else
-    (void)cell_size_ratio;
+    std::ignore = cell_size_ratio;
```
sslattery, line 665: Maybe just move the definition of `cell_size_ratio` into the `#if` block above because it is a parameter specific to the `VerletList`?

core/unit_test/tstNeighborList.hpp
dalg24: No,  it performs a parallel reduce and returns the maximum element in a view.

core/unit_test/tstNeighborList.hpp
dalg24: Warnings are gone as far as I can tell

PR #238, opened 2020-05-21
Team only VerletList fill

streeve:
This completes #230 so that both the count and fill parts of the `VerletList` build uses the team and team vector tags. 

~~I added a `chunk_size` of 1 to pass tests for serial loops (before set to 4 for all cases), but maybe there's a better way to deal with this.~~
I added a `Kokkos::single` where I didn't expect to need it for the nested serial loops.

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -335,7 +335,7 @@ struct VerletListBuilder
             } );
     }
 
-    // Neighbor count team vector operator (only used for CSR lists).
+    // Neighbor count team vector loop (only used for CSR lists).
     KOKKOS_INLINE_FUNCTION void
     neighbor_reduce( const typename CountNeighborsPolicy::member_type &team,
                      const std::size_t pid, const double x_p, const double y_p,
@@ -350,7 +350,7 @@ struct VerletListBuilder
             cell_count );
     }
 
-    // Neighbor count serial operator (only used for CSR lists).
+    // Neighbor count serial loop (only used for CSR lists).
     KOKKOS_INLINE_FUNCTION
     void neighbor_reduce( const typename CountNeighborsPolicy::member_type,
                           const std::size_t pid, const double x_p,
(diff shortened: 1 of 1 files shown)


core/src/Cabana_VerletList.hpp
```diff
@@ -262,6 +265,12 @@ struct VerletListBuilder
 
         // We will use the square of the distance for neighbor determination.
         rsqr = neighborhood_radius * neighborhood_radius;
+
+        // Set chunk size to one if using team/serial loops
+        if ( std::is_same<BuildOpTag, TeamOpTag>::value )
```
sslattery, line 270: This could be done statically

core/src/Cabana_VerletList.hpp
```diff
@@ -685,11 +709,12 @@ class VerletList
         // count and fill at the same time, unless the array size is exceeded,
         // at which point only counting is continued to reallocate and refill.
         typename builder_type::FillNeighborsPolicy fill_policy(
-            builder.bin_data_1d.numBin(), Kokkos::AUTO, 4 );
+            builder.bin_data_1d.numBin(), Kokkos::AUTO, builder.chunk_size );
```
sslattery, line 712: So the test was failing in serial because it was trying to launch 4 threads when none were available?

streeve: It failed because it was running through the inner loop 4 times, quadrupling the neighbor counts/lists

streeve:
This is the same timing comparison as #230 slight speedup to use team only. 

I was surprised that now both options are faster than before any `BuildTag` options:

```
Cell size to cutoff distance ratio: 1
Cutoff distance to minimum distance ratio: 4 and 10

Particles Neighbors TeamVector Team      Diff   No options Change from before tag options
10000     ~100      104ms      102ms     1.9%   124.5ms    18%
100000    ~100      1166.3ms   1150.7ms  1.3%   1378.5ms   16%
10000     ~1000     1129.1ms   1124.6ms  0.40%  1440.7ms   21%
```

streeve:
Ok @sslattery everything passed with the updated version using `single` instead of changing chunk sizes

sslattery:
> This is the same timing comparison as #230, again with a slight speedup to use team only.
 Nice - V100?

PR #239, opened 2020-05-22
Use per target flags

### LINKED ISSUE #149: CMake Clean up
Move to per-target based flags everywhere.

streeve:
Closes #149 (unless I missed something)

### DIFF
--- a/core/performance_test/peakflops/CMakeLists.txt
+++ b/core/performance_test/peakflops/CMakeLists.txt
@@ -4,41 +4,33 @@ if(NOT Cabana_ENABLE_TESTING)
   return()
 endif()
 
-set(Cabana_PERFORMANCE_EXPECTED_FLOPS "0" CACHE STRING "Value for expected flops in performance tests")
-set(Cabana_PERFORMANCE_ERROR_MARGIN "0" CACHE STRING "Error margin to use in performance tests")
+# For intel haswell, broadwell: set expected flops/cycle = 32, veclength=8, iterations=2e6
+# For intel skylake           : set expected flops/cycle = 64, veclength=16, iterations=2e8
+set(Cabana_PERFORMANCE_EXPECTED_FLOPS "32" CACHE STRING "Value for expected flops in performance tests")
+set(Cabana_PERFORMANCE_ERROR_MARGIN "0.9" CACHE STRING "Error margin to use in performance tests")
 set(Cabana_PERFORMANCE_VECLENGTH "8" CACHE STRING "Vector Length to use when running expected flops")
-set(Cabana_PERFORMANCE_ITERATIONS "10" CACHE STRING "Number of iterations to repeat the test for (increases flops)")
-set(Cabana_PERFORMANCE_SEED "0" CACHE STRING "Seed for performance test RNG")
+set(Cabana_PERFORMANCE_ITERATIONS "2000000" CACHE STRING "Number of iterations to repeat the test for (increases flops)")
+set(Cabana_PERFORMANCE_SEED "76843802738543" CACHE STRING "Seed for performance test RNG")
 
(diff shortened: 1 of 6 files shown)


streeve:
@rfbird let me know if you have a different idea about defaults

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -1,48 +1,44 @@
...
-set(Cabana_PERFORMANCE_VECLENGTH "" CACHE STRING "Vector Length to use when running expected flops")
-set(Cabana_PERFORMANCE_ITERATIONS "" CACHE STRING "Number of iterations to repeat the test for (increases flops)")
-set(Cabana_PERFORMANCE_SEED "" CACHE STRING "Seed for performance test RNG")
+set(Cabana_PERFORMANCE_EXPECTED_FLOPS "0" CACHE STRING "Value for expected flops in performance tests")
+set(Cabana_PERFORMANCE_ERROR_MARGIN "0" CACHE STRING "Error margin to use in performance tests")
+set(Cabana_PERFORMANCE_VECLENGTH "8" CACHE STRING "Vector Length to use when running expected flops")
+set(Cabana_PERFORMANCE_ITERATIONS "10" CACHE STRING "Number of iterations to repeat the test for (increases flops)")
+set(Cabana_PERFORMANCE_SEED "0" CACHE STRING "Seed for performance test RNG")
 
 if(NOT Cabana_PERFORMANCE_EXPECTED_FLOPS STREQUAL "")
```
junghans, line 13: If it is always define, couldn't we drop this check?

core/performance_test/peakflops/CMakeLists.txt
```diff
@@ -1,48 +1,44 @@
...
   return()
 endif()
 
-set(Cabana_PERFORMANCE_EXPECTED_FLOPS "" CACHE STRING "Value for expected flops in performance tests")
-set(Cabana_PERFORMANCE_ERROR_MARGIN "" CACHE STRING "Error margin to use in performance tests")
-set(Cabana_PERFORMANCE_VECLENGTH "" CACHE STRING "Vector Length to use when running expected flops")
-set(Cabana_PERFORMANCE_ITERATIONS "" CACHE STRING "Number of iterations to repeat the test for (increases flops)")
-set(Cabana_PERFORMANCE_SEED "" CACHE STRING "Seed for performance test RNG")
+set(Cabana_PERFORMANCE_EXPECTED_FLOPS "0" CACHE STRING "Value for expected flops in performance tests")
+set(Cabana_PERFORMANCE_ERROR_MARGIN "0" CACHE STRING "Error margin to use in performance tests")
```
rfbird, line 8: I think the defaults should probably match what's in the code, here:

https://github.com/ECP-copa/Cabana/blob/dd7101b63411cad5208dad0299e7c417ea2a7dcf/core/performance_test/peakflops/common.h

 and 

https://github.com/ECP-copa/Cabana/blob/0f4de9cad7feadc52d1ee15e11fd0b51f7294e78/core/performance_test/peakflops/01_cpp_simple.cpp#L140

If we're hoisting the default to here, we should probably tidy those bits of the code too?

streeve: Much cleaner, but it does now use the same number of iterations per test

core/unit_test/CMakeLists.txt
```diff
@@ -64,7 +62,8 @@ macro(Cabana_add_tests)
       )
       set(_target ${_test}_test_${_device})
       add_executable(${_target} ${_file} ${CABANA_UNIT_TEST_MAIN})
-      target_include_directories(${_target} PUBLIC ${_dir})
+      target_include_directories(${_target} PUBLIC ${_dir}
```
junghans, line 65: It seems `PUBLIC` doesn't make sense for executables (https://reviews.llvm.org/D40823).

PR #243, opened 2020-06-03
Add ArborX to travis

streeve:
Also sets cmake variables for optional dependencies.


sslattery:
I see that ArborX code is being picked up in the doxygen build. I guess no way to see if the ArborX part of the neighbor list test is building/running because the test output isn't verbose?

streeve:
> I see that ArborX code is being picked up in the doxygen build. 

Any quick way to stop that?

> I guess no way to see if the ArborX part of the neighbor list test is building/running because the test output isn't verbose?

I think it's worth making the test output verbose. Any disagreement?

sslattery:
> > I see that ArborX code is being picked up in the doxygen build.
> 
> Any quick way to stop that?
> 
Thats not a bad thing - I as just scanning the build to see if stuff was there. Not a problem

> > I guess no way to see if the ArborX part of the neighbor list test is building/running because the test output isn't verbose?
> 
> I think it's worth making the test output verbose. Any disagreement?
Only downside I can see is that we expand the log file even more but I think it could be nice to see that extra info. I vote for adding a `-VV` to ctest. @junghans do you have an opinion on this?

junghans:
What kind of info do you hope to see?

sslattery:
> What kind of info do you hope to see?

In this case it would be nice to see that the ArborX sub-test in neighbor list is running/passing indicating the configuration was effective.

junghans:
Could you just make it into a separate test?

sslattery:
If you rebase on `master` it should fix the build errors on Jenkins

dalg24:
> If you rebase on `master` it should fix the build errors on Jenkins

I don't think that failure had anything to do with being rebased or not.  CI build on Jenkins tests the changes proposed merged into the master branch (unless they are conflicts in which case it uses the current HEAD)

sslattery:
> > If you rebase on `master` it should fix the build errors on Jenkins
> 
> I don't think that failure had anything to do with being rebased or not. CI build on Jenkins tests the changes proposed merged into the master branch (unless they are conflicts in which case it uses the current HEAD)

It looked like it died due to being out of resources so it seemed like the fix you put in would correct that.

PR #244, opened 2020-06-04
clean up hypre dependency install

sslattery:
The hypre dependency was not being correctly installed so downstream packages would have had to included hypre in their `CMAKE_PREFIX_PATH` even though it is an implementation detail in Cajita. This PR adds proper install for this.

PR #248, opened 2020-06-04
Enable generation of a 2D neighbor list using ArborX

dalg24:


sslattery:
weird this one is dying on building `HelloWorld` as well but in a different way than #244. Is this related to #245?

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -114,36 +114,77 @@ namespace Experimental
...
-            out( i );
+            out( primitive_index );
         }
     }
 };
 
-template <>
-struct NeighborDiscriminatorCallback<Cabana::HalfNeighborTag>
+template <typename Counts, typename Tag>
+struct NeighborDiscriminatorCallback2D_FirstPass
```
sslattery, line 154: By `FirstPass` and `SecondPass` do you mean these are called during the count and fill phases in ArborX? Is there something in the type tags that lets ArborX decide which functor to call when?

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -114,36 +114,77 @@ namespace Experimental
...
+    template <typename Predicate, typename OutputFunctor>
+    KOKKOS_FUNCTION void operator()( Predicate const &predicate,
+                                     int primitive_index,
+                                     OutputFunctor const & ) const
+    {
+        int const predicate_index = getData( predicate );
+        if ( CollisionFilter<Tag>::keep( predicate_index, primitive_index ) )
+        {
+            assert( count( predicate_index ) < (int)neighbors.extent( 1 ) );
+            neighbors( predicate_index, counts( predicate_index )++ ) =
```
sslattery, line 186: If I remember correctly the fill will happen in serial per predicate. In the future I guess `counts` would need to be an atomic view if hierarchical parallelism was used.

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -179,8 +220,59 @@ auto makeNeighborList( Tag, Slice const &coordinate_slice,
...
+
+    auto const predicates =
+        Impl::makePredicates( coordinate_slice, first, last, radius );
+
+    auto const n_queries = ArborX::Traits::
+        Access<decltype( predicates ), ArborX::Traits::PredicatesTag>::size(
+            predicates );
+
+    Kokkos::View<int *, DeviceType> counts( "counts", n_queries );
+    bvh.query(
```
sslattery, line 259: OK so you do your own 2-pass here - do you force ArborX to do the queries in 1-pass mode so this is effectively still just a 2-pass process as you are filling your own data structure?

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -205,13 +297,38 @@ class NeighborList<Experimental::CrsGraph<MemorySpace, Tag>>
...
     {
-        assert( (int)p >= 0 && p < crs_graph.total );
         assert( n < numNeighbor( crs_graph, p ) );
         p -= crs_graph.shift;
         return crs_graph.col_ind( crs_graph.row_ptr( p ) + n );
     }
 };
 
+template <typename MemorySpace, typename Tag>
+class NeighborList<Experimental::Dense<MemorySpace, Tag>>
```
sslattery, line 307: I like this `Dense` type name

core/src/Cabana_Experimental_NeighborList.hpp
sslattery: I guess you must check if there is enough space allocated by the user but you never fill the native data structure so ArborX will always do 1-pass because the internal count never goes past 0?

core/src/Cabana_Experimental_NeighborList.hpp
sslattery: I figured this out - see comments below

core/src/Cabana_Experimental_NeighborList.hpp
dalg24: Yes.  I should had a word of warning here.  You are right I am taking advantage of my knowledge of how the traverse is currently implemented here.

core/src/Cabana_Experimental_NeighborList.hpp
dalg24: Correct.  We detect that you never called the output functor and do a single pass.  Note that we are working on adding a new `BVH::query()` overload that does only takes the callback (no offsets and values).

core/src/Cabana_Experimental_NeighborList.hpp
```diff
@@ -114,36 +114,77 @@ namespace Experimental
...
+    using tag = ArborX::Details::InlineCallbackTag;
+    template <typename Predicate, typename OutputFunctor>
+    KOKKOS_FUNCTION void operator()( Predicate const &predicate,
+                                     int primitive_index,
+                                     OutputFunctor const & ) const
+    {
+        int const predicate_index = getData( predicate );
+        if ( CollisionFilter<Tag>::keep( predicate_index, primitive_index ) )
+        {
+            assert( count( predicate_index ) < (int)neighbors.extent( 1 ) );
```
dalg24, line 185: Oops.  You can tell I had `NDEBUG` defined on my local build...

core/src/Cabana_Experimental_NeighborList.hpp
dalg24: I added comments above the specialization to make this more obvious.

core/src/Cabana_Experimental_NeighborList.hpp
dalg24: I added a comment.

PR #250, opened 2020-06-08
Remove ghosted global index spaces

sslattery:
Some work with boundary conditions in Cajita revealed that we can't produce `IndexSpace` objects which contain `Global` indices in a `Ghost` decomposition and hence they have been removed. 

The reason for this is that `IndexSpace` covers a contiguous set of indices which can be satisfied for `Local` indices in both `Ghost` and `Own` decompositions as well as for `Global` indices in `Own` decompositions. However, when periodicity is used, we no longer have contiguous index sets of `Global` indices in `Ghost` decompositions for boundary blocks (the global indices wrap around the periodic boundary in the halo) and therefore we cannot make an `IndexSpace` of such indices. This makes sense as global indices are globally unique while local indices are not. Hence, the implementation of these was fundamentally a bug.

I am going to make another PR (#251) which builds a global-to-local and local-to-global indexer which should make life easier for implementing boundary conditions.

### DIFF
--- a/cajita/unit_test/tstLocalMesh.hpp
+++ b/cajita/unit_test/tstLocalMesh.hpp
@@ -569,8 +569,6 @@ void irregularTest( const std::array<int, 3> &ranks_per_dim )
         local_grid->indexSpace( Ghost(), Cell(), Local() );
     auto own_cell_global_space =
         local_grid->indexSpace( Own(), Cell(), Global() );
-    auto ghost_cell_global_space =
-        local_grid->indexSpace( Ghost(), Cell(), Global() );
     auto own_cell_local_space =
         local_grid->indexSpace( Own(), Cell(), Local() );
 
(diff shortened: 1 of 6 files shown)


sslattery:
retest this please

sslattery:
Note that Travis build passed but is not reporting status updates back. The Jenkins hardware is currently down.

sslattery:
retest this please

dalg24:
Retest this please

cajita/unit_test/tstLocalMesh.hpp
```diff
@@ -613,18 +611,24 @@ void irregularTest( const std::array<int, 3> &ranks_per_dim )
                      k_func( own_cell_global_space.max( Dim::K ) ) );
 
     EXPECT_FLOAT_EQ( ghost_lc_m( Dim::I ),
-                     i_func( ghost_cell_global_space.min( Dim::I ) ) );
+                     i_func( own_cell_global_space.min( Dim::I ) -
+                             own_cell_local_space.min( Dim::I ) ) );
```
streeve, line 615: Does it make sense to add a function to return ghost low and high corners for convenience?

sslattery: The new local-to-global indexer in #251 should do this.

streeve: Yep, just realized that when I went to the next one

PR #252, opened 2020-06-09
Update Interpolation

sslattery:
This PR does the following:

1. Cleans up comments in interpolation to clarify that a functor with view-like semantics may be used as the source of grid data in G2P interpolation functors. Just a `value_type` type alias and `Scalar operator()(const int i, const int j, const int k, const int d) const` is needed for accessing the data. 

2. Adds the ability to control which elements of `SplineData` are computed and/or stored. At a minimum, the stencil is always computed which requires generation of the cell size, inverse cell size, and logical position. A user can optionally store these parameters. A user can also optional compute (and store) the spline weight values, gradient values in the physical frame, and distance in the physical frame. The code is backwards compatible in that the default behavior is to compute and store all data components so existing user code won't need to be changed. The optional values are added with additional template parameters.

3. Static assertions are added to all interpolation classes to ensure the `SplineData` given to an interpolation functor has all of the requisite data. For power users that choose to control what parts of spline data get computed/stored, this will allow a clear way for them to see at compile time if they specified the right data values in their kernel required to compute the desired interpolant. This is also useful when writing your own interpolants as you can specify at compile time which data dependencies you have.

### DIFF
--- a/cajita/src/Cajita_Splines.hpp
+++ b/cajita/src/Cajita_Splines.hpp
@@ -17,6 +17,8 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <type_traits>
+
 namespace Cajita
 {
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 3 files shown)


cajita/src/Cajita_Splines.hpp
```diff
@@ -446,72 +448,321 @@ struct Spline<3>
...
+    data.dx[d] = dx;
+}
+
+template <typename Scalar, int Order, class EntityType, class DataTags>
+KOKKOS_INLINE_FUNCTION void
+setSplineData( std::false_type, SplinePhysicalCellSize,
+               SplineData<Scalar, Order, EntityType, DataTags> &, const int,
+               const Scalar )
+{
+}
```
dalg24, line 614: Why do you make the caller pass `true_type` or `false_type`?
Why don't you just use SFINAE?
```suggestion
// Assign physical cell size to the spline data.
template <typename Scalar, int Order, class EntityType, class DataTags>
KOKKOS_INLINE_FUNCTION std::enable_if_t<typename SplineData<Scalar, Order, EntityType, DataTags>::has_physical_cell_size{}>
setSplineData( SplinePhysicalCellSize,
               SplineData<Scalar, Order, EntityType, DataTags> &data,
               const int d, const Scalar dx )
{
    data.dx[d] = dx;
}

template <typename Scalar, int Order, class EntityType, class DataTags>
KOKKOS_INLINE_FUNCTION std::enable_if_t<!typename SplineData<Scalar, Order, EntityType, DataTags>::has_physical_cell_size{}>
setSplineData( SplinePhysicalCellSize,
               SplineData<Scalar, Order, EntityType, DataTags> &, const int,
               const Scalar )
{
}
```

sslattery: good idea - I implemented this

PR #253, opened 2020-06-09
Adding default constructors/copy operators to index space

sslattery:


sslattery:
> Why are you adding these?
> 
> They can all be defaulted.

It wasn't clear from looking at `Kokkos::Array` that default constructors are available - will they automatically be deep copied?

dalg24:
> > Why are you adding these?
> > They can all be defaulted.
> 
> It wasn't clear from looking at `Kokkos::Array` that default constructors are available - will they automatically be deep copied?

Yes

sslattery:
OK I have updated with defaulted versions.

dalg24:
> OK I have updated with defaulted versions.

You did not answer my initial question.  Why are you adding them in the first place?

sslattery:
> You did not answer my initial question. Why are you adding them in the first place?

Because they were not getting generated. For example, you could not include an `IndexSpace` as a member of class unless it was initialized in the constructor initializer list.

cajita/unit_test/tstIndexSpace.hpp
```diff
@@ -535,6 +535,18 @@ void comparisonTest()
     EXPECT_TRUE( is1 != is3 );
 }
 
+//---------------------------------------------------------------------------//
+void copyAssignmentTest()
+{
+    IndexSpace<3> is1( {9, 2, 1}, {12, 16, 4} );
+    auto is2 = is1;
```
dalg24, line 542: This is copy construction not copy assignment

sslattery:
Updated to only add default constructor and a test that fails if there is no default constructor.

cajita/src/Cajita_IndexSpace.hpp
```diff
@@ -32,6 +32,13 @@ class IndexSpace
     //! Number of dimensions.
     static constexpr long Rank = N;
 
+    //! Default constructor.
+    IndexSpace()
+    {
+        std::fill( _min.data(), _min.data() + Rank, -1 );
+        std::fill( _max.data(), _max.data() + Rank, -1 );
+    }
```
dalg24, line 40: You won't be able to default construct on the device.  I thought you need that.

PR #257, opened 2020-06-11
Fix incorrect assignment in spline data

sslattery:
Incorrectly used a data member that might not exist if a user doesn't specify it.


sslattery:
@dalg24 Looks like our ArborX usage got out of date with updates to the traits interface - hence failing tests when ArborX is enabled.

PR #258, opened 2020-06-11
Heffte link fix

### LINKED ISSUE #259: CMake: Heffte include directories not passed to targets
When building CabanaMD and including Cabana-Cajita's`FastFourierTransform.hpp`, which itself includes `heffte_fft3d.h`, a build error occurs.

Here is how I am building Cabana:
```cmake
cmake \
     -D CMAKE_BUILD_TYPE="Debug" \
     -D CMAKE_PREFIX_PATH="$KOKKOS_INSTALL_DIR;$HEFFTE_INSTALL_DIR" \
     -D CMAKE_INSTALL_PREFIX=$CABANA_INSTALL_DIR \
     -D Cabana_ENABLE_MPI=ON \
     -D Cabana_ENABLE_TESTING=ON \
     -D Cabana_ENABLE_EXAMPLES=ON \
     -D Cabana_ENABLE_CAJITA=ON \
     -D Cabana_ENABLE_HEFFTE=ON \
     .. ;
```
 and CabanaMD: 
```cmake
cmake -D CMAKE_PREFIX_PATH="$CABANA_INSTALL_DIR" \
    -D CabanaMD_ENABLE_Serial=OFF \
    -D CabanaMD_ENABLE_OpenMP=ON \
    -D CabanaMD_ENABLE_Cuda=OFF \
    -D CabanaMD_ENABLE_LongRange=ON \
    .. ;
```

#258 is in progress and will fix this, unless the cause is actually something in CabanaMD itself or an error in how I'm building.

Error prints: `Cajita_FastFourierTransform.hpp:20:26: fatal error: heffte_fft3d.h: No such file or directory`

sfogerty:
This involves a simple fix (case change for `HEFFTE_FOUND` to `Heffte_FOUND`), and a change I'm not sure about:

I would like to build CabanaMD without needing to specify the location of a Heffte include directory. I just want to point to Cabana, which will point to Heffte itself. Currently the CMake I've written does not accomplish this. Suggestions?

Fixes #259

### DIFF
--- a/cajita/src/CMakeLists.txt
+++ b/cajita/src/CMakeLists.txt
@@ -36,7 +36,7 @@ if(HYPRE_FOUND)
     )
 endif()
 
-if(HEFFTE_FOUND)
+if(Heffte_FOUND)
   list(APPEND HEADERS_PUBLIC
     Cajita_FastFourierTransform.hpp
     )
--- a/cajita/src/CMakeLists.txt
+++ b/cajita/src/CMakeLists.txt
@@ -55,6 +55,7 @@ endif()
 
 if(Heffte_FOUND)
   target_link_libraries(Cajita Heffte::Heffte)
+  target_include_directories(Cajita PUBLIC ${Heffte_INCLUDE_DIR})
 endif()
 
 target_include_directories(Cajita
--- a/cajita/src/CMakeLists.txt
+++ b/cajita/src/CMakeLists.txt
@@ -44,18 +44,17 @@ endif()
 
 add_library(Cajita ${SOURCES_PUBLIC})
 
-target_link_libraries(Cajita
+target_link_libraries(Cajita PUBLIC
   Kokkos::kokkos
   MPI::MPI_CXX
   )
 
 if(HYPRE_FOUND)
-  target_link_libraries(Cajita HYPRE::hypre)
+  target_link_libraries(Cajita PUBLIC HYPRE::hypre)
 endif()
 
 if(Heffte_FOUND)
-  target_link_libraries(Cajita Heffte::Heffte)
-  target_include_directories(Cajita PUBLIC ${Heffte_INCLUDE_DIR})
+  target_link_libraries(Cajita PUBLIC Heffte::Heffte)
 endif()
 
 target_include_directories(Cajita


sslattery:
If we do it right then you should just need to link Cabana and it will find Heffte automatically. We may not be installing that dependency correctly.

sslattery:
Let's make an issue for the configure errors to track this.

cajita/src/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ endif()
 
 if(Heffte_FOUND)
   target_link_libraries(Cajita Heffte::Heffte)
+  target_include_directories(Cajita PUBLIC ${Heffte_INCLUDE_DIR})
```
junghans, line 58: Huh, isn't that part of `Heffte::Heffte`? (https://bitbucket.org/icl/heffte/src/458ecdc90e1d27e1cfd4d75bbee18b2a13f28a67/CMakeLists.txt#lines-91)

cajita/src/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ endif()
 
 if(Heffte_FOUND)
   target_link_libraries(Cajita Heffte::Heffte)
```
junghans, line 57:
```suggestion
  target_link_libraries(Cajita PUBLIC Heffte::Heffte)
```

sslattery: ooooh that might be it

cajita/src/CMakeLists.txt
sfogerty: Yes, I don't think adding this did anything helpful. Now I've removed it.

cajita/src/CMakeLists.txt
sfogerty: I've added the PUBLIC keyword to this and other instances of `target_link_libraries(Cajita...` but still seem to have the same issue.

sfogerty: Okay adding the `PUBLIC` keyword and fixing a CMake error in CabanaMD looks like it fixes this. Cajita wasn't being properly linked there (just cabanacore)
So I think this now fixes #259 and is ready to merge.

sslattery:
Rebase this on `master` and it will fix the travis build

cajita/src/CMakeLists.txt
```diff
@@ -36,25 +36,25 @@ if(HYPRE_FOUND)
...
 
-target_link_libraries(Cajita
+target_link_libraries(Cajita PUBLIC
   Kokkos::kokkos
   MPI::MPI_CXX
   )
 
 if(HYPRE_FOUND)
-  target_link_libraries(Cajita HYPRE::hypre)
+  target_link_libraries(Cajita PUBLIC HYPRE::hypre)
```
junghans, line 53: Are we using hypre headers in our headers? Otherwise this should be `PRIVATE`.

sslattery: Yes - I use their headers in the linear solver implementation.

PR #263, opened 2020-06-26
WIP: Particle-grid periodic communication

### LINKED ISSUE #77: Add MPI support for Periodic Boundaries and Cartesian Decompositions

streeve:
This adds the first direct particle-grid capability in Cabana within `core`, with dependency on Cajita. ~~This is all in `particle_grid/` for now, but I'm open to any reorganization and renaming. `CabanaPG` directly relies on Core and Cajita and is enabled by default if Cajita is.~~

This wraps calls to `Distributor`, `migrate`, `Halo`, and `gather` to simplify things for any user with periodic boundaries or load balancing needs.

New options include: 
 - [x] `gridDistributor` (stolen from ExaMPM), determining new ranks destinations and wrapping through periodic boundaries
 - [x] `gridMigrate` using `gridDistributor` and calling `migrate`
 - [x] migrate tests
 - [x] `gridHalo` determining particles to be ghosted and any necessary periodic shifts
 - [x] `gridGather` using `gridHalo` and calling `gather`
 - [x] Update to `gather` for modifying the send buffer with periodic shifts
 - [x] halo tests

Closes #77


particle_grid/unit_test/tstPeriodic.hpp
```diff
@@ -0,0 +1,183 @@
...
+                                                        ranks_per_dim );
+
+    // Create the global grid.
+    std::array<bool, 3> is_periodic = {true, true, true};
+    auto global_grid = Cajita::createGlobalGrid( MPI_COMM_WORLD, global_mesh,
+                                                 is_periodic, partitioner );
+
+    // Create a  grid local_grid.
+    auto local_grid = Cajita::createLocalGrid( global_grid, 1 );
+    auto local_mesh = Cajita::createLocalMesh<TEST_DEVICE>( *local_grid );
```
sslattery, line 50: Does it seem reasonable to have to make all of these objects for cases like MD with short-range only where you don't need a mesh? I could see automating how often `migrate` is called based on setting some halo in MD. That would then eliminate the need to tune how often migrate happens (e.g. every 100 time steps or something). It would instead be based on some distance criterion which would correspond to a halo width. In that case you would want to create all of these objects then as your cell size would have some distance relative to something meaningful in the simulation (like the interaction distance).

sslattery: Note this extra grid we would use to choose communications would also be the same grid from which we would build `LinkedCellList` objects.

sslattery: And now that I think about it we need a `LinkedCellList` that uses Cajita instead of the grid we have as an implementation detail. Even more reason for all of this stuff to just be in `core`.

particle_grid/src/Cabana_Periodic.hpp
```diff
@@ -0,0 +1,223 @@
...
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef CABANA_PERIODIC_HPP
```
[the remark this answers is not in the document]
streeve: Done. Yes, I waited on the periodic halo just to get the organization done first, but it should be fairly similar.

CMakeLists.txt
```diff
@@ -168,6 +168,19 @@ if(Cabana_ENABLE_CORE)
   add_subdirectory(core)
 endif()
 
+option(Cabana_ENABLE_PARTICLEGRID "Build ParticleGrid" ${Cabana_ENABLE_CAJITA})
```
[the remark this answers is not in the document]
streeve: Done - this is much simpler now.

particle_grid/unit_test/tstPeriodic.hpp
streeve: I guess I'm not sure how to get around creating all these objects, but I have considered specializations of the mesh/grid that don't hold as much information since we basically only need the bounds of the box in MD

And since we have to rebuild the neighbor list every so often, we do the same with migration, but you're right that it could potentially be done automatically with a check for anything crossing halo boundaries. 

I have the `LinkedCellList` on my list after the `PeriodicHalo`

streeve:
I think I addressed everything. Now that I did it as you suggested, this makes much more sense than a separate particle-grid section

particle_grid/src/Cabana_Periodic.hpp
```diff
@@ -0,0 +1,223 @@
...
+    // Locate the particles in the global grid and get their destination
+    // rank. The particle halo should be constructed such that particles will
+    // only move to a location in the 26 neighbor halo or stay on this
+    // rank. If the particle crosses a periodic boundary update it's
+    // positions to represent the shift.
+    auto local_mesh = Cajita::createLocalMesh<Kokkos::HostSpace>( local_grid );
+    const auto &global_grid = local_grid.globalGrid();
+    const auto &global_mesh = global_grid.globalMesh();
+    const Kokkos::Array<double, 3> local_low = {
+        local_mesh.lowCorner( Cajita::Own(), Cajita::Dim::I ),
```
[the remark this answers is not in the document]
streeve: The migrate code now handle variable grid halo widths and is tested up to 3 times the mesh width

junghans:
@streeve that helped with the build, but now the `Cabana_ParticleGridCommunication_test_OPENMP_1` test fails.

streeve:
> @streeve that helped with the build, but now the `Cabana_ParticleGridCommunication_test_OPENMP_1` test fails.

Thanks - that's expected at the moment. `gridMigrate` passes everything, but `gridGather` has some more issues to fix.

streeve:
This will fail until rebased with 295 in master; after that, probably a few more fixes to pass halo tests and it will be ready for review.

sslattery:
#295 has been merged - let's proceed with this one and see if it resolves our comm seg faults.

sslattery:
Let's rebase this one to get the latest changes for the number of testing ranks and see what the CI does. We likely still have a UVM error though.

streeve:
@sslattery this is ready for another review. @junghans codecov is telling me I reduced coverage for `CommPlan` to zero - any ideas?

core/src/Cabana_ParticleGridCommunication.hpp
```diff
@@ -0,0 +1,790 @@
...
+                    nid[d] = 2;
+            }
+
+            // Compute the destination MPI rank.
+            destinations( p ) = neighbor_ranks(
+                nid[Cajita::Dim::I] +
+                3 * ( nid[Cajita::Dim::J] + 3 * nid[Cajita::Dim::K] ) );
+        } );
+    // TODO: fuse kernels
+    periodicWrap( local_grid, positions );
```
sslattery, line 223: Why not fuse them in this PR? They are already fused in the Picasso implementation.

core/unit_test/tstParticleGridCommunication.hpp
```diff
@@ -0,0 +1,445 @@
...
+    auto lo_z = local_mesh.lowCorner( Cajita::Own(), Cajita::Dim::K );
+    auto width_x = local_mesh.extent( Cajita::Own(), Cajita::Dim::I ) / 2.0;
+    auto width_y = local_mesh.extent( Cajita::Own(), Cajita::Dim::J ) / 2.0;
+    auto width_z = local_mesh.extent( Cajita::Own(), Cajita::Dim::K ) / 2.0;
+    auto center_x = width_x + lo_x;
+    auto center_y = width_y + lo_y;
+    auto center_z = width_z + lo_z;
+    auto shift_x = width_x - ( halo_width - 0.1 ) * dx;
+    auto shift_y = width_y - ( halo_width - 0.1 ) * dy;
+    auto shift_z = width_z - ( halo_width - 0.1 ) * dz;
```
sslattery, line 273: All of the setup code looks the same for both tests - perhaps it can be combined.

core/unit_test/tstParticleGridCommunication.hpp
```diff
@@ -0,0 +1,445 @@
...
+
+    Cabana::AoSoA<DataTypes, Kokkos::HostSpace> initial( "initial", num_data );
+    Cabana::deep_copy( initial, data_host );
+    auto pos_initial = Cabana::slice<pos_index>( initial );
+
+    // Copy to the device.
+    Cabana::AoSoA<DataTypes, TEST_MEMSPACE> data_src( "data_src", num_data );
+    Cabana::deep_copy( data_src, data_host );
+
+    if ( test_type == 0 )
```
sslattery, line 313: Did you anticipate adding other values for `test_type`?

core/unit_test/tstParticleGridCommunication.hpp
```diff
@@ -0,0 +1,445 @@
...
+        EXPECT_DOUBLE_EQ( pos_host( i, Cajita::Dim::K ),
+                          pos_initial( i, Cajita::Dim::K ) );
+    }
+    for ( int i = local_num_data; i < new_num_data; ++i )
+    {
+        bool within_x = true;
+        bool within_y = true;
+        bool within_z = true;
+        // Make sure all ghosts are in halo region in at least one direction.
+        if ( pos_host( i, Cajita::Dim::I ) < lo_x )
```
sslattery, line 352: Could you add a particle field which contains the original MPI rank and then use that to back-calculate what the position should be exactly?

sslattery (changes requested):

core/src/Cabana_ParticleGridCommunication.hpp
streeve: Forgot about that - done

core/src/Cabana_ParticleGridCommunication.hpp
```diff
@@ -0,0 +1,790 @@
...
+  \param local_grid The local grid containing periodicity and system bound
+  information.
+
+  \param positions The particle positions.
+
+  \return Distributor for later migration.
+*/
+template <class LocalGridType, class PositionSliceType>
+Distributor<typename PositionSliceType::device_type>
+gridDistributor( const LocalGridType& local_grid, PositionSliceType& positions )
```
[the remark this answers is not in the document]
streeve: Done

core/src/Cabana_ParticleGridCommunication.hpp
```diff
@@ -0,0 +1,790 @@
...
+  \param min_halo_width Number of halo mesh widths to include for ghosting.
+
+  \param max_export_guess The allocation size for halo export ranks, IDs, and
+  periodic shifts
+
+  \return Pair containing the Halo and PeriodicShift.
+*/
+template <class LocalGridType, class PositionSliceType,
+          std::size_t PositionIndex>
+auto gridHalo(
```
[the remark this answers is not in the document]
streeve: Done

core/src/Cabana_ParticleGridCommunication.hpp
```diff
@@ -0,0 +1,790 @@
...
+
+    // Create the Cabana Halo.
+    auto halo =
+        Halo<device_type>( local_grid.globalGrid().comm(), positions.size(),
+                           ids, destinations, topology );
+
+    // Create the Shifts.
+    auto periodic_shift = PeriodicShift<device_type, PositionIndex>( shifts );
+
+    return std::make_pair( halo, periodic_shift );
```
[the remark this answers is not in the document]
streeve: Done - although I'm not sure if it's composed to be extensible in the way you mentioned

core/unit_test/tstParticleGridCommunication.hpp
streeve: Combined

core/unit_test/tstParticleGridCommunication.hpp
streeve: I added a slice variant of `gridGather` for a second `test_type`

PR #267, opened 2020-07-09
WIP: Sparsegrid - Sparse Index Space

YuxingQiu:
_I'm making this pull request to keep the work, especially the design and logical part, tracked. 
I'm still testing and clean the code._

Some conventions for naming: 
- blockID = 3D block indexing, i.e., (i, j, k); 
- blockNo = 1D number indicating the block position in the array; 
- blockKey/hashKey = 1D number, computed with the given indexing method, i.e. Naive or lexicographical, and Morton

What have been done:
1. Local index inside a block
2. Index blocks with the lexicographical sequence and the Morton sequence ( with some bit operations provided )
3. Execution space for cells inside a given block

What I'll finish this week:
1. Refactor the current SparseIndexSpace, extract the block index space out. I.e., SparseIndexSpace = (BlockIndexSpace + (LocalIndexSpace inside the block))
2. Add more useful interfaces I can think of in the sparseIndexSpace. 
3. Test the current code, fix bugs.
4. Consider the allocations.

Something can be improved:
1. The Morton coding is hand-coded for 3D now. This can be improved by recursive refactor, which can support arbitrary dimensions with cleaner code.

Here are some thoughts, and current drawbacks.
1. About the indexing curve for the blocks. 
     - If we use the Unordered Map, the indexing curve tends to be useless.  The block ID would be decided by the sequence to insert the blocks, the indexing curve is only used for computing the hash key. 
     - Possible solutions: (1) Use an ordered map; (2) Assume the current capacity would all be allocated; restrict the hash key to be less than the current capacity, deal with conflicts by some offsets, record/detect the conflicts by the Unordered Hash Table.
     - Some thoughts/questions: Will the sequence of the blocks significantly influence the performance, if the blocks are sparse. We need to maintain the sequence of the blocks every time step, which could be expensive.

2. More thinking about the indexing curve:
     - If we can make more constrictions on the size of the block size ( eg the block size must be 2^n ), then we can make use of the bit computations to calculate the global(block) index and the local index in a more efficient way. OR we can still use bit computations if some memory waste is acceptable.
     - Now I separate the global and local index into two structs. We can actually combine the global and local index into one struct, and specify struct according to the indexing method. 

3. About the label of Owned Grid and Ghost Grid.
     - It could be easy to handle the owned/ghost label in the uniform grid. But things could be complicated for the sparse grid.
     - Possible solutions: (1) Set up two tables, one for the owned grid, and another for ghost grid. (2) set up two tables, one for all the grids, and another for the grid label.


sslattery:
A few comments:

- Regarding the naming convention of `Block` let's change this to `Tile` as `Block` is used elsewhere in the library to represent a partition
- I think the ordering may matter a lot and we want to do this ordering in the context of global indices rather than local indices. In that case we need to compute the has ourselves so we will need a Morton ordering. We could eventually sort tiles by Morton ordering for improved performance.
    - Question: Can you insert your own hash into the Kokkos unordered map?
- We will have another partitioning object which defines what part of the grid is `Own` and what part is `Ghost` so don't worry about that yet.
- Tiles should be a power-of-2
- Assume the global domain is indexed by power-of-2 for efficiency and the actual domain is just a sparse subset of that.

sslattery:
Also I would like all of this sparse stuff to go in the `Cajita::Experimental` namespace for now.

YuxingQiu:
Okay. These days I mainly worked on the SIGGRAPH presentation and some other stuff so didn't push that forward here. Sorry about that.
Actually, the hash we used in the multi-GPU project is relatively trivial and still not ordered. We use the Morton code as the hash key to map the block 'ijk' to an id number. The id number is computed according to the insert sequence of the blocks. So it's basically the same in the current circumstances. We didn't put much time into exploring the hash algorithms. So I don't think adding my current hash table will make a great difference.

YuxingQiu:
As long as the same hash table is shared by every MPI rank.

sslattery:
> As long as the same hash table is shared by every MPI rank.

It will not - we won't want to communicate. In that case we need to make hash keys and/or some kind of unique global identifier for each tile independent of if the hash table is shared

YuxingQiu:
Making the hash key unique is feasible as long as the hash key is computed from the global block 'ijk'. Thus, we can pass in the global block 'ijk' when inserting a block into the sparse index space.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+    {
+        Naive = 0,
+        Morton = 1,
+    };
+};
+
+//---------------------------------------------------------------------------//
+// bit operations for Morton Code computing
+template <typename Integer>
+constexpr Integer bit_length( Integer N ) noexcept
```
sslattery, line 24: Will these get computed at compile-time? If not we could do this with templates

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+        : _tilenum( tnum )
+    {
+    }
+    TileID2HashKey( int i, int j, int k )
+        : _tilenum( i, j, k )
+    {
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr auto operator()( std::array<int, 3> tileid ) -> Key
```
sslattery, line 124: lets have this just take 3 const integers for the tile id instead of an `std::array`

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr auto operator()( int tileid_x, int tileid_y, int tileid_z ) -> Key
+    {
+        return tileid_x * _tilenum[1] * _tilenum[2] + tileid_y * _tilenum[2] +
+               tileid_z;
+    }
+
+  private:
+    std::array<int, 3> _tilenum;
```
sslattery, line 138: store this as either 3 separate integers or in a `Kokkos::Array` so we can trivially copy to and from the device

cajita/src/Cajita_SparseIndexSpace.hpp
sslattery: actually now that I see the function below you can just get rid of this one.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr auto operator()( std::array<int, 3> tileid ) -> Key
+    {
+        return tileid[0] * _tilenum[1] * _tilenum[2] + tileid[1] * _tilenum[2] +
+               tileid[2];
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr auto operator()( int tileid_x, int tileid_y, int tileid_z ) -> Key
```
sslattery, line 131: use `i`, `j`, and `k` instead of `x`, `y`, and `z`

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr auto operator()( std::array<int, 3> tileid ) -> Key
+    {
+        return bit_spread( page_zmask, tileid[2] ) |
+               bit_spread( page_ymask, tileid[1] ) |
+               bit_spread( page_xmask, tileid[0] );
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr auto operator()( int tileid_x, int tileid_y, int tileid_z ) -> Key
```
sslattery, line 162: same here, just use this function signature and change xyz to ijk

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+        : _tilenum( tnum )
+    {
+    }
+    HashKey2TileID( int i, int j, int k )
+        : _tilenum( i, j, k )
+    {
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    constexpr void operator()( Key tilekey, std::array<int, 3> &tileid )
```
sslattery, line 184: also eliminate the array functions in these `HashKey2TileID` specializations and use ijk instead of xyz

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+    //! space size (global), channel size
+    Kokkos::Array<int, Rank> _min;
+    Kokkos::Array<int, Rank> _max;
+};
+
+//---------------------------------------------------------------------------//
+template <unsigned long long TBits, unsigned long long TSizePerDim,
+          unsigned long long TSize, typename Hash, typename Device,
+          typename Key, typename Value>
+class BlockIndexSpace
```
sslattery, line 318: In this case is a `Block` the set of `Tiles` owned by an MPI rank?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+        : _blkIdSpace( 1 << bit_count( capacity ), rehash_factor, size )
+        , _tileIdSpace()
+    {
+        std::fill( _min.data(), _min.data() + Rank, 0 );
+        std::copy( size.begin(), size.end(), _max.data() );
+    }
+
+    //! given a cell ijk, insert the tile such cell reside in
+    KOKKOS_FORCEINLINE_FUNCTION
+    ValueType insert_cell( int cellid_x, int cellid_y, int cellid_z )
```
sslattery, line 278: Given that it is a map I'm assuming the tile won't be inserted if the key already exists? Also you could just call `insert_tile` in this function with the tile ids you created - it would be a bit more clear that way.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    ValueType insert_tile( int tileid_x, int tileid_y, int tileid_z )
+    {
+        return _blkIdSpace.insert( tileid_x, tileid_y, tileid_z );
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    ValueType query_cell( int cellid_x, int cellid_y, int cellid_z )
```
sslattery, line 292: What does this function do?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+
+  private:
+    //! block index space, map tile ijk to tile No
+    BlockIndexSpace<TileBitsPerDim, TileSizePerDim, TSize, HashType, Device,
+                    KeyType, ValueType>
+        _blkIdSpace;
+    //! tile index space, map cell ijk to cell local No inside a tile
+    TileIndexSpace<TileBitsPerDim, TileSizePerDim, TSize> _tileIdSpace;
+    //! space size (global), channel size
+    Kokkos::Array<int, Rank> _min;
```
sslattery, line 310: Will these min/max values be used anywhere?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+    Kokkos::Array<int, Rank> _max;
+};
+
+//---------------------------------------------------------------------------//
+template <unsigned long long TBits, unsigned long long TSizePerDim,
+          unsigned long long TSize, typename Hash, typename Device,
+          typename Key, typename Value>
```
sslattery, line 317: Make the template names and order consistent across the different index space types.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+        if ( _tile_table.size() >= _tile_capacity )
+        {
+            _tile_capacity = static_cast<int>( _tile_capacity * _rehash_coeff );
+            _tile_table.rehash( _tile_capacity );
+            // TODO: view reallocate needed
+        }
+        const KeyType tileKey = _op_ijk2key( tileid_x, tileid_y, tileid_z );
+        ValueType tileNo;
+        if ( ( tileNo = _key_table.find( tileKey ) ) ==
+             Kokkos::UnorderedMap::invalid_index )
```
sslattery, line 355: This is probably the best way to do this in the short term. I am curious if a map ordered by hash will be more performant in terms of stencil-based memory access like `P2G` and `G2P`

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+            tileNo = _blk_table.size();
+            _tile_table.insert( tileKey, tileNo );
+        }
+        return tileNo;
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
+    void sortTile()
+    {
+        // pass
```
sslattery, line 368: would your objective be to sort the map here?

cajita/src/Cajita_SparseIndexSpace_Old.hpp
```diff
@@ -0,0 +1,488 @@
+#ifndef CAJITA_SPARSE_INDEXSPACE_HPP
+#define CAJITA_SPARSE_INDEXSPACE_HPP
+
```
sslattery, line 3: Can we get rid of this old file?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+  \class SparseIndexSpace
+  \brief Sparse index space, hierarchical structure (cell->tile->block)
+  \ ValueType : tileNo type
+ */
+template <int N = 3, unsigned long long TileNPerDim = 4,
```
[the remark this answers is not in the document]
YuxingQiu: I did consider this part. I'm trying to fit the design as the index space of uniform grids. Here I only considered 3D cases, with the first three for the `ijk`, the last one, if there is, for the channel. That is, only N=3 or 4 are possible. I actually don't think its a generalizable design and also thinking about changing it. Because if the channel dimension involves, I'm not sure how to identify if the spatial dimension is 2 or 3 when N = 3.

The mapping between ijk and id number is hardcoded for 3D now, I can change them to recursive implementation for support for both 2D and 3D in the next round refactor.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,499 @@
...
+          typename Hash = HashTypes::Naive,
+          typename Device = Kokkos::DefaultExecutionSpace,
+          typename Key = uint64_t, typename Value = uint64_t>
+class SparseIndexSpace
+{
+  public:
+    //! Number of dimensions, 3 = ijk, or 4 = ijk + ch
```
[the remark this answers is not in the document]
YuxingQiu: Yes I will refactor the mapping between ijk and id to a recursive implementation to support an arbitrary dimension.

cajita/src/Cajita_SparseIndexSpace_Old.hpp
YuxingQiu: Yes I will delete it in next push

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Yes, I'm trying to see if we can sort after all tiles are inserted. Since each tile is mapped to a Morton code, what we should do here is to get all the Morton code inside the hash table, and sort them, and then give them a new id.

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Yes, this block means the block owned by an MPI rank. block = a set of tiles

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Well, just wanna match the indexspace design of uniform grids. The last dimension would carry the channel information, which could be useful. Not sure how to have the channel dimension involved, so just reserved the size info here.

cajita/src/Cajita_SparseIndexSpace.hpp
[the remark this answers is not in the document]
sslattery: That is fine - we can overload for 2d later. No need to do recursive implementation. I think the bit operations will give better performance and work for what we want.

cajita/src/Cajita_SparseIndexSpace.hpp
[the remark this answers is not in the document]
sslattery: No need to refactor here - just stick with 3D for now and the current bit-operation implementation

cajita/src/Cajita_SparseIndexSpace.hpp
sslattery: OK - let's make accessors in the class for me so we can get that information.

PR #268, opened 2020-07-10
Fused Halo

sslattery:
Generalizes `Cajita::Halo` to allow for the halo operations of an arbitrary number of arrays to be fused together. The original `Halo` creation functions remain but the interface is now slightly different:

- The class has a single template `MemorySpace` instead of two (was `Scalar` and `DeviceType`)
- Construction happens from arrays so we can get the value types of the arrays for byte computations
- `gather()` and `scatter()` now require an execution space instance be passed for the buffer pack/unpack
- `scatter()` requires the user to explicitly say which `ScatterReduce` operation they would like performed

NOTE: This PR breaks backwards compatibility as part of the work towards the 1.0 release so downstream packages using the `Cajita::Halo` will likely need to be updated to reflect the interface changes.

### DIFF
--- a/cajita/src/Cajita_Halo.hpp
+++ b/cajita/src/Cajita_Halo.hpp
@@ -14,7 +14,8 @@
 
 #include <Cajita_Array.hpp>
 #include <Cajita_IndexSpace.hpp>
-#include <Cajita_MpiTraits.hpp>
+
+#include <Kokkos_Core.hpp>
 
 #include <mpi.h>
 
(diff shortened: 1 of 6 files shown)


sslattery:
One item for discussion: I ultimately see getting rid of the old `Cajita::Halo` implementation and just having this new code as it is more general. I could do a few things:

1. Leave the PR as is for now and eventually delete `Halo` and rename `MultiHalo` too halo. I don't really like this as we will need to do a bunch of renaming in the future.

2. Just delete `Halo` and rename `MultiHalo` to `Halo` in this PR. It does what we want but breaks backwards compatibility now instead of later.

3. Merge `Halo` and `MultiHalo` such that there is one halo that supports both interfaces. With this option we can only have one `Halo` and then annotate the old interface as deprecated to warn users. The downside here is that the current `Halo` and `MultiHalo` classes have different template parameters so this might not work anyway.

cajita/src/Cajita_Halo.hpp
```diff
@@ -124,52 +249,41 @@ struct Max
...
-    Halo( const ArrayLayout_t &layout, const HaloPattern &pattern,
-          const int width = -1 )
+    template <class... ArrayTypes>
+    MultiHalo( const HaloPattern &pattern, const int width,
+               ArrayTypes... arrays )
     {
-        // Duplicate the communicator so we have our own communication space.
-        MPI_Comm_dup( layout.localGrid()->globalGrid().comm(), &_comm );
+        // Get the MPI communicator. All arrays should have the same
+        // communicator.
```
dalg24, line 282: should -> must and document it

cajita/src/Cajita_Halo.hpp
```diff
@@ -124,52 +249,41 @@ struct Max
...
+      width of the block.
+      \param arrays The arrays to build the halo for. These arrays must be
+      provided in the same order
     */
-    template <class ArrayLayout_t>
-    Halo( const ArrayLayout_t &layout, const HaloPattern &pattern,
-          const int width = -1 )
+    template <class... ArrayTypes>
+    MultiHalo( const HaloPattern &pattern, const int width,
+               ArrayTypes... arrays )
```
dalg24, line 279: Did you consider passing by const reference instead of by value?

cajita/src/Cajita_Halo.hpp
```diff
@@ -427,37 +503,488 @@ class Halo
...
+        byte_sizes[array_idx] = sizeof( typename Array_t::value_type );
+    }
+
+    template <class Array_t, class... ArrayTypes>
+    void getByteSizes( std::vector<int> &byte_sizes, const int array_idx,
+                       const Array_t &, ArrayTypes... arrays )
+    {
+        byte_sizes[array_idx] = sizeof( typename Array_t::value_type );
+        getByteSizes( byte_sizes, array_idx + 1, arrays... );
+    }
```
dalg24, line 531: You know the size of `byte_sizes` at compile time so it could/should be a `std::array`.
Also consider returning it instead of taking it by reference to improve readability.

cajita/src/Cajita_Halo.hpp
sslattery: Yes - have changed this to passing by const reference.

cajita/src/Cajita_Halo.hpp
sslattery: I have changed this to `std::array`

cajita/src/Cajita_Halo.hpp
sslattery: I have documented this more and added an explicit check when I get the communicators by checking each array comm with `MPI_Comm_compare`

sslattery:
> One item for discussion: I ultimately see getting rid of the old `Cajita::Halo` implementation and just having this new code as it is more general. I could do a few things:
> 
> 1. Leave the PR as is for now and eventually delete `Halo` and rename `MultiHalo` to halo. I don't really like this as we will need to do a bunch of renaming in the future.
> 2. Just delete `Halo` and rename `MultiHalo` to `Halo` in this PR. It does what we want but breaks backwards compatibility now instead of later.
> 3. Merge `Halo` and `MultiHalo` such that there is one halo that supports both interfaces. With this option we can only have one `Halo` and then annotate the old interface as deprecated to warn users. The downside here is that the current `Halo` and `MultiHalo` classes have different template parameters so this might not work anyway.

@dalg24 @kwitaechong @rfbird @streeve I am going to push a commit which implements option 2. I can only think of a few projects using the halo at this point and most of them are mine so I can update. I think this is the best option as we push towards a 1.0 release in the near term.

cajita/src/Cajita_Halo.hpp
```diff
@@ -427,37 +509,512 @@ class Halo
...
+        static_assert(
+            std::is_same<typename Array_t::memory_space, memory_space>::value,
+            "Memory space of array is different than halo" );
+        byte_sizes[array_idx] = sizeof( typename Array_t::value_type );
+    }
+
+    template <std::size_t N, class Array_t, class... ArrayTypes>
+    void getByteSizes( std::array<std::size_t, N> &byte_sizes,
+                       const std::size_t array_idx, const Array_t &,
+                       const ArrayTypes &... arrays )
```
dalg24, line 538: I really don't get why this need to be so convoluted.

First of, I think this overload never gets called.
Second if you must use a function for something that trivial (you only use it at one single place so you might as well inline the definition of the array) you could do
```cpp
template <class... ArrayTypes>
std::array<std::size_t, sizeof...(ArrayTypes)> getByteSizes( const ArrayTypes &... arrays )
{
    return {sizeof( typename ArrayTypes::value_type )...};
}

sslattery: good idea - I eliminated this function in favor of assignment.

cajita/src/Cajita_Halo.hpp
sslattery: I also added a check that the arrays have the same halo width.

cajita/src/Cajita_Halo.hpp
sslattery: Was also able to eliminate the recursion that gets the shared index spaces.

rfbird:
> One item for discussion: I ultimately see getting rid of the old `Cajita::Halo` implementation and just having this new code as it is more general. I could do a few things:
> 
> 1. Leave the PR as is for now, deprecate `Halo` and eventually delete `Halo` and rename `MultiHalo` to halo. I don't really like this as we will need to do a bunch of renaming in the future. This avoids breaking backwards compatibility right now.
> 2. Just delete `Halo` and rename `MultiHalo` to `Halo` in this PR. It does what we want but breaks backwards compatibility now instead of later.
> 3. Merge `Halo` and `MultiHalo` such that there is one halo that supports both interfaces. With this option we can only have one `Halo` and then annotate the old interface as deprecated to warn users. The downside here is that the current `Halo` and `MultiHalo` classes have different template parameters so this might not work anyway. This won't break backwards compatibility now but we will have to update the template interface of `MultiHalo` in the future.

I vote #2, It's nice to keep backwards compat, but not many people are using it and they can stay on an old version until they're ready. I think not having the mess in the 1.0 release is priority for me here

sslattery:
OK I just pushed option 2. I left the creation functions for some backwards compatibility. The main difference is that the new halo is templated on just a `MemorySpace` instead of a `Scalar` and `DeviceType`. The `gather()` and `scatter()` functions now take execution space instances and one must always provide the `ScatterReduce` type to `scatter()` for clarity.

sslattery:
@rfbird @streeve @kwitaechong can I get a review on this one when you get a chance?

sslattery:
> 1. We could perhaps better document to the user that we're doing this nice fusion and MPI comms reduction stuff, to discourage the overly enthusiastic user from trying to manually splat things together
> 2. As a consequence of #1, we could spell out to the user if they don't want things to be fused, for some reason, that it's trivial to just make a second halo (and that they should play nice together, etc)

I can make the fusion concept a bit more clear and give examples

> Finally, a question: Is there any downside to doing all the packing in MPI_BYTES? It just makes the coding here a bit harder but no other real down side, right?

The downside is more read/write operations and longer loops. I had originally tried to pack/unpack with casting using the `value_type` of a given array. The problem is you end up casting across alignment boundaries in the buffer and so some systems (like NVIDIA) were segfaulting. Hence we need to cast to bytes such that a cast of size 1 is never across an alignment boundary. If there is a portable `memcpy` where we just copy bytes from one buffer to the next that would help I bet.

PR #275, opened 2020-08-03
Empty parameter pack

sslattery:
Places the `ParameterPack` in its own header because it has proven useful elsewhere and adds a unit test. Also fix a bug to allow for empty parameter packs to be constructed.


sslattery:
> Looks good. Just failing on formatting

good catch - format fixed

PR #276, opened 2020-08-06
Unique communication neighbor ranks

streeve:
Enforce unique neighbor ranks with `CommunicationPlan` rather than requiring it from users.

### DIFF
--- a/core/src/Cabana_CommunicationPlan.hpp
+++ b/core/src/Cabana_CommunicationPlan.hpp
@@ -428,7 +428,7 @@ class CommunicationPlan
       \param neighbor_ranks List of ranks this rank will send to and receive
       from. This list can include the calling rank. This is effectively a
       description of the topology of the point-to-point communication
-      plan. The elements in this list must be unique.
+      plan. Only the unique elements in this list are used.
 
       \return The location of each export element in the send buffer for its
       given neighbor.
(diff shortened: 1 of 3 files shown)


sslattery:
This was essentially a requirement and we always made people do it outside of the code - this is a good idea.

dalg24:
My question is why do we not want to require it any more

sslattery:
> My question is why do we not want to require it any more

I think this is useful for periodicity where repeated self-sends can possibly occur. Generally when you know communication topology you a priori know which ranks are send/receive pairs. This saves some code for the user so they don't have to filter out repeated self sends - the can just loop through neighbors and give the ranks.

streeve:
> > My question is why do we not want to require it any more
> 
> I think this is useful for periodicity where repeated self-sends can possibly occur. Generally when you know communication topology you a priori know which ranks are send/receive pairs. This saves some code for the user so they don't have to filter out repeated self sends - the can just loop through neighbors and give the ranks.

Agreed, this is partially just a small convenience for the user. But also, even though we previously required a unique list, we did not enforce it.

dalg24:
I just think it is weird.  The current version is we provide two overloads:
1. one with a larger contract that does more work, and
2. one with a narrow contract that is able to do less because the user passes more knowledge about his input

Now the changes proposed here relax the pre-condition on 2. and for the sake of convenience makes it do more work.  One extreme case would be the user passes a `std::vector` argument that essentially duplicates the entries in the `Kokkos::View` argument and you have a valid input for 2. that performs poorly compared to calling 1.

streeve:
That edge case is now possible, but the only added work is reducing the vector - we still avoid one global communication (in the other overload) and do not do any more communication than when passing a unique vector. 

So we will 1) enforce unique ranks 2) remove the requirement the user always make their neighbor ranks unique, which has been the case for every real app I've seen so far

PR #281, opened 2020-08-19
Add extent to local mesh

streeve:
Add `extent` of the locally owned mesh (with or without ghosted cells) for convenience. This matches the global mesh `extent` and will be useful for the grid communication features.

### DIFF
--- a/cajita/src/Cajita_LocalMesh.hpp
+++ b/cajita/src/Cajita_LocalMesh.hpp
@@ -120,6 +120,19 @@ class LocalMesh<Device, UniformMesh<Scalar>>
         return _ghost_high_corner[dim];
     }
 
+    // Get the extent of a given dimension.
+    KOKKOS_INLINE_FUNCTION
+    Scalar extent( Own, const int dim ) const
+    {
+        return _own_high_corner[dim] - _own_low_corner[dim];
+    }
+
+    KOKKOS_INLINE_FUNCTION
+    Scalar extent( Ghost, const int dim ) const
+    {
+        return _ghost_high_corner[dim] - _ghost_low_corner[dim];
+    }
+
     // Get the coordinates of an entity of the given type given the local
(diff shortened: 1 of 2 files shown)


cajita/src/Cajita_LocalMesh.hpp
```diff
@@ -120,6 +120,19 @@ class LocalMesh<Device, UniformMesh<Scalar>>
         return _ghost_high_corner[dim];
     }
 
+    // Get the extent of a given dimension.
+    KOKKOS_INLINE_FUNCTION
+    Scalar extent( Own, const int dim ) const
+    {
+        return _own_high_corner[dim] - _own_low_corner[dim];
```
dalg24, line 127: I would provide a templated version that calls `highCorner` and `lowCorner`

sslattery: You want a single function for both instantiations?

dalg24:
```C++
template <typename OwnOrGhost>
KOKKOS_FUNCTION Scalar extent( OwnOrGhost w, const int dim ) const
{
       return highCorner(w, dim) - lowCorner(w, dim);
}
```

sslattery: oh yeah good call

streeve:
Ok, I added your suggestion

cajita/src/Cajita_LocalMesh.hpp
```diff
@@ -120,6 +120,13 @@ class LocalMesh<Device, UniformMesh<Scalar>>
         return _ghost_high_corner[dim];
     }
 
+    // Get the extent of a given dimension.
+    template <typename OwnOrGhost>
```
sslattery, line 124: Change `OwnOrGhost` to `Decomposition` to be consistent with other classes.

PR #286, opened 2020-08-25
heFFTe v1.0.1 API update

af-ayala:
Integration of heFFTe v 1.0.1 into Cabana ECP.
The current API (fully based on C++11) shall remain for future developments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,14 +50,14 @@ endif()
 
 # standard dependency macro
 macro(Cabana_add_dependency)
-  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE" "" ${ARGN})
-  find_package( ${CABANA_DEPENDENCY_PACKAGE} QUIET )
+  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE;VERSION" "" ${ARGN})
+  find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} QUIET )
   string(TOUPPER "${CABANA_DEPENDENCY_PACKAGE}" CABANA_DEPENDENCY_OPTION )
   option(
     Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION}
     "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support" ${CABANA_DEPENDENCY_PACKAGE}_FOUND)
   if(Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION})
-    find_package( ${CABANA_DEPENDENCY_PACKAGE} REQUIRED )
+    find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} REQUIRED )
   endif()
   set(Cabana_ENABLE_${CABANA_DEPENDENCY_OPTION} ${${CABANA_DEPENDENCY_PACKAGE}_FOUND})
 endmacro()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,7 @@ if(HYPRE_FOUND)
 endif()
 
 # find heffte
-Cabana_add_dependency( PACKAGE Heffte )
+Cabana_add_dependency( PACKAGE Heffte VERSION 1.0.1 )
 
 #------------------------------------------------------------------------------#
 # Tests and Documentation
--- a/cajita/src/Cajita_FastFourierTransform.hpp
+++ b/cajita/src/Cajita_FastFourierTransform.hpp
@@ -29,6 +29,7 @@ namespace Experimental
 {
 //---------------------------------------------------------------------------//
 
+// TODO: Add HIP backend specialization when available.
 #ifdef KOKKOS_ENABLE_CUDA
 template <>
 struct HeffteBackendTraits<Kokkos::Cuda>
--- a/cajita/src/Cajita_FastFourierTransform.hpp
+++ b/cajita/src/Cajita_FastFourierTransform.hpp
@@ -128,7 +128,7 @@ class FastFourierTransform
     std::array<int, 3> global_high;
     std::array<int, 3> global_low;
 
-    FastFourierTransform( const ArrayLayout<EntityType, MeshType> &layout )
+    FastFourierTransform( const ArrayLayout<EntityType, MeshType>& layout )
     {
         // Get the local dimensions of the problem.
         auto entity_space =
@@ -139,17 +139,17 @@ class FastFourierTransform
             (int)entity_space.extent( Dim::I ) };
 
         // Get the global grid.
-        const auto &global_grid = layout.localGrid()->globalGrid();
+        const auto& global_grid = layout.localGrid()->globalGrid();
 
         // Get the low corner of the global index space on this rank.
-        global_low = {(int)global_grid.globalOffset( Dim::K ),
-                      (int)global_grid.globalOffset( Dim::J ),
-                      (int)global_grid.globalOffset( Dim::I )};
+        global_low = { (int)global_grid.globalOffset( Dim::K ),
+                       (int)global_grid.globalOffset( Dim::J ),
+                       (int)global_grid.globalOffset( Dim::I ) };
 
         // Get the high corner of the global index space on this rank.
-        global_high = {global_low[Dim::I] + local_num_entity[Dim::I] - 1,
-                       global_low[Dim::J] + local_num_entity[Dim::J] - 1,
-                       global_low[Dim::K] + local_num_entity[Dim::K] - 1};
+        global_high = { global_low[Dim::I] + local_num_entity[Dim::I] - 1,
+                        global_low[Dim::J] + local_num_entity[Dim::J] - 1,
+                        global_low[Dim::K] + local_num_entity[Dim::K] - 1 };
     }
 };
 
@@ -164,16 +164,18 @@ class HeffteFastFourierTransform
     using mesh_type = MeshType;
     using device_type = DeviceType;
     using exec_space = typename device_type::execution_space;
-    using backend_type = typename HeffteBackendTraits<exec_space, value_type>::backend_type;
-    using complex_type = typename HeffteBackendTraits<exec_space, value_type>::complex_type;
+    using backend_type =
+        typename HeffteBackendTraits<exec_space, value_type>::backend_type;
+    using complex_type =
+        typename HeffteBackendTraits<exec_space, value_type>::complex_type;
 
     /*!
       \brief Constructor
       \param layout The array layout defining the vector space of the transform.
       \param params Parameters for the 3D FFT.
     */
-    HeffteFastFourierTransform( const ArrayLayout<EntityType, MeshType> &layout,
-                                const FastFourierTransformParams &params )
+    HeffteFastFourierTransform( const ArrayLayout<EntityType, MeshType>& layout,
+                                const FastFourierTransformParams& params )
         : FastFourierTransform<Scalar, EntityType, MeshType, DeviceType>(
               layout )
     {
@@ -181,8 +183,8 @@ class HeffteFastFourierTransform
             throw std::logic_error(
                 "Only 1 complex value per entity allowed in FFT" );
 
-        heffte::box3d inbox = {this->global_low, this->global_high};
-        heffte::box3d outbox = {this->global_low, this->global_high};
+        heffte::box3d inbox = { this->global_low, this->global_high };
+        heffte::box3d outbox = { this->global_low, this->global_high };
 
         heffte::plan_options heffte_params =
             heffte::default_options<backend_type>();
@@ -204,7 +206,7 @@ class HeffteFastFourierTransform
             throw std::logic_error( "Expected FFT allocation size smaller "
                                     "than local grid size" );
 
-        _fft_work = Kokkos::View<complex_type *, DeviceType>(
+        _fft_work = Kokkos::View<complex_type*, DeviceType>(
             Kokkos::ViewAllocateWithoutInitializing( "fft_work" ), fftsize );
     }
 
@@ -213,17 +215,17 @@ class HeffteFastFourierTransform
       \param in The array on which to perform the forward transform.
     */
     template <class Array_t>
-    void forward( const Array_t &x, const FFTScaleNone )
+    void forward( const Array_t& x, const FFTScaleNone )
     {
         compute( x, 1, heffte::scale::none );
     }
     template <class Array_t>
-    void forward( const Array_t &x, const FFTScaleFull )
+    void forward( const Array_t& x, const FFTScaleFull )
     {
         compute( x, 1, heffte::scale::full );
     }
     template <class Array_t>
-    void forward( const Array_t &x, const FFTScaleSymmetric )
+    void forward( const Array_t& x, const FFTScaleSymmetric )
     {
         compute( x, 1, heffte::scale::symmetric );
     }
@@ -233,17 +235,17 @@ class HeffteFastFourierTransform
      \param out The array on which to perform the reverse transform.
     */
     template <class Array_t>
-    void reverse( const Array_t &x, const FFTScaleNone )
+    void reverse( const Array_t& x, const FFTScaleNone )
     {
         compute( x, -1, heffte::scale::none );
     }
     template <class Array_t>
-    void reverse( const Array_t &x, const FFTScaleFull )
+    void reverse( const Array_t& x, const FFTScaleFull )
     {
         compute( x, -1, heffte::scale::full );
     }
     template <class Array_t>
-    void reverse( const Array_t &x, const FFTScaleSymmetric )
+    void reverse( const Array_t& x, const FFTScaleSymmetric )
     {
         compute( x, -1, heffte::scale::symmetric );
     }
@@ -252,7 +254,7 @@ class HeffteFastFourierTransform
     KOKKOS_INLINE_FUNCTION ComplexType copyFromKokkosComplex(
         Kokkos::complex<value_type> x_view_val, ComplexType work_view_val,
         typename std::enable_if<( is_cuda_complex<ComplexType>::value ),
-                                int>::type * = 0 )
+                                int>::type* = 0 )
     {
         work_view_val.x = x_view_val.real();
         work_view_val.y = x_view_val.imag();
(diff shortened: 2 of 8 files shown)


streeve (dismissed):
Thanks for putting this together. Looks like travis is failing because of formatting (we enforce clang-format 7)

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -52,101 +52,102 @@ struct HeffteMemoryTraits<Kokkos::HostSpace>
...
+//     // Buffer packing type.
+//     int packflag;
+
+//     // Forward scaling option.
+//     int scaled;
+// };
 
 //---------------------------------------------------------------------------//
-template <class Scalar, class EntityType, class MeshType, class DeviceType>
+template <class Backend, class Scalar, class EntityType, class MeshType, class DeviceType>
```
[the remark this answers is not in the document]
sfogerty: I think so - seems wise. Should we try to incorporate that restriction now, or kick that can down the road a bit?

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -163,25 +164,26 @@ class FastFourierTransform
...
+        // : _fft( layout.localGrid()->globalGrid().comm() )
     {
         if ( 1 != layout.dofsPerEntity() )
             throw std::logic_error(
                 "Only 1 complex value per entity allowed in FFT" );
 
         // Set the memory type. For now we will just do FFTs on the host until
         // we find the HEFFTE GPU memory bug.
-        _fft.mem_type = HEFFTE_MEM_CPU;
+        // _fft.mem_type = HEFFTE_MEM_CPU;  // * Memory automatically defined from backend type
```
sfogerty, line 177: This implies that the `HeffteMemoryTraits` (lines 31-52) don’t seem to be needed anymore. So I think we should go ahead and remove them.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -214,21 +216,35 @@ class FastFourierTransform
...
         if ( fftsize < (int)entity_space.size() )
-            throw std::logic_error( "HEFFTE expected allocation size smaller "
+            throw std::logic_error( "heFFTe expected allocation size smaller "
                                     "than local grid size" );
 
         // Allocate the work array.
         _fft_work = Kokkos::View<Scalar *, DeviceType>(
             Kokkos::ViewAllocateWithoutInitializing( "fft_work" ),
-            2 * fftsize );
+            2 * fftsize );
```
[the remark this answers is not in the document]
sfogerty: I'm somewhat confused here...  What I understand is you could use just allocate `fftsize` if defining input and output data to be complex? This would be for a complex-to-complex, in-place FFT. 

Could we do something like:
 ```
 _fft_work = Kokkos::View<Kokkos::complex<Scalar> *, DeviceType>(
            Kokkos::ViewAllocateWithoutInitializing( "fft_work" ),
            fftsize );

```

**Comment by streeve** [comment]
Sorry, @junghans I broke it again

**Comment by sfogerty** [comment]
@streeve @junghans My reasoning to build with `Cabana_ENABLE_HEFFTE` and not `Cabana_REQUIRE_HEFFTE` is in the details of `Cabana_add_dependency()`. 
If we set the `REQUIRE` var to ON, then we are calling `find_package(Heffte)` twice. Unfortunately Heffte v1.0 is not guarded against this, and a CMake error occurs. (My Heffte [PR](https://bitbucket.org/icl/heffte/pull-requests/83/guard-for-multiple-targets-of-same-lib/diff) just a couple commits after v1.0)

**Comment by streeve** [comment]
`Cabana_ENABLE_HEFFTE` doesn't do anything, but I understand why you removed `REQUIRE` now. We'll need a minor release for that then

**Comment by sfogerty** [comment]
> `Cabana_ENABLE_HEFFTE` doesn't do anything, but I understand why you removed `REQUIRE` now. We'll need a minor release for that then

Ah, right. I'm not sure there is an ideal solution at the moment. We could, for now, have travis just clone and build with the most recent heFFTe commit (not glued to v1.0 tag). Our build will start breaking when a new heFFTe version releases, but at that point we can redirect travis to build that new tagged version (v1.0.1, or whatever)?

File: CMakeLists.txt [hunk 4]
```diff
@@ -50,14 +50,15 @@ endif()
...
+  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE;VERSION" "" ${ARGN})
   string(TOUPPER "${CABANA_DEPENDENCY_PACKAGE}" CABANA_DEPENDENCY_OPTION )
   option(
     Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION}
     "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support" ${CABANA_DEPENDENCY_PACKAGE}_FOUND)
   if(Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION})
-    find_package( ${CABANA_DEPENDENCY_PACKAGE} REQUIRED )
+    find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} REQUIRED )
+  else()
+    find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} QUIET )
```
junghans, line 61: You cannot rearrange this as the result (`*_FOUND`) is used in the `option()` call.

junghans (dismissed):

CMakeLists.txt
junghans: I would say put it back for now, and just call `find_package(Heffte 1.0.0)` below.

sfogerty: Reverted to original, and using the new heFFTe v1.0.1 now 👍

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -28,232 +28,180 @@ namespace Cajita
...
+            Kokkos::ViewAllocateWithoutInitializing( "fft_work" ), fftsize );
     }
 
     /*!
       \brief Do a forward FFT.
       \param in The array on which to perform the forward transform.
     */
     template <class Array_t>
-    void forward( const Array_t &x )
+    void forward( const Array_t &x, const FFTScaleNone )
```
sslattery, line 168: These should be an `override` implementation of a pure virtual function in the base class. That way future FFT implementations have to conform to the same interface.

sslattery (changes requested):

cajita/src/Cajita_FastFourierTransform.hpp
streeve: That was my plan, but how would I do that with the template?

streeve:
This should be ready for another review

cajita/src/Cajita_FastFourierTransform.hpp
sslattery: explicit overloads with each template variant

sslattery:
You guys will need to do a rebase to resolve conflicts - they are likely from recent updates to the code formatting style.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -28,232 +28,281 @@ namespace Cajita
...
+        compute( x, -1, heffte::scale::symmetric );
+    }
+
+    template <class ComplexType>
+    KOKKOS_INLINE_FUNCTION ComplexType copyFromKokkosComplex(
+        Kokkos::complex<value_type> x_view_val, ComplexType work_view_val,
+        typename std::enable_if<( is_cuda_complex<ComplexType>::value ),
+                                int>::type * = 0 )
+    {
+        auto realpart = x_view_val.real();
```
sslattery, line 260: why make the temporaries here?

sslattery (changes requested):
Other than my few minor comments this looks great!

sfogerty:
Retest this please

cajita/src/Cajita_FastFourierTransform.hpp
sfogerty: No need really - I removed them

cajita/unit_test/tstFastFourierTransform.hpp
```diff
@@ -31,7 +31,8 @@ using namespace Cajita;
 namespace Test
 {
 //---------------------------------------------------------------------------//
-void memoryTest()
+// New heFFTe version - this should be re-done
+/*void memoryTest()
```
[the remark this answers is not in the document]
sfogerty: Removed this test

streeve:
I rebased to clean up the commit history. 

Also just a note that the interpolation test failed again (2 ranks)

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -28,232 +28,281 @@ namespace Cajita
...
       \param in The array on which to perform the forward transform.
     */
     template <class Array_t>
-    void forward( const Array_t &x )
+    void forward( const Array_t &x, const FFTScaleNone )
+    {
+        compute( x, 1, heffte::scale::none );
+    }
+    template <class Array_t>
+    void forward( const Array_t &x, const FFTScaleFull )
```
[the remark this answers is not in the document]
streeve: Done

cajita/src/Cajita_FastFourierTransform.hpp
streeve: At your suggestion I used CRTP for this. It's a much better design, but I couldn't see how to simplify the template parameters of the derived class like you mentioned.

streeve:
@junghans any idea why codecov is messed up?

junghans:
Try to merge with master first.

cajita/unit_test/tstFastFourierTransform.hpp
```diff
@@ -97,19 +86,53 @@ void forwardReverseTest()
...
+    }
+    else if ( use_default )
+    {
+        auto fft =
+            Experimental::createHeffteFastFourierTransform<double, TEST_DEVICE>(
+                *vector_layout );
+        fft->forward( *lhs, Experimental::FFTScaleFull() );
+        fft->reverse( *lhs, Experimental::FFTScaleNone() );
+    }
+#ifndef TEST_CUDA
```
sslattery, line 120: Why is this here?

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -23,291 +23,509 @@
 #include <memory>
 #include <type_traits>
 
+#ifdef KOKKOS_ENABLE_HIP // FIXME_HIP
+#error FFT not supported with Kokkos::Experimental::HIP backend
```
sslattery, line 27: This error should be a `static_assert` in the HIP specialization instead of a macro-level error here. That way it would only fail when people actually tried to use it. The way this is written here it will possibly preclude users from having HIP kernels and an FFT on the host.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -23,291 +23,509 @@
...
+*/
+struct FFTScaleSymmetric
+{
+};
+
+/*!
+  \class FFTBackendFFTW
+  \brief Tag specifying FFTW backend for FFT (host default).
+*/
+struct FFTBackendFFTW
```
sslattery, line 62: These backend options are heFFTe specific. All backend stuff should be put in some kind of heFFTe namespace or in the heFFTe class.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -23,291 +23,509 @@
...
-      0: point-to-point
-      1: all-to-all
-      2: combination
+//---------------------------------------------------------------------------//
+/*!
+  \class FastFourierTransform
+  \brief 3D distributed fast Fourier transform base implementation.
+*/
+template <class EntityType, class MeshType, class Scalar, class DeviceType,
+          class InterfaceType>
```
sslattery, line 167: Rename `InterfaceType` to `Derived`. It will make it very obvious that you implemented the CRTP.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -23,291 +23,509 @@
...
+      \param x The array on which to perform the forward transform.
+      \param scaling Method of scaling data.
     */
-    FastFourierTransformParams& setPackType( const int type )
+    template <class Array_t, class ScaleType>
+    void forward( const Array_t& x, const ScaleType scaling )
     {
-        packflag = type;
-        return *this;
+        checkArray( x );
```
sslattery, line 248: A better way to do this would be to write some kind of meta type checker (i.e a concept) instead of having a function which only does compile-time work. You could simply make all those `is_*` checks in the `static_assert` calls another type checker that would inherit from `std::is_true` only if they are all satisfied.

cajita/src/Cajita_FastFourierTransform.hpp
sslattery: So I would put them in the `Impl` namespace below and prefix all the struct names with `Heffte`.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -23,291 +23,509 @@
...
+#ifdef Heffte_ENABLE_CUDA
+#ifdef KOKKOS_ENABLE_CUDA
+template <>
+struct HeffteBackendTraits<Kokkos::Cuda, double, Impl::FFTBackendDefault>
+{
+    using backend_type = heffte::backend::cufft;
+    using complex_type = cufftDoubleComplex;
+};
+template <>
+struct HeffteBackendTraits<Kokkos::Cuda, float, Impl::FFTBackendDefault>
```
sslattery, line 306: Here is where I would write a HIP specialization and just put `static_assert( false, "HIP not supported" );`

cajita/src/Cajita_FastFourierTransform.hpp
sslattery: If we find that other FFT libraries do the same thing we can relax this and generalize the backend concept to some kind of 1D FFT concept.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -23,291 +23,509 @@
...
-        if ( 1 != x.layout()->dofsPerEntity() )
-            throw std::logic_error(
-                "Only 1 complex value per entity allowed in FFT" );
+    /*!
+     \brief Copy data from Kokkos::complex to std::complex.
+     \param x_view_val Kokkos::complex value.
+     \param work_view_val std::complex value.
+    */
+    template <class ComplexType>
+    ComplexType copyFromKokkosComplex(
```
sslattery, line 459: These need to be `KOKKOS_INLINE_FUNCTION`. Even if `std::complex` won't run on device we want it to be `inline`.

sslattery (changes requested):

cajita/src/Cajita_FastFourierTransform.hpp
streeve: Done

cajita/src/Cajita_FastFourierTransform.hpp
streeve: I left them for now because the same tags could be used for the other FFT libraries. The only issue there is disabling backends that don't exist for a given library (fftMPI has one more option than heFFTe)

cajita/src/Cajita_FastFourierTransform.hpp
streeve: Done

cajita/src/Cajita_FastFourierTransform.hpp
streeve: Done

cajita/src/Cajita_FastFourierTransform.hpp
streeve: I used `inline` since it can't be called from the device

cajita/unit_test/tstFastFourierTransform.hpp
streeve: This section only tests the explicit use of an FFT backend, rather than the default for the Kokkos backend. Since there are only options for the host, I needed to disable this from the automatic test expansion on all Kokkos backends

streeve:
Retest this please

streeve:
This is ready for review (HIP failure is unrelated - not yet using HIP with FFT)

cajita/unit_test/tstFastFourierTransform.hpp
```diff
@@ -30,21 +30,10 @@ using namespace Cajita;
...
-    int nbytes = size * sizeof( double );
-    double* ptr = (double*)fft_mem.smalloc( nbytes, mtype );
-    EXPECT_NE( ptr, nullptr );
-    fft_mem.sfree( ptr, mtype );
-}
 
 //---------------------------------------------------------------------------//
-void forwardReverseTest()
+template <class HostBackendType>
+void forwardReverseTest( bool use_default, bool use_params )
```
sfogerty, line 36: It strikes me as odd to have both `use_default` and `use_params` here - maybe I am missing something.
The way these are used, I might call them `use_host` and `use_params`, or something of that nature.

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -28,286 +28,548 @@ namespace Cajita
 namespace Experimental
```
sfogerty, line 28: I think we are nearing the point where these features could live outside the `Experimental` namespace. Should we take that step in this PR, or hold off?

sslattery:
retest this please

cajita/unit_test/tstFastFourierTransform.hpp
streeve: Fair point - what I meant was whether to use default backend and default parameters

core/unit_test/TestOPENMP_Category.hpp
```diff
@@ -12,6 +12,7 @@
 #ifndef CABANA_TEST_OPENMP_CATEGORY_HPP
 #define CABANA_TEST_OPENMP_CATEGORY_HPP
 
+#define TEST_HOST
```
sslattery, line 15: This is not the place to put `TEST_HOST` - add something to the unit test that checks the backends and triggers the right type. These files are specific to a more general portable unit test strategy.

sslattery (changes requested):

core/unit_test/TestOPENMP_Category.hpp
streeve: I thought this was cleaner (thinking maybe this would come up in other places), but I can do this in the FFT unit test

streeve:
Retest this please

cajita/unit_test/tstFastFourierTransform.hpp
streeve: Replaced with a check for any enabled device backend

cajita/src/Cajita_FastFourierTransform.hpp
streeve: As we discussed, we'll hold off on this for now

core/unit_test/TestOPENMP_Category.hpp
streeve: Removed

streeve:
This is ready for (hopefully final) review

sslattery (approved):

PR #287, opened 2020-08-25
sparse grid

YuxingQiu:
_I'm making this pull request to keep the work, especially the design and logical part, tracked. 
I'm still testing and clean the code._

Some conventions for naming: 
- tileID = 3D tile indexing, i.e., (i, j, k); 
- tileNo = 1D number indicating the tile position in the array; 
- tileKey/hashKey = 1D number, computed with the given indexing method, i.e. Naive or lexicographical, and Morton

What have been done:
1. Local index inside a block
2. Index blocks with the lexicographical sequence and the Morton sequence ( with some bit operations provided )
3. Hierarchical design of the current SparseIndexSpace, extract the tile index space out. I.e., SparseIndexSpace = (BlockIndexSpace + TileIndexSpace), where BlockIndexSpace provide tile indices in a single block(MPI rank), TileIndexSpace indices inside a single tile.

Further plans:
1. Test the current code, fix bugs.
2. Consider the allocations.

Something can be improved:
1. The Morton coding is hand-coded for 3D now. This can be improved by recursive refactor, which can support arbitrary dimensions with cleaner code.

Here are some thoughts, and current drawbacks.
1. About the indexing curve for the tiles. 
     - If we use the Unordered Map, the indexing curve tends to be useless.  The tile ID would be decided by the sequence to insert the tiles, the indexing curve is only used for computing the hash key. 
     - Possible solutions: (1) Use an ordered map; (2) Assume the current capacity would all be allocated; restrict the hash key to be less than the current capacity, deal with conflicts by some offsets, record/detect the conflicts by the Unordered Hash Table.
     - Some thoughts/questions: Will the sequence of the tiles significantly influence the performance, if the tiles are sparse. We need to maintain the sequence of the blocks every time step, which could be expensive.

2. More thinking about the indexing curve:
     - If we can make more constrictions on the size of the tile size ( eg the tile size must be 2^n ), then we can make use of the bit computations to calculate the global(tile) index and the local index in a more efficient way. OR we can still use bit computations if some memory waste is acceptable.
     - Now I separate the global and local index into two structs. We can actually combine the global and local index into one struct, and specify struct according to the indexing method. 

3. About the label of Owned Grid and Ghost Grid.
     - It could be easy to handle the owned/ghost label in the uniform grid. But things could be complicated for the sparse grid.
     - Possible solutions: (1) Set up two tables, one for the owned grid, and another for ghost grid. (2) set up two tables, one for all the grids, and another for the grid label.

### DIFF
--- a/cajita/src/Cajita_SparseIndexSpace.hpp
+++ b/cajita/src/Cajita_SparseIndexSpace.hpp
@@ -1,0 +1,878 @@
+#ifndef CAJITA_SPARSE_INDEXSPACE_HPP
+#define CAJITA_SPARSE_INDEXSPACE_HPP
+
+#include <Kokkos_Core.hpp>
+#include <Kokkos_UnorderedMap.hpp>
+
+#include <array>
+#include <string>
+
+//---------------------------------------------------------------------------//
+// Naming convension:
+//  Tile ID / Tile ijk = 3D tile indexing, i.e., (i, j, k)
+//  Tile No. = 1D number indicating the tile position in the allocated array
+//  Tile key / Hash key = 1D number, computed with the given indexing method and
+//  used as hash key
+//---------------------------------------------------------------------------//
+
+namespace Cajita
+{
+
+//---------------------------------------------------------------------------//
+// Hash table type tag.
+//---------------------------------------------------------------------------//
+enum class HashTypes : unsigned char
+{
+    Naive = 0, // Lexicographical Order
+    Morton = 1 // Morton Curve
+};
+
+//---------------------------------------------------------------------------//
+// bit operations for Morton Code computing
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief (Host/Device) Compute the least bit number/length needed to represent
+  the given input integer
+*/
+template <typename Integer>
+KOKKOS_INLINE_FUNCTION constexpr Integer bitLength( Integer input_int ) noexcept
+{
+    if ( input_int )
+        return bitLength( input_int >> 1 ) + static_cast<Integer>( 1 );
+    else
+        return 0;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief (Host/Device) Compute the lease bit number needed to index input
+  integer
+*/
+template <typename Integer>
+KOKKOS_INLINE_FUNCTION constexpr Integer bitCount( Integer input_int ) noexcept
+{
+    return bitLength( input_int - 1 );
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief (Host/Device) Given a integer, reverse the corresponding binary string,
+  return the resulting integer.
+*/
+template <typename Integer>
+KOKKOS_INLINE_FUNCTION constexpr Integer
+binaryReverse( Integer input_int, char loc = sizeof( Integer ) * 8 - 1 )
+{
+    if ( input_int == 0 )
+        return 0;
+    return ( ( input_int & 1 ) << loc ) |
+           binaryReverse( input_int >> 1, loc - 1 );
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief (Host/Device) Count the leading zeros in the corresponding binary
+  string of the input integer
+*/
+template <typename Integer>
+KOKKOS_INLINE_FUNCTION constexpr unsigned countLeadingZeros( Integer input_int )
+{
+    unsigned res{ 0 };
+    input_int = binaryReverse( input_int );
+    if ( input_int == 0 )
+        return sizeof( Integer ) * 8;
+    while ( ( input_int & 1 ) == 0 )
+        res++, input_int >>= 1;
+    return res;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief (Host/Device) Pack up the data bits where the corresponding bit of the
+  mask is 1
+*/
+KOKKOS_INLINE_FUNCTION
+constexpr int bitPack( const uint64_t mask, const uint64_t data )
+{
+    uint64_t slresult = 0;
+    uint64_t& ulresult{ slresult };
+    uint64_t uldata = data;
+    int count = 0;
+    ulresult = 0;
+
+    uint64_t rmask = binaryReverse( mask );
+    unsigned char lz{ 0 };
+
+    while ( rmask )
+    {
+        lz = countLeadingZeros( rmask );
+        uldata >>= lz;
+        ulresult <<= 1;
+        count++;
+        ulresult |= ( uldata & 1 );
+        uldata >>= 1;
+        rmask <<= lz + 1;
+    }
+    ulresult <<= 64 - count; // 64 bits (maybe not use a constant 64 ...?)
+    ulresult = binaryReverse( ulresult );
+    return (int)slresult;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief (Host/Device) Spread out the data bits where the corresponding bit of
+  the mask is 1
+*/
+KOKKOS_INLINE_FUNCTION
+constexpr uint64_t bitSpread( const uint64_t mask, const int data )
+{
+    uint64_t rmask = binaryReverse( mask );
+    int dat = data;
+    uint64_t result = 0;
+    unsigned char lz{ 0 };
+    while ( rmask )
+    {
+        lz = countLeadingZeros( rmask ) + 1;
+        result = result << lz | ( dat & 1 );
+        dat >>= 1, rmask <<= lz;
+    }
+    result = binaryReverse( result ) >> countLeadingZeros( mask );
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+// Tile Hash Key (hash) <=> TileID.
+//---------------------------------------------------------------------------//
+template <typename Key, HashTypes HashT>
+struct TileID2HashKey;
+
+template <typename Key, HashTypes HashT>
+struct HashKey2TileID;
+
+/*!
+  \struct TileID2HashKey
+  \brief Compute the hash key from the 3D tile ijk
+
+  Lexicographical order specialization
+  Can be rewrite in a recursive way
+*/
+template <typename Key>
+struct TileID2HashKey<Key, HashTypes::Naive>
+{
+    using tid_to_key_type = TileID2HashKey<Key, HashTypes::Naive>;
+    static constexpr HashTypes hash_type = HashTypes::Naive;
+
+    //! Constructor (Host) from a given initializer list
+    TileID2HashKey( std::initializer_list<int> tnum )
+    {
+        std::copy( tnum.begin(), tnum.end(), _tile_num.data() );
+        _tile_num_j_mul_k = _tile_num[1] * _tile_num[2];
+    }
+    //! Constructor (Host) from three integers
+    TileID2HashKey( int i, int j, int k )
+    {
+        _tile_num[0] = i;
+        _tile_num[1] = j;
+        _tile_num[2] = k;
+        _tile_num_j_mul_k = j * k;
+    }
+
+    /*!
+    \brief (Device) Compute the 1D hash key of tile in a lexicographical way
+    */
+    KOKKOS_INLINE_FUNCTION
+    Key operator()( int tile_i, int tile_j, int tile_k ) const
+    {
+        return tile_i * _tile_num_j_mul_k + tile_j * _tile_num[2] + tile_k;
+    }
+
+  private:
+    // The index bounds of the tiles on the current MPI rank
+    Kokkos::Array<int, 3> _tile_num;
+    int _tile_num_j_mul_k;
+};
+
+/*!
+  \struct TileID2HashKey
+  \brief Compute the hash key from the 3D tile ijk
+
+  Morton code specialization
+*/
+template <typename Key>
+struct TileID2HashKey<Key, HashTypes::Morton>
+{
+    using tid_to_key_type = TileID2HashKey<Key, HashTypes::Morton>;
+    static constexpr HashTypes hash_type = HashTypes::Morton;
+
+    //! Constructor (Host) from a given initializer list
+    TileID2HashKey( std::initializer_list<int> tnum )
+    {
+        std::copy( tnum.begin(), tnum.end(), _tile_num.data() );
+    }
+    //! Constructor (Host) from three integers
+    TileID2HashKey( int i, int j, int k )
+    {
+
+        _tile_num[0] = i;
+        _tile_num[1] = j;
+        _tile_num[2] = k;
+    }
+    //! Page mask used for computing the morton code
+    enum : uint64_t
+    { // hand-coded now, can be improved by iterative func
+        page_kmask = ( 0x9249249249249249UL ),
+        page_jmask = ( 0x2492492492492492UL ),
+        page_imask = ( 0x4924924924924924UL )
+    };
+
+    /*!
+    \brief (Device) Compute the 1D hash key of tile in a morton way
+    */
+    KOKKOS_INLINE_FUNCTION
+    Key operator()( int tile_i, int tile_j, int tile_k ) const
+    {
+        return bitSpread( page_kmask, tile_k ) |
+               bitSpread( page_jmask, tile_j ) |
+               bitSpread( page_imask, tile_i );
+    }
+
+  private:
+    // The index bounds of the tiles on the current MPI rank
+    Kokkos::Array<int, 3> _tile_num;
+};
+
+/*!
+  \struct HashKey2TileID
+  \brief Compute the 3D tile ijk from the hash key
+
+  Lexicographical order specialization
+  Can be rewrite in a recursive way
+*/
+template <typename Key>
+struct HashKey2TileID<Key, HashTypes::Naive>
+{
+    using key_to_tid_type = HashKey2TileID<Key, HashTypes::Naive>;
+    static constexpr HashTypes hash_type = HashTypes::Naive;
+
+    //! Constructor (Host) from a given initializer list
+    HashKey2TileID( std::initializer_list<int> tnum )
+        : _tile_num()
+    {
+        std::copy( tnum.begin(), tnum.end(), _tile_num.data() );
+    }
+    //! Constructor (Host) from three integers
+    HashKey2TileID( int i, int j, int k )
+    {
+        _tile_num[0] = i;
+        _tile_num[1] = j;
+        _tile_num[2] = k;
+    }
+
+    /*!
+    \brief (Device) Compute the tile ijk from the lexicographical tile hash key
+    */
+    KOKKOS_INLINE_FUNCTION
+    void operator()( Key tile_key, int& tile_i, int& tile_j, int& tile_k ) const
+    {
+        tile_k = tile_key % _tile_num[2];
+        tile_j = static_cast<Key>( tile_key / _tile_num[2] ) % _tile_num[1];
+        tile_i = static_cast<Key>( tile_key / _tile_num[2] / _tile_num[1] ) %
+                 _tile_num[0];
+    }
+
+  private:
+    // The index bounds of the tiles on the current MPI rank
+    Kokkos::Array<int, 3> _tile_num;
+};
+
+/*!
+  \struct HashKey2TileID
+  \brief Compute the 3D tile ijk from the hash key
+
+  Morton code specialization
+*/
+template <typename Key>
+struct HashKey2TileID<Key, HashTypes::Morton>
+{
+    using key_to_tid_type = HashKey2TileID<Key, HashTypes::Morton>;
+    static constexpr HashTypes hash_type = HashTypes::Morton;
+
+    //! Constructor (Host) from a given initializer list
+    HashKey2TileID( std::initializer_list<int> tnum )
+    {
+        std::copy( tnum.begin(), tnum.end(), _tile_num.data() );
+    }
+    //! Constructor (Host) from three integers
+    HashKey2TileID( int i, int j, int k )
+    {
+        _tile_num[0] = i;
+        _tile_num[1] = j;
+        _tile_num[2] = k;
+    }
+    //! Page mask used for computing the morton code
+    enum : uint64_t
+    { // hand-coded now, can be improved by iterative func
+        page_kmask = ( 0x9249249249249249UL ),
+        page_jmask = ( 0x2492492492492492UL ),
+        page_imask = ( 0x4924924924924924UL )
+    };
+
+    /*!
+    \brief (Device) Compute the tile ijk from the lexicographical tile hash key
+    */
+    KOKKOS_INLINE_FUNCTION
+    void operator()( Key tile_key, int& tile_i, int& tile_j, int& tile_k ) const
+    {
+        tile_k = bitPack( page_kmask, tile_key );
+        tile_j = bitPack( page_jmask, tile_key );
+        tile_i = bitPack( page_imask, tile_key );
+    }
+
+  private:
+    // The index bounds of the tiles on the current MPI rank
+    Kokkos::Array<int, 3> _tile_num;
+};
+
+//---------------------------------------------------------------------------//
+// Hierarchical index spaces
+// SparseIndexSpace <- BlockIndexSpace <- TileIndexSpace
+// Naming:
+//      Block = MPI Rank
+//      Tile = Sub-block with several cells
+//      Cell = Basic grid unit
+//---------------------------------------------------------------------------//
+
+//! Declaration of BlockIndexSpace
+template <typename MemorySpace, unsigned long long CBits,
+          unsigned long long CNumPerDim, unsigned long long CNumPerTile,
+          HashTypes Hash, typename Key, typename Value>
+class BlockIndexSpace;
+
+//! Declaration of TileIndexSpace
+template <int CBits, int CNumPerDim, int CNumPerTile>
+class TileIndexSpace;
+
+/*!
+  \class SparseIndexSpace
+  \brief Sparse index space, with a hierarchical structure (cell->tile->block)
+  \tparam MemorySpace Memory space to store the IndexSpace(Hash Table)
+  \tparam CellPerTileDim Cell number inside each tile per dimension
+  \tparam Hash Hash type (lexicographical or morton)
+  \tparam Key Type of the tile/cell hash key
+  \tparam Value Type of the tile/cell No.
+ */
+template <typename MemorySpace, unsigned long long CellPerTileDim = 4,
+          HashTypes Hash = HashTypes::Naive, typename Key = uint64_t,
+          typename Value = uint64_t>
+class SparseIndexSpace
+{
+  public:
+    //! Number of dimensions, 3 = ijk
+    static constexpr int rank = 3;
+    //! Number of bits (per dimension) needed to index the cells inside a tile
+    static constexpr unsigned long long cell_bits_per_tile_dim =
+        bitCount( CellPerTileDim );
+    //! Number of cells inside each tile (per dimension), tile size reset to
+    //! power of 2
+    static constexpr unsigned long long cell_num_per_tile_dim =
+        1 << cell_bits_per_tile_dim;
+    //! Cell mask (per dimension), indicating which part of the tile
+    //! address(binary bits) will index the cells inside each tile
+    static constexpr unsigned long long cell_mask_per_tile_dim =
+        ( 1 << cell_bits_per_tile_dim ) - 1;
+    //! Number of bits (total) needed to index the cells inside a tile
+    static constexpr unsigned long long cell_bits_per_tile =
+        cell_bits_per_tile_dim + cell_bits_per_tile_dim +
+        cell_bits_per_tile_dim;
+    //! Number of cells (total) inside each tile
+    static constexpr unsigned long long cell_num_per_tile =
+        cell_num_per_tile_dim * cell_num_per_tile_dim * cell_num_per_tile_dim;
+    //! Types
+    using key_type = Key;                        // tile hash key
+    using value_type = Value;                    // tile No.
+    static constexpr HashTypes hash_type = Hash; // hash table
+
+    /*!
+      \brief (Host) Constructor
+      \param size The size of the block (MPI rank) (Unit: cell)
+      \param capacity Expected capacity of the allocator to store the tiles
+      when tile nums exceed the capacity
+    */
+    SparseIndexSpace( const std::array<int, rank> size,
+                      const unsigned int capacity )
+        : _block_id_space( size[0] >> cell_bits_per_tile_dim,
+                           size[1] >> cell_bits_per_tile_dim,
+                           size[2] >> cell_bits_per_tile_dim,
+                           1 << bitCount( capacity ) )
+    {
+        std::fill( _min.data(), _min.data() + rank, 0 );
+        std::copy( size.begin(), size.end(), _max.data() );
+    }
+
+    /*!
+      \brief (Device) Insert a cell (given a cell ijk, insert the tile where the
+      cell reside in to hash table)
+      \param cell_i cell id in dim-x
+      \param cell_j cell id in dim-y
+      \param cell_k cell id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    void insertCell( int cell_i, int cell_j, int cell_k ) const
+    {
+        insertTile( cell_i >> cell_bits_per_tile_dim,
+                    cell_j >> cell_bits_per_tile_dim,
+                    cell_k >> cell_bits_per_tile_dim );
+    }
+
+    /*!
+      \brief (Device) Insert a tile (to hash table)
+      \param tile_i tile id in dim-x
+      \param tile_j tile id in dim-y
+      \param tile_k tile id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    void insertTile( int tile_i, int tile_j, int tile_k ) const
+    {
+        _block_id_space.insert( tile_i, tile_j, tile_k );
+    }
+
+    /*!
+      \brief (Device) Query the 1D tile key from the 3D cell ijk
+      \param cell_i cell id in dim-x
+      \param cell_j cell id in dim-y
+      \param cell_k cell id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    value_type queryTile( int cell_i, int cell_j, int cell_k ) const
+    {
+        // query the tile No.
+        auto tile_id = _block_id_space.find( cell_i >> cell_bits_per_tile_dim,
+                                             cell_j >> cell_bits_per_tile_dim,
+                                             cell_k >> cell_bits_per_tile_dim );
+        return tile_id;
+    }
+
+    /*!
+      \brief (Device) Query the 1D cell key from the 3D cell ijk
+      \param cell_i cell id in dim-x
+      \param cell_j cell id in dim-y
+      \param cell_k cell id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    value_type queryCell( int cell_i, int cell_j, int cell_k ) const
+    {
+        // query the tile No.
+        auto tile_id = queryTile( cell_i, cell_j, cell_k );
+        auto cell_id = _tile_id_space.coordToOffset(
+            cell_i & cell_mask_per_tile_dim, cell_j & cell_mask_per_tile_dim,
+            cell_k & cell_mask_per_tile_dim );
+        return static_cast<value_type>( ( tile_id << cell_bits_per_tile ) |
+                                        cell_id );
+    }
+
+    /*!
+      \brief (Host) Clear tile hash table (Host, required by unordered map
+      clear())
+    */
+    void clear() { _block_id_space.clear(); }
+
+    /*!
+      \brief Set new capacity lower bound on the unordered map (Host)
+      \param capacity New capacity lower bound
+    */
+    bool reserve( const value_type capacity )
+    {
+        return _block_id_space.reserve( capacity );
+    }
+
+    /*!
+     \brief (Host/Device) Require capacity of the index hash table
+    */
+    KOKKOS_INLINE_FUNCTION
+    uint32_t capacity() const { return _block_id_space.capacity(); }
+
+    /*!
+      \brief (Host) Valid tile number inside current block (MPI rank)
+    */
+    value_type size() const { return _block_id_space.validTileNumHost(); }
+
+  private:
+    //! block index space, map tile ijk to tile
+    BlockIndexSpace<MemorySpace, cell_bits_per_tile_dim, cell_num_per_tile_dim,
+                    cell_num_per_tile, hash_type, key_type, value_type>
+        _block_id_space;
+    //! tile index space, map cell ijk to cell local No inside a tile
+    TileIndexSpace<cell_bits_per_tile_dim, cell_num_per_tile_dim,
+                   cell_num_per_tile>
+        _tile_id_space;
+    //! space size (global), channel size
+    Kokkos::Array<int, rank> _min;
+    Kokkos::Array<int, rank> _max;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+  \class BlockIndexSpace
+  \brief Block index space, mapping tile ijks to tile No. through a hash table
+  (Kokkos unordered map)
+  \tparam CBits Bits number (per dimension) neded to
+  index cells inside each tile
+  \tparam CNumPerDim Cell number (per dimension)
+  inside each tile
+  \tparam CNumPerTile Cel number (total) inside each tile
+  \tparam MemorySpace Mem space to store unordered map
+  \tparam CBits Number of bits (per dimension) needed to index the cells inside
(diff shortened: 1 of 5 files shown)


cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+//---------------------------------------------------------------------------//
+// RUN TESTS
+//---------------------------------------------------------------------------//
+TEST( TEST_CATEGORY, sparse_index_space_test )
+{
+    testTileSpace();
+    // testBlockSpace<HashTypes::Naive>();
+    // testBlockSpace<HashTypes::Morton>();
+    // testSparseIndexSpace();
```
dalg24, line 173: Why are these commented out?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,525 @@
...
+        }
+    };
+
+    template <typename Func, typename... Coords>
+    KOKKOS_FORCEINLINE_FUNCTION static constexpr auto
+    from_coord( Coords &&... coords ) -> uint64_t
+    {
+        // if ( sizeof...( Coords ) != Rank )
+        //     throw std::invalid_argument( "Dimension of coordinate mismatch"
+        //     );
```
dalg24, line 468: This can be checked in a `static_assert`

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,525 @@
...
+    //! Coord  <=> Offset Computations
+    struct Coord2OffsetDim
+    {
+        template <typename Coord>
+        KOKKOS_FORCEINLINE_FUNCTION constexpr auto operator()( int dimNo,
+                                                               Coord &&i )
+            -> uint64_t
+        {
+            uint64_t result = (uint64_t)i;
+            for ( int i = 0; i < dimNo; i++ )
```
dalg24, line 444: Collision between name of the function parameter and the loop counter variable is unfortunate.  Consider renaming one of the two.

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: Still some changes are needed for these tests. 
I don't have enough local computer resources, so I want to check the tests one by one with CI. 
I will add them back.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+  private:
+    Kokkos::Array<int, 3> _tilenum;
+};
+
+template <typename Key>
+struct TileID2HashKey<Key, HashTypes::Morton>
+{
+    TileID2HashKey( std::initializer_list<int> &&tnum ) {} // ugly
```
dalg24, line 146: Same comment here and below.  Pass `std::initializer_list` by value.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,525 @@
...
+    HashKey2TileID( std::initializer_list<int> &&tnum )
+        : _tilenum()
+    {
+        std::copy( tnum.begin(), tnum.end(), _tilenum.data() );
+    }
+    HashKey2TileID( int i, int j, int k )
+    {
+        std::fill( _tilenum.data(), _tilenum.data() + 1, i );
+        std::fill( _tilenum.data() + 1, _tilenum.data() + 2, j );
+        std::fill( _tilenum.data() + 2, _tilenum.data() + 3, k );
```
dalg24, line 178: Same comment here.  Assign directly 
```suggestion
        _tilenum = { i, j, k };
```

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Sure, will fix these in the next commit

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Sure, will fix these in the next commit

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,386 @@
...
+#include <Kokkos_Core.hpp>
+
+#include <gtest/gtest.h>
+
+using namespace Cajita;
+
+namespace Test
+{
+//---------------------------------------------------------------------------//
+void testAtomicOr()
```
sslattery, line 12: @dalg24 This test seems to reproduce the `Kokkos::atomic_fetch_or` failure for HIP on the CI. Could you look at this and let us know if you see anything obvious in this code?

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,386 @@
...
+{
+    constexpr int size = 64;
+
+    Kokkos::View<bool ** * [2], TEST_MEMSPACE> _tile_insert_record(
```
sslattery, line 16: Let's try changing the `bool` to `int` here to see if that is the issue.

sslattery: Per @dalg24 we suspect we might be hitting this: https://github.com/kokkos/kokkos/blob/35613e69dcda1dd2274b66fa282d623add732dea/core/src/impl/Kokkos_Atomic_Generic.hpp#L403

Possibly the string showing the error message wasn't showing up in our log.

YuxingQiu: Please refer to this commit [cc716ad](https://github.com/ECP-copa/Cabana/pull/287/commits/cc716ad1eaa228171defdbbe8544f8fd89a53331)

YuxingQiu: Works when changing `bool` to `int`

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,441 @@
...
+            for ( int k = 0; k < size_per_dim; k++ )
+                EXPECT_EQ( qid_mirror( i, j, k ), cid_mirror( i, j, k ) );
+}
+
+//---------------------------------------------------------------------------//
+// RUN TESTS
+//---------------------------------------------------------------------------//
+TEST( sparse_grid, sparse_index_space_test )
+{
+    // testAtomicOr();
```
sslattery, line 421: Uncomment these tests so they run every time and make sure we don't hit the bug again.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,441 @@
...
+            for ( int tile_j = 0; tile_j < size; tile_j++ )
+                for ( int tile_k = 0; tile_k < size; tile_k++ )
+                {
+                    if ( Kokkos::atomic_fetch_or(
+                             &( _tile_insert_record( tile_i, tile_j, tile_k,
+                                                     0 ) ),
+                             1 ) == 0 )
+                    {
+                        tile_label( tile_i, tile_j, tile_k ) = true;
+                        // Kokkos::atomic_fetch_or(
```
sslattery, line 45: Remove the commented code if it is not being used.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,441 @@
...
+        Kokkos::RangePolicy<TEST_EXECSPACE>( 0, size ),
+        KOKKOS_LAMBDA( int tile_i ) {
+            for ( int tile_j = 0; tile_j < size; tile_j++ )
+                for ( int tile_k = 0; tile_k < size; tile_k++ )
+                {
+                    if ( Kokkos::atomic_fetch_or(
+                             &( _tile_insert_record( 0, 0, 0, 0 ) ), 1 ) == 0 )
+                    {
+                        tile_label( tile_i, tile_j, tile_k ) = true;
+                        // Kokkos::atomic_fetch_or(
```
sslattery, line 96: Remove the commented code if it is not being used

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+enum class HashTypes : unsigned char
+{
+    Naive = 0,
+    Morton = 1
+};
+
+//---------------------------------------------------------------------------//
+// bit operations for Morton Code computing
+template <typename Integer>
+CUDA_HOSTDEV constexpr Integer bit_length( Integer N ) noexcept
```
sslattery, line 29: For functions we use camel case starting with lower case. So this function would be called `bitLength()`. Please change the naming conventions for all functions.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
+#ifndef CAJITA_SPARSE_INDEXSPACE_HPP
+#define CAJITA_SPARSE_INDEXSPACE_HPP
+
+#include <Kokkos_Atomic.hpp>
+#include <Kokkos_Core.hpp>
+#include <Kokkos_UnorderedMap.hpp>
+
+#ifdef __CUDACC__
+#define CUDA_HOSTDEV __host__ __device__
```
sslattery, line 9: Kokkos has macros for this already. Delete this and just use `KOKKOS_INLINE_FUNCTION` everywhere that it might be a host/device function if the GPU is used.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+template <typename Integer>
+CUDA_HOSTDEV constexpr Integer bit_length( Integer N ) noexcept
+{
+    if ( N )
+        return bit_length( N >> 1 ) + static_cast<Integer>( 1 );
+    else
+        return 0;
+}
+
+template <typename Integer>
```
sslattery, line 37: Every function needs at least a comment proceeding it describing what it does, even if it is an implementation detail.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+KOKKOS_FORCEINLINE_FUNCTION
+constexpr int bit_pack( const uint64_t mask, const uint64_t data )
+{
+    uint64_t slresult = 0;
+    uint64_t &ulresult{slresult};
+    uint64_t uldata = data;
+    int count = 0;
+    ulresult = 0;
+
+    uint64_t rmask = binary_reverse( mask );
```
sslattery, line 74: Longer functions like this should have comments internally as well to describe what you are doing.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+        dat >>= 1, rmask <<= lz;
+    }
+    result = binary_reverse( result ) >> count_leading_zeros( mask );
+    return result;
+}
+//---------------------------------------------------------------------------//
+
+// Tile Hash Key (hash) <=> TileID
+template <typename Key, HashTypes HashT>
+struct TileID2HashKey;
```
sslattery, line 112: This naming convention is fine - we use upper case camel case for `struct` or `class` types

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+        std::copy( tnum.begin(), tnum.end(), _tilenum.data() );
+    }
+    TileID2HashKey( int i, int j, int k )
+    {
+        _tilenum[0] = i;
+        _tilenum[1] = j;
+        _tilenum[2] = k;
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION
```
sslattery, line 136: just make this `KOKKOS_INLINE_FUNCTION` - no need for the force here.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+    {
+        to_coord<Offset2CoordDim>( std::forward<Key>( key ), coords... );
+    }
+
+  protected:
+    //! Coord  <=> Offset Computations
+    struct Coord2OffsetDim
+    {
+        template <typename Coord>
+        KOKKOS_FORCEINLINE_FUNCTION constexpr auto operator()( int dimNo,
```
sslattery, line 480: for variables our naming convention is lower case separated by `_`. So in this case `dimNo` would be `dim_no`. Please change all of your variable names in this code to match this convention.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,564 @@
...
+    HashKey2TileID<KeyType, HashType> _op_key2ijk;
+};
+
+//---------------------------------------------------------------------------//
+template <int CBits, int CNumPerDim, int CNumPerTile>
+class TileIndexSpace
+{
+  public:
+    //! Number of cells inside each tile
+    static constexpr int CellBitsPerTileDim = CBits;
```
sslattery, line 455: We consider static class members like this to be variables so I would do `cell_bits_per_tile_dim` for this and other objects like it. `CBits` is OK because it is a template parameter and we do upper-case camel case for the template parameters.

sslattery (changes requested):
We need to start by cleaning up style/naming conventions and adding lots of comments to the code so it is clear what you are implementing - even in implementation details.

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Oh okay. Are you using [google name convention](https://google.github.io/styleguide/cppguide.html) or sth else? I can go over that and make changes accordingly!

cajita/src/Cajita_SparseIndexSpace.hpp
sslattery: Ours is a little different. Admittedly it is some combination I developed over the years. In hindsight maybe it would be better to use something more well-known but we will stick with this for now.

YuxingQiu: So is there any doc to follow the convention? I can make a changed version based on that so you don't need to comment line by line here.

streeve: I can add a style guide to the contributing doc if that's the best place

YuxingQiu: Thank you very much for providing this option. I can first follow other files/codes as style guides. I will post here if there are still problems!

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+    value_type queryCell( int cell_i, int cell_j, int cell_k ) const
+    {
+        // query the tile No.
+        auto tile_id = _block_id_space.find( cell_i >> cell_bits_per_tile_dim,
+                                             cell_j >> cell_bits_per_tile_dim,
+                                             cell_k >> cell_bits_per_tile_dim );
+        auto cell_id = _tile_id_space.coordToOffset(
+            cell_i & cell_mask_per_tile_dim, cell_j & cell_mask_per_tile_dim,
+            cell_k & cell_mask_per_tile_dim );
+        return ( value_type )( ( tile_id << cell_bits_per_tile ) | cell_id );
```
sslattery, line 452: use `static_cast<value_type>` here instead of the c-style cast.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+    }
+
```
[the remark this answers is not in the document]
sslattery: OK so `Kokkos::UnorderedMap` has a lot of memory management tools we can expose - e.g: https://github.com/kokkos/kokkos/blob/03ded4e816da5927b946065b005ebfdf38e5a438/containers/src/Kokkos_UnorderedMap.hpp#L333

We just need to poke some holes in the `BlockIndexSpace` to do this. Also `Kokkos::UnorderedMap` accepts a capacity hint which we will want to utilize. Note we can also `rehash` to reset the capacity.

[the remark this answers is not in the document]
sslattery: Another note on the unordered map - perhaps we can write our own implementation of their `Hasher` in the future to use Morton ordering?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+    BlockIndexSpace( const int size_x, const int size_y, const int size_z,
+                     const value_type capacity, const float rehash_factor )
+        : _tile_table_info( "hash_table_info" )
+        , _tile_inserted_label( "insert_label", size_x * size_y * size_z )
+        , _tile_table( size_x * size_y * size_z )
+        , _op_ijk2key( size_x, size_y, size_z )
+        , _op_key2ijk( size_x, size_y, size_z )
+    {
+        // hash table related init
+        _tile_table.clear();
```
sslattery, line 519: Do you need to `clear()` on construction?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+      \param cell_k cell id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    value_type queryCell( int cell_i, int cell_j, int cell_k ) const
+    {
+        // query the tile No.
+        auto tile_id = _block_id_space.find( cell_i >> cell_bits_per_tile_dim,
+                                             cell_j >> cell_bits_per_tile_dim,
+                                             cell_k >> cell_bits_per_tile_dim );
+        auto cell_id = _tile_id_space.coordToOffset(
```
sslattery, line 449: So for now we are ordering the tiles by whatever the `UnorderedMap` does for us while the cell ordering in a tile is either `Morton` or `Naive`?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+        _block_size[2] = size_z;
+        int total_size = size_x * size_y * size_z;
+        _rehash_coeff[0] = rehash_factor;
+
+        // insert status related init
+        auto tile_inserted_label_mirror = Kokkos::create_mirror_view(
+            Kokkos::HostSpace(), _tile_inserted_label );
+        for ( int i = 0; i < total_size; i++ )
+        {
+            tile_inserted_label_mirror( i ) = 0;
```
sslattery, line 538: for this one you can just do `Kokkos::deep_copy( _tile_inserted_label, 0 )` instead of the explicit mirror creation, loop, and copy.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+    */
+    struct Coord2OffsetDim
+    {
+        template <typename Coord>
+        KOKKOS_INLINE_FUNCTION constexpr auto operator()( int dim_no,
+                                                          Coord&& i )
+            -> uint64_t
+        {
+            uint64_t result = (uint64_t)i;
+            for ( int i = 0; i < dim_no; i++ )
```
sslattery, line 699: If we knew that we were always 3D we could hand unroll this loop and get rid of some of these multiply operations.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+      \brief Transfer function: tile ijk to tile key
+    */
+    struct Coord2OffsetDim
+    {
+        template <typename Coord>
+        KOKKOS_INLINE_FUNCTION constexpr auto operator()( int dim_no,
+                                                          Coord&& i )
+            -> uint64_t
+        {
+            uint64_t result = (uint64_t)i;
```
sslattery, line 698: use `static_cast` instead of c-style cast

sslattery (changes requested):
My review is still in progress but wanted to get these comments out there.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+        }
+        Kokkos::deep_copy( _tile_inserted_label, tile_inserted_label_mirror );
+    }
+
+    /*!
+      \brief Valid tile number inside current block (MPI rank)
+    */
+    KOKKOS_INLINE_FUNCTION
+    value_type validTileNum() const { return _tile_table_info( 0 ); }
+
```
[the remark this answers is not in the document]
sslattery: Good thing is if we do this and the table rehashes it isn't a problem because the map contents store the index we care about.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,424 @@
...
+//---------------------------------------------------------------------------//
+TEST( sparse_grid, sparse_index_space_test )
+{
+    testAtomicOr();
+    testAtomicOrPro();
+    testTileSpace();
+    testBlockSpace<HashTypes::Naive>();
+    testBlockSpace<HashTypes::Morton>();
+    testSparseIndexSpace();
+}
```
sslattery, line 420: After the above changes with respect to memory management are made let's add another test which checks inserting, clearing, and inserting again. You could even do a small number of inserts, clear, reserve more capacity, insert again with more stuff, and make sure it is correct.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+    /*!
+      \brief Insert a tile into the hash table
+      \param tile_i Tile Id in dim-x
+      \param tile_j Tile Id in dim-y
+      \param tile_k Tile Id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    void insert( int tile_i, int tile_j, int tile_k ) const
```
sslattery, line 556: Here is another way to implement this with only one atomic and without the `_tile_inserted_label_view` but instead just a global `value_type _num_tile` counter!
```
// Get the tile key
key_type tile_key = ijk2key( tile_i, tile_j, tile_k );

// Try to insert the tile key into the map. Give it a dummy value for now.
auto insert_result = _tile_table.insert( tile_key, 0 );

// If the tile key was actually inserted, atomically increment the counter so we know which 
// local id the tile has. Note that this atomic should be pretty scalable because 
// only threads that actually did a successful insert will call it so only one particle 
// in a tile will actually hit this call, substantially reducing contention.
if ( !insert_result.existing() )
    _tile_table.value_at( insert_result.index() ) = Kokkos::atomic_fetch_add( &_num_tile, 1 );
```

Note that `_num_tile` is a class member of type `value_type` that must be initialized to zero before any `insert` sequence. I think this is more scalable because it eliminates the storage of `_tile_inserted_label_view` completely, only has one atomic, and uses the thread-scalable insert in the unordered map instead to decide if the key should go into the table. 

I think this approach also lends itself to better OpenMP thread scaling and potentially a use for duplication in the future.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+      \param size_y The size of the block (MPI rank) in dim-y (Unit: tile)
+      \param size_z The size of the block (MPI rank) in dim-z (Unit: tile)
+      \param capacity Expected capacity of the allocator to store the tiles
+      \param rehash_factor The factor multiplies to the capacity for resizing,
+      when tile nums exceed the capcity
+    */
+    BlockIndexSpace( const int size_x, const int size_y, const int size_z,
+                     const value_type capacity, const float rehash_factor )
+        : _tile_table_info( "hash_table_info" )
+        , _tile_inserted_label( "insert_label", size_x * size_y * size_z )
```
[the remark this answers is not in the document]
sslattery: See my idea below for eliminating `_tile_inserted_table` below in favor of a single integer.

sslattery (changes requested):
This is looking great! 

I think I have a new way to implement `insert()` in the `BlockIndexSpace` would should improve performance and reduce storage. If you agree, please implement and see if it passes the test. Lots of other comments/discussion as well.

cajita/src/Cajita_SparseIndexSpace.hpp
sslattery: One more thing here - let’s have insert() return the inserted tile index and also have the insert function in the sparse space return the cell index - we may want to use it upon insertion

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+class BlockIndexSpace;
+
+//! Declaration of TileIndexSpace
+template <int CBits, int CNumPerDim, int CNumPerTile>
+class TileIndexSpace;
+
+/*!
+  \class SparseIndexSpace
+  \brief Sparse index space, with a hierarchical structure (cell->tile->block)
+  \tparam CellPerTileDim Cell number inside each tile per dimension
```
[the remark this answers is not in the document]
YuxingQiu: Currently, the tile id is decided by the inserting order of the tiles. The `HashType` would only determine the hash key of the tile. So without sorting, I cannot say if Morton is better or not. As for N, if there is no need to handle the channel here, I think we can delete this tparam and set the rank = 3. For `CellPerTileDim`, the user can define any number, e.g., 4 or 3.  My current code will enlarge this input to the `cell_num_per_tile_dim`, which is larger and the closest power-of-two to `CellPerTileDim`.

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: For now, we didn't sort by hand. Not sure if `UnorderedMap` will handle the ordering with the hashed key in their application.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,424 @@
...
+        for ( int j = 0; j < size[1]; j++ )
+            for ( int k = 0; k < size[2]; k++ )
+            {
+                test[qid_mirror( i, j, k )] = true;
+            }
+    for ( int i = 0; i < size[0]; i++ )
+        for ( int j = 0; j < size[1]; j++ )
+            for ( int k = 0; k < size[2]; k++ )
+            {
+                EXPECT_EQ( test[i * size[2] * size[1] + j * size[2] + k],
```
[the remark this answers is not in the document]
YuxingQiu: Added

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+    //! hash table (tile hash key => tile No)
+    Kokkos::UnorderedMap<key_type, value_type, ExecutionSpace> _tile_table;
+    //! Ops: transfer between tile ijk <=> tile hash key
+    TileID2HashKey<key_type, hash_type> _op_ijk2key;
+    HashKey2TileID<key_type, hash_type> _op_key2ijk;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+  \class BlockIndexSpace
```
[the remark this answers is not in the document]
YuxingQiu: Fixed

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+      \brief
+    */
+    template <typename Key, typename... Coords>
+    KOKKOS_INLINE_FUNCTION static constexpr void
+    offsetToCoord( Key&& key, Coords&... coords )
+    {
+        toCoord<Offset2CoordDim>( std::forward<Key>( key ), coords... );
+    }
+
+  protected:
```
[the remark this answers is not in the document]
YuxingQiu: Changed to private

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+        return to_coord_impl<Func>( 0, std::forward<Key>( key ), coords... );
+    }
+
+    /*!
+      \brief Implementation (coords => single number)
+      \param dim_no Current coord dimension
+      \param i The coord value
+    */
+    template <typename Func, typename Coord>
+    KOKKOS_INLINE_FUNCTION static constexpr auto from_coord_impl( int&& dim_no,
```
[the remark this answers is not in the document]
YuxingQiu: Fixed

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -119,44 +113,47 @@ struct HashKey2TileID;
...
+        _tile_num[1] = j;
+        _tile_num[2] = k;
     }
 
-    KOKKOS_FORCEINLINE_FUNCTION
+    KOKKOS_INLINE_FUNCTION
     Key operator()( int tile_i, int tile_j, int tile_k ) const
     {
-        return tile_i * _tilenum[1] * _tilenum[2] + tile_j * _tilenum[2] +
+        return tile_i * _tile_num[1] * _tile_num[2] + tile_j * _tile_num[2] +
```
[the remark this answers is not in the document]
YuxingQiu: Added a new variable to store it

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,817 @@
...
+ */
+template <typename ExecutionSpace, int N = 3,
+          unsigned long long CellPerTileDim = 4,
+          HashTypes Hash = HashTypes::Naive, typename Key = uint64_t,
+          typename Value = uint64_t>
+class SparseIndexSpace
+{
+  public:
+    //! Number of dimensions, 3 = ijk, or 4 = ijk + ch
+    static constexpr int rank = N;
```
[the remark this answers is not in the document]
YuxingQiu: Well, that's a good idea!!! I will assume we only serve 3D now. I have deleted the tparam `N`, and set rank = 3.

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: Added

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,681 @@
...
+            for ( int k = 0; k < size; k++ )
+            {
+                if ( tile_label_mirror( i, j, k ) == 1 )
+                    count++;
+            }
+
+    EXPECT_EQ( count, 1 );
+}
+
+void testTileSpace()
```
sslattery, line 112: Could this just be one test function templated on `size_bit` and `size`? To me the 2x2x2, 4x4x4, and 8x8x8 tests look the same except for the different `constexpr` integers.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,681 @@
...
+    for ( int i = 0; i < size; i++ )
+        for ( int j = 0; j < size; j++ )
+            for ( int k = 0; k < size; k++ )
+            {
+
+                EXPECT_EQ( test[i * size * size + j * size + k], true );
+            }
+}
+
+void testSparseIndexSpaceFullInsert()
```
sslattery, line 335: Need to test `capacity()`, `size()`, and `reserve()` in this test!

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,681 @@
...
+        EXPECT_EQ( test[idx], true );
+        ++idx;
+    }
+    while ( idx < total_size )
+    {
+        EXPECT_EQ( test[idx], false );
+        ++idx;
+    }
+
+    sis.clearIndexSpace();
```
sslattery, line 599: Should check `size()` and `capacity()` after this function is called to ensure correctness. `size()` should be zero and `capacity()` should be unchanged.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,681 @@
...
+    auto cbt = sis.cell_bits_per_tile;
+    EXPECT_EQ( cbt, 6 );
+
+    auto cnt = sis.cell_num_per_tile;
+    EXPECT_EQ( cnt, 64 );
+
+    Kokkos::View<int***, TEST_DEVICE> qid_res( "query_id", size[0], size[1],
+                                               size[2] );
+
+    Kokkos::parallel_for(
```
sslattery, line 363: should check the result of `size()` and `capacity()` after the insert to check for correctness. Can you set the initial capacity/rehash factor such that this insertion test would trigger a rehash? That way the capacity should change before and after the insert procedure.

sslattery (changes requested):
Almost there! Interfaces are good. Just a few questions and requested changes mostly in the unit tests.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,857 @@
...
+        std::copy( size.begin(), size.end(), _max.data() );
+    }
+
+    /*!
+      \brief Insert a cell (given a cell ijk, insert the tile where the cell
+      reside in to hash table) \param cell_i cell id in dim-x \param cell_j cell
+      id in dim-y \param cell_k cell id in dim-z
+    */
+    KOKKOS_INLINE_FUNCTION
+    void insertCell( int cell_i, int cell_j, int cell_k ) const
```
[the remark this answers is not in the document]
YuxingQiu: Yes, I believe it would be useful! Actually, I had this return in the beginning. But later due to my implementation of `insert` I had some data race so I have to remove this return. With the current implementation of `insert` I believe we can add it back,

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: That works. I will change that

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,857 @@
...
+    /*!
+      \brief Clear tile hash table (Host, required by unordered map clear())
+    */
+    void clear()
+    {
+        _tile_table.clear(); // clear hash table
+        auto tile_table_info_mirror =
+            Kokkos::create_mirror_view( Kokkos::HostSpace(), _tile_table_info );
+        tile_table_info_mirror( 0 ) = 0;
+        Kokkos::deep_copy( _tile_table_info,
```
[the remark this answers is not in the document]
YuxingQiu: I will change it to `create_mirror_view_and_copy`. In this way, the original setting of capacity would be reserved here. How do you like this method? Or should we provide interfaces for users to reset the capacity value(i.e. to indicate their preferred allocation size)?

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,857 @@
...
+    uint32_t capacity() const
+    {
+        return _tile_table.capacity(); // hash_table capacity
+    }
+
+    /*!
+      \brief Valid tile number inside current block (MPI rank)
+    */
+    KOKKOS_INLINE_FUNCTION
+    value_type validTileNum() const { return _tile_table_info( 0 ); }
```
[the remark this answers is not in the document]
YuxingQiu: Great suggestion! I added the `(Host)`/`(Device)`/`(Host/Device)` label to all functions.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,857 @@
...
+    using key_type = Key;                        // tile hash key
+    using value_type = Value;                    // tile No.
+    static constexpr HashTypes hash_type = Hash; // hash table
+
+    /*!
+      \brief Constructor
+      \param size The size of the block (MPI rank) (Unit: cell)
+      \param capacity Expected capacity of the allocator to store the tiles
+      \param rehash_factor The factor multiplies to the capacity for resizing,
+      when tile nums exceed the capcity
```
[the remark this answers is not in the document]
YuxingQiu: Yes, 0 should work. So let me make it clearer for the capacity(I'm not sure if it's the best design). We actually have two capacities — one for the unordered map, another for the pre-allocation size.

1. For the unordered map, I set the default size as the number of all uniform grids (`size_x * size_y * size_z`). The reason is that the insertion is executed on the device in a parallel way. During this process, it seems that Kokko unordered map cannot automatically resize. So I have to make sure that the inserting number does not exceed the hash table's capacity. If I do a full insertion (insert all uniform grids) in parallel with a smaller (unordered map) capacity, I would have a runtime error. Also, it would worth noticing that we cannot exactly control the capacity of Kokkos unordered map. The capacity value we feed to the unordered map would finally be a lower bound instead of an exact size. 

2. For the pre-allocation size, I use `capacity` and `rehash_factor` here for allocation resize purposes. (Well, when writing to this point, I think it would be better to rename these two variables to `prealloc_size` and `realloc_factor`). For this part, I only did the initialization operations. I plan to leave the resizes operations to the allocation PR to make things more consistent. Or maybe we can consider moving these two parameters allocation related structs? Still thinking about this design.

cajita/src/Cajita_SparseIndexSpace.hpp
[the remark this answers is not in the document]
YuxingQiu: I cannot reply to that comment on writing our own `Hasher` (no reply blank), so I will do here. I think it's workable to implement our own hasher. From my understanding, the reason for us to rewrite a hasher is that we hope (1) the blocks can be sorted according to their hashed key (2) to use a lighter/simpler hasher so we could probably have better performance. For the first reason, there can be two solutions, one is to explicitly sort after all insertion in each time step, another is to build up a hasher that has this sorting ability. For the second one, I believe that the Kokkos unordered map must have been well developed and optimized. So just not sure if our performance can beat the Kokkos version (囧).

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: Added

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: As I mentioned in the previous comments, the rehash factor doesn't really mean the capacity of the hash table. It's for allocation. The capacity of the hash table is initialized as the size of the full grid domain. Looking forward to your suggestions on this setting.

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: Added

streeve:
Retest this please

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,890 @@
...
+
+    /*!
+      \brief (Host) Constructor
+      \param size The size of the block (MPI rank) (Unit: cell)
+      \param capacity Expected capacity of the allocator to store the tiles
+      \param rehash_factor The factor multiplies to the capacity for resizing,
+      when tile nums exceed the capacity
+    */
+    SparseIndexSpace( const std::array<int, rank> size,
+                      const unsigned int capacity, const float rehash_factor )
```
sslattery, line 405: Can we just eliminate `rehash_factor`? It isn't used in your implementation as far as I can tell.

sslattery (changes requested):
Sorry I was wrong about the `insert()` return value - I can't see a way to make it thread safe!

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -0,0 +1,636 @@
...
+    uint32_t cap = sis.capacity();
+    EXPECT_EQ( cap >= total_tile_num, true );
+
+    auto s = sis.size();
+    EXPECT_EQ( s, total_tile_num );
+
+    uint32_t new_set_cap = total_tile_num * 10;
+    sis.reserve( new_set_cap );
+    uint32_t new_cap = sis.capacity();
+    EXPECT_EQ( new_cap >= new_set_cap, true );
```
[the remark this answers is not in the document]
YuxingQiu: The capacity is handled by the `capacity` interface from the Kokkos Unordered map. It can only ensure the lower bound but not the exact capacity value.

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -0,0 +1,890 @@
...
+        // now.
+        auto insert_result = _tile_table.insert( tile_key, 0 );
+        // If the tile key was actually inserted, atomically increment the
+        // counter Only threads that actually did a successful insert will call
+        // it
+        if ( !insert_result.existing() )
+            _tile_table.value_at( insert_result.index() ) =
+                Kokkos::atomic_fetch_add( &( _tile_table_info( 0 ) ), 1 );
+
+        return _tile_table.value_at( _tile_table.find( tile_key ) );
```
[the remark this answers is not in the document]
YuxingQiu: No problem. I will remove the return values and corresponding tests.

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Sure!

sslattery (approved):

PR #289, opened 2020-08-28
rearrange find_packages in Cabana_add_dependency()

sfogerty:
Currently, `Cabana_add_dependency()` calls `find_package()` two times if `Cabana_REQUIRE_$PACKAGE` is set to ON.
Unless guarded against in the target package, this causes a CMake error. (example: Heffte package for v1.0)

This change rearranges `Cabana_add_dependency()` to ensure `find_package()` is only called once.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ macro(Cabana_add_dependency)
   string(TOUPPER "${CABANA_DEPENDENCY_PACKAGE}" CABANA_DEPENDENCY_OPTION )
   option(
     Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION}
-    "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support" ${CABANA_DEPENDENCY_PACKAGE}_FOUND)
+    "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support")
   if(Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION})
     find_package( ${CABANA_DEPENDENCY_PACKAGE} REQUIRED )
   else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,13 +51,14 @@ endif()
 # standard dependency macro
 macro(Cabana_add_dependency)
   cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE" "" ${ARGN})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -51,13 +51,14 @@ endif()
 # standard dependency macro
 macro(Cabana_add_dependency)
   cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE" "" ${ARGN})
-  find_package( ${CABANA_DEPENDENCY_PACKAGE} QUIET )
   string(TOUPPER "${CABANA_DEPENDENCY_PACKAGE}" CABANA_DEPENDENCY_OPTION )
   option(
     Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION}
     "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support" ${CABANA_DEPENDENCY_PACKAGE}_FOUND)
```
dalg24, line 57:
```suggestion
    "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support")
```

CMakeLists.txt
```diff
@@ -51,13 +51,14 @@ endif()
 # standard dependency macro
 macro(Cabana_add_dependency)
   cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE" "" ${ARGN})
-  find_package( ${CABANA_DEPENDENCY_PACKAGE} QUIET )
   string(TOUPPER "${CABANA_DEPENDENCY_PACKAGE}" CABANA_DEPENDENCY_OPTION )
   option(
     Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION}
-    "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support" ${CABANA_DEPENDENCY_PACKAGE}_FOUND)
+    "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support")
```
sfogerty, line 57: @junghans Could the `*_FOUND` be removed, as shown here?

junghans: Then everything would be `OFF` by default.

sfogerty:
Based on our conversations, I will close this PR. 
I think relying on downstream packages to guard against multiple calls of `find_package()` isn't ideal, but there is only a specific issue in this instance with older versions of heFFTe. Maybe a future version of CMake will find a way to avoid this problem.

PR #292, opened 2020-09-03
CI: add cron to build again kokkos develop

junghans:


.github/workflows/CI.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
+on:
+  push:
+    branches:
+      - kokkos_develop_testing
```
junghans, line 5: @sslattery, it will also trigger when you push on the `kokkos_develop_testing` branch in case we ever have to debug this. Or we could also add a manual trigger.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,36 @@
...
+  schedule:
+    - cron:  '0 1 * * *'
```
junghans, line 7: Every day one minute after midnight, ok?

streeve:
Do we want to build tests as well?

streeve (approved):

.github/workflows/CI.yml
sslattery: Will this branch get automatically updated with the current state of `master`?

.github/workflows/CI.yml
sslattery: Works for me!

.github/workflows/CI.yml
```diff
@@ -0,0 +1,37 @@
...
+          cmake --build build --parallel 2
+          cmake --install build
+      - name: Checkout Cabana
+        uses: actions/checkout@v2.2.0
+      - name: Build Cabana
+        run: |
+          cmake -B build -DCMAKE_INSTALL_PREFIX=$HOME/Cabana -DCMAKE_PREFIX_PATH="$HOME/kokkos" -DCabana_ENABLE_TESTING=ON -DCabana_ENABLE_EXAMPLES=ON -DCabana_REQUIRE_${{ matrix.backend }}=ON
+          cmake --build build --parallel 2
+          cmake --build build --target test
+          cmake --install build
```
sslattery, line 37: Will this build with Cajita and MPI? I guess I'd like to build this with all of our dependencies if possible so every code path gets triggered.

sslattery (changes requested):
A question about how much is built. I would like to have everything build/run.

.github/workflows/CI.yml
streeve: Yeah, it found MPI and built with both. A good way to test without the `REQUIRE` variables

sslattery (approved):

PR #295, opened 2020-09-11
CommunicationPlan copy

streeve:
First, adding a test which fails in copying the `CommunicationPlan` due to duplicating communicators. Add missing copy constructor/assignment operator.

Found during periodic communication #263 with @sslattery 

Increased coverage by explicitly setting `MPI_EXEC_NUMPROCS=2` for travis (previously running everything one one rank)


core/unit_test/tstCommunicationPlan.hpp
```diff
@@ -575,6 +575,12 @@ void test7( const bool use_topology )
         EXPECT_EQ( n, host_steering( n ) );
 }
 
+void test_copy()
+{
+    Cabana::CommunicationPlan<TEST_DEVICE> comm_plan( MPI_COMM_WORLD );
```
sslattery, line 580: I would say this test is not sufficient - we want to be able to check that the state of the `CommunicationPlan` has not changed either. The operators you have written only duplicate the comm but we will also will need them to copy over the rest of the state (e.g. all of the `std::vector` objects, etc.)

sslattery (changes requested):

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -300,6 +300,23 @@ class CommunicationPlan
         MPI_Comm_dup( comm, &_comm );
     }
 
+    /*!
+      \brief Copy constructor.
+    */
+    CommunicationPlan( const CommunicationPlan &cp )
```
dalg24, line 306: You need to copy these guys too
https://github.com/ECP-copa/Cabana/blob/a8554efcfae9168e83868a173de58faecc7f8fc1/core/src/Cabana_CommunicationPlan.hpp#L821-L827

core/unit_test/tstCommunicationPlan.hpp
streeve: I copied inside an existing test instead - I can add more or check each member individually

core/src/Cabana_CommunicationPlan.hpp
streeve: Done

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -300,6 +300,37 @@ class CommunicationPlan
...
+    {
+        MPI_Comm_dup( cp.comm(), &_comm );
+        _neighbors = cp._neighbors;
+        _total_num_export = cp._total_num_export;
+        _total_num_import = cp._total_num_import;
+        _num_export = cp._num_export;
+        _num_import = cp._num_import;
+        _num_export_element = cp._num_export_element;
+        _export_steering = cp._export_steering;
```
dalg24, line 315: We need to discuss that copying the view.  If it was a deep copy, then `CommunicationPlan` would have value semantics.  But with that lien it neither has value nor reference semantics which makes it harder to reason about.

streeve:
Updated to match ArborX/ArborX#369

streeve:
It looks like the codecov patch is failing because we're only running with one MPI rank on travis (and therefore not testing a reasonable fraction of the `CommPlan`). @junghans is there an easy way to get this to run on at least 2 MPI ranks?

junghans:
Try to set `MPIEXEC_MAX_NUMPROCS=2`.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -460,11 +466,11 @@ class CommunicationPlan
 
         // Get the size of this communicator.
         int comm_size = -1;
-        MPI_Comm_size( _comm, &comm_size );
+        MPI_Comm_size( *_comm_ptr, &comm_size );
```
dalg24, line 469:
```suggestion
        MPI_Comm_size( comm(), &comm_size );
```

streeve: Fixed

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -762,7 +768,7 @@ class CommunicationPlan
...
```
dalg24, line 771:
```suggestion
        MPI_Comm_size( comm(), &comm_size );
```

streeve:
> Try to set `MPIEXEC_MAX_NUMPROCS=2`.

@junghans I added to Travis, but it didn't seem to change anything

junghans:
> > Try to set `MPIEXEC_MAX_NUMPROCS=2`.
> 
> @junghans I added to Travis, but it didn't seem to change anything

fixed

streeve:
@sslattery this is ready for re-review

sslattery (approved):

PR #298, opened 2020-09-21
Change clang format to specify pointer alignment middle

### LINKED ISSUE #297: Clang Format Inconsistencies
Currently we don't prescribe how pointers should be handled in the clang format file, and instead leave it to be inferred (via `DerivePointerAlignment`). This means: 

1. We have inconsistent formatting in the code; and 
2. It allows different versions of clang format to do different things (which is currently causing me issues, where my format disagrees with Travis). 

I propose we switch to specifying a value for `PointerAlignment`, and will make a PR. It looks like in the code currently we (mostly) do middle alignment for pointers and right alignment for references...

rfbird:
Currently we don't prescribe how pointers should be handled in the clang
format file, and instead leave it to be inferred (via
        DerivePointerAlignment). This means:

We have inconsistent formatting in the code; and It allows different
versions of clang format to do different things (which is currently
        causing me issues, where my format disagrees with Travis).

This sets pointer alignment for pointers and references to be "middle"

Fixes #297


rfbird:
Travis is failing because the two clang formats disagree about how to handle function call params with nested function calls. i.e.:

```
-    grid_parallel_for(
-        "fill_rank_1", TEST_EXECSPACE(), is1,
-        KOKKOS_LAMBDA( const int i ) { v1( i ) = 1.0; } );
```
vs
```
+    grid_parallel_for( "fill_rank_1", TEST_EXECSPACE(), is1,
+                       KOKKOS_LAMBDA( const int i ) { v1( i ) = 1.0; } );
```

Locally I can't get either version of clang-format (9 or 11) to agree with Travis, and can't find an option for handling next function calls (only some old docs saying it's not supported). 

Any ideas appreciated

dalg24:
Did you make sure you were using the exact same version of clang-format?

rfbird:
> Did you make sure you were using the exact same version of clang-format?

Nope, thats next. In general I'd prefer we prescribe the behavior (where possible) in the format file rather than relying on the user matching versions implicitly, etc

dalg24:
TO be more defensive you could also dump the entire configuration instead of relying implicitly on defaults like we do now.

rfbird:
> TO be more defensive you could also dump the entire configuration instead of relying implicitly on defaults like we do now.

That's a nice idea. Is there an elegant way to dump the config for travis, or shall I just hack it into the `.travis` file? (i.e add a needless call to `clang-format -style=file -dump-config`

rfbird:
I fed the dumped format of clang-format `v7` (which gives me results matching travis) into clang-format `v11` and it gives me a different answer to `v7`, so I think something internal to clang-format changed..

junghans:
format this please

rfbird:
> format this please

What happened here? 8 and 7 disagree? fun

junghans:
> > format this please
> 
> What happened here? 8 and 7 disagree? fun

We should maybe port the CI from travis to GitHub Actions.

gtest/gtest/gtest-all.cc
```diff
@@ -109,8 +109,8 @@
 #ifndef GTEST_INCLUDE_GTEST_GTEST_SPI_H_
 #define GTEST_INCLUDE_GTEST_GTEST_SPI_H_
 
-
-namespace testing {
+namespace testing
```
streeve, line 112: We should probably exclude gtest from formatting

PR #299, opened 2020-09-21
GHA: add format action

junghans:
This will add an action that reacts on "format this please" (full message) and format all code with clang-format (pick a version).

Useful for e.g. #298

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,47 @@
+name: Format
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  edit:
+    name: clang-format
+    if: ${{ github.event.comment.body == 'format this please' }}
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


rfbird (approved):
I like this idea, and the code looks pretty sane to me.

Let me test to see if v8 acts more like 7 or 9, and we can use that to pick a version. 

Clang 7 isn't all that old (September 2018) but I had to build it by hand to get that version on my machine

junghans:
We might need to  disable the check in Travis then.

rfbird:
> I like this idea, and the code looks pretty sane to me.
> 
> Let me test to see if v8 acts more like 7 or 9, and we can use that to pick a version.
> 
> Clang 7 isn't all that old (September 2018) but I had to build it by hand to get that version on my machine

Clang-format 8 acts like clang 7, so I think we're implicitly saying we want clang-format version <9

PR #300, opened 2020-09-22
GPU isolation on Jenkins CI builds

dalg24:
Some of the testing machines have multiple GPUs.  This ensures we only use the specified one which is defined on the Jenkins side as an environment variable.


sslattery (approved):

dalg24:
Retest this please

PR #301, opened 2020-09-24
Set max numprocs

streeve:
Set `MPIEXEC_MAX_NUMPROCS` for jenkins. 

Set unit test max procs and threads in the same way for both Cabana and Cajita


junghans (approved):

streeve:
@dalg24 just wanted to check if 2 ranks is most reasonable for jenkins

dalg24:
> @dalg24 just wanted to check if 2 ranks is most reasonable for jenkins

Machines have all at least 8 cores but a single GPU.

streeve:
@sslattery do you think we should stick with 2 ranks, or just 1 for the GPU?

sslattery:
> @sslattery do you think we should stick with 2 ranks, or just 1 for the GPU?

Let's just do 1 rank if there is only 1 GPU. Our current programming model assumes a single MPI rank per GPU. We will still get some MPI tests because of problems with periodic boundaries.

sslattery (approved):

PR #302, opened 2020-09-24
travis: update distro for new clang-format

junghans:


junghans:
As excepted still clang-7 even Ubuntu 20.04, let install the package explicitly.

junghans:
format this please

junghans:
@dalg24 @rfbird now everything uses clang-10 for formatting.

CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ Cabana_add_dependency( PACKAGE MPI )
 Cabana_add_dependency( PACKAGE ArborX )
 
 # find Clang Format
-Cabana_add_dependency( PACKAGE CLANG_FORMAT )
+find_package( CLANG_FORMAT )
```
sslattery, line 72: Does this imply that this is now a requirement?

.travis.yml
```diff
@@ -67,7 +67,8 @@ addons:
       - libopenmpi-dev
       - libgsl0-dev
       - libfftw3-dev
-      - python-pip
+      - python3-pip
+      - clang-format
```
sslattery, line 71: Does this need to specifically be `clang-format-10`?

sslattery (commented):
A few questions.

CMakeLists.txt
junghans: No. The clang_format just doesn't need our magic to create an `ENABLE_*` options, plus one could even see anymore what version was used, but we could certainly undo this change.

.travis.yml
junghans: We could do that, usually the clang never gets bump within a ubuntu release, so that should be ok for a while.

sslattery (approved):

rfbird:
I think we still need to specify pointer alignment, as right not it is not prescribed (and may be set to derived)

PR #303, opened 2020-09-26
Add .gitattributes to not ship files in release

junghans:


dalg24 (commented):
I would also list the `docker/` directory

junghans:
I thought these would be useful if a user ever wants to build a own container

dalg24:
> I thought these would be useful if a user ever wants to build a own container

Fair enough

sslattery (approved):

PR #306, opened 2020-10-01
Make clang-format bind pointers to the type (left)

### LINKED ISSUE #297: Clang Format Inconsistencies

rfbird:
Previously pointer alignment was "derived" based on context in the code.
This PR explicitly specifies it should be left binding, which should
give more consistent behaviour across clang format versions

The changes in this PR are cosmetic only.

Fixes #297

There is still a difference between how clang 9 and clang 10 format code. We can make 10 act like 9 by specifying `Cpp11BracedListStyle = false`, but I can't find a way to make 9 act like 10....

I'll try up the minimum version in another PR

### DIFF
--- a/cajita/src/Cajita_Halo.hpp
+++ b/cajita/src/Cajita_Halo.hpp
@@ -42,7 +42,7 @@ class HaloPattern
     virtual ~HaloPattern() = default;
 
     // Assign the neighbors that are in the halo pattern.
-    void setNeighbors( const std::vector<std::array<int, 3>> &neighbors )
+    void setNeighbors( const std::vector<std::array<int, 3>>& neighbors )
     {
         _neighbors = neighbors;
     }
@@ -67,7 +67,7 @@ class FullHaloPattern : public HaloPattern
             for ( int j = -1; j < 2; ++j )
                 for ( int k = -1; k < 2; ++k )
                     if ( !( i == 0 && j == 0 && k == 0 ) )
-                        neighbors.push_back( { i, j, k } );
+                        neighbors.push_back( {i, j, k} );
         this->setNeighbors( neighbors );
     }
 };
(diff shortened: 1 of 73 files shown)


cajita/src/Cajita_Halo.hpp
```diff
@@ -519,11 +519,11 @@ class Halo
     // alignment boundaries.
     template <class ArrayView>
     KOKKOS_INLINE_FUNCTION void
-    packElement( const Kokkos::View<char *, memory_space> &buffer,
-                 const Kokkos::View<int *[6], memory_space> &steering,
-                 const int element_idx, const ArrayView &array_view ) const
+    packElement( const Kokkos::View<char*, memory_space>& buffer,
+                 const Kokkos::View<int* [6], memory_space>& steering,
```
sslattery, line 523: Too bad these view data types get spaces in them.

sslattery (approved):

cajita/src/Cajita_Halo.hpp
rfbird: What would you prefer to see? I can try twiddle the options

sslattery: it would be nice if there was no space in Kokkos view data types but that might be hard.

rfbird: So ` Kokkos::View<int*[6],memory_space>&` but only for kokkos views? Yea, that sounds hard..

streeve:
Is this ready to merge?

rfbird:
> Is this ready to merge?

Yes from my perspective

PR #312, opened 2020-11-04
Require Kokkos 3.2

streeve:
Also re-enable HIP with `VerletList` hierarchical reductions

PR #314, opened 2020-11-05
Add OpenMPTarget support

streeve:
Add specializations for OpenMPTarget 
- `PerformanceTrait`
- `CommunicationPlan`

And disable a few things for now:
- CSR `VerletList`
- Some `ArrayOp`
- `grid_parallel_for`
- Create neighbor test random particles on the host

Needs #312 to be merged. Still needs CI testing (not all tests pass yet)

### DIFF
--- a/cajita/src/CMakeLists.txt
+++ b/cajita/src/CMakeLists.txt
@@ -23,6 +23,10 @@ set(HEADERS_PUBLIC
   Cajita_UniformDimPartitioner.hpp
   )
 
+if(Kokkos_ENABLE_OPENMPTARGET) #FIXME_OPENMPTARGET
+  list(REMOVE_ITEM HEADERS_PUBLIC Cajita_Parallel.hpp)
+endif()
+
 set(SOURCES_PUBLIC
   Cajita_LocalGrid.cpp
   Cajita_GlobalGrid.cpp
--- a/cajita/src/Cajita.hpp
+++ b/cajita/src/Cajita.hpp
@@ -26,7 +26,11 @@
 #include <Cajita_LocalMesh.hpp>
 #include <Cajita_ManualPartitioner.hpp>
 #include <Cajita_MpiTraits.hpp>
+
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
 #include <Cajita_Parallel.hpp>
+#endif
+
 #include <Cajita_ParameterPack.hpp>
 #include <Cajita_Partitioner.hpp>
 #include <Cajita_ReferenceStructuredSolver.hpp>
--- a/cajita/src/Cajita_Array.hpp
+++ b/cajita/src/Cajita_Array.hpp
@@ -518,6 +518,7 @@ struct DotFunctor
   array. This vector should be pre-sized to the number of degrees-of-freedom
   per entity.
 */
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
 template <class Array_t>
 void dot( const Array_t& a, const Array_t& b,
           std::vector<typename Array_t::value_type>& products )
@@ -541,6 +542,7 @@ void dot( const Array_t& a, const Array_t& b,
                    MpiTraits<typename Array_t::value_type>::type(), MPI_SUM,
                    a.layout()->localGrid()->globalGrid().comm() );
 }
+#endif
 
 //---------------------------------------------------------------------------//
 // Infinity norm
@@ -588,6 +590,7 @@ struct NormInfFunctor
   \param norms The norms for each degree-of-freedom in the array. This vector
   should be pre-sized to the number of degrees-of-freedom per entity.
 */
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
 template <class Array_t>
 void normInf( const Array_t& array,
               std::vector<typename Array_t::value_type>& norms )
@@ -611,6 +614,7 @@ void normInf( const Array_t& array,
                    MpiTraits<typename Array_t::value_type>::type(), MPI_MAX,
                    array.layout()->localGrid()->globalGrid().comm() );
 }
+#endif
 
 //---------------------------------------------------------------------------//
 // One norm
@@ -655,6 +659,7 @@ struct Norm1Functor
   \param norms The norms for each degree-of-freedom in the array. This vector
   should be pre-sized to the number of degrees-of-freedom per entity.
 */
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
 template <class Array_t>
 void norm1( const Array_t& array,
             std::vector<typename Array_t::value_type>& norms )
@@ -678,6 +683,7 @@ void norm1( const Array_t& array,
                    MpiTraits<typename Array_t::value_type>::type(), MPI_SUM,
                    array.layout()->localGrid()->globalGrid().comm() );
 }
+#endif
 
 //---------------------------------------------------------------------------//
 // Two norm
@@ -722,6 +728,7 @@ struct Norm2Functor
   \param norms The norms for each entity degree-of-freedom in the array. This
   vector should be pre-sized to the number of degrees-of-freedom per entity.
 */
+#ifndef KOKKOS_ENABLE_OPENMPTARGET
 template <class Array_t>
 void norm2( const Array_t& array,
             std::vector<typename Array_t::value_type>& norms )
@@ -748,7 +755,7 @@ void norm2( const Array_t& array,
     for ( auto& n : norms )
         n = std::sqrt( n );
 }
-
+#endif
 //---------------------------------------------------------------------------//
 
 } // end namespace ArrayOp
--- a/cajita/src/Cajita_Parallel.hpp
+++ b/cajita/src/Cajita_Parallel.hpp
@@ -17,6 +17,10 @@
 
 #include <string>
 
+#ifdef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
+#error Not supported when Kokkos::Experimental::OpenMPTarget backend enabled.
+#endif
+
 namespace Cajita
 {
 //---------------------------------------------------------------------------//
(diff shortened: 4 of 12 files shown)


cajita/src/Cajita_Parallel.hpp
```diff
@@ -17,6 +17,10 @@
 
 #include <string>
 
+#ifdef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
+#error Not supported when Kokkos::Experimental::OpenMPTarget backend enabled.
+#endif
```
dalg24, line 22: We could also catch that at configure time

cajita/unit_test/CMakeLists.txt
```diff
@@ -76,11 +76,14 @@ endmacro()
...
   GlobalMesh
   IndexSpace
-  Parallel
   ParameterPack
   Splines
   )
 
+if(NOT Kokkos_ENABLE_OPENMPTARGET) #FIXME_OPENMPTARGET
+  list(APPEND SERIAL_TESTS Parallel)
+endif()
```
dalg24, line 85: I would prefer conditionally removing from the list

core/src/impl/Cabana_PerformanceTraits.hpp
```diff
@@ -83,6 +83,16 @@ class PerformanceTraits<Kokkos::Experimental::HIP>
 };
 #endif
 
+//---------------------------------------------------------------------------//
+#if defined( KOKKOS_ENABLE_OPENMPTARGET )
+template <>
+class PerformanceTraits<Kokkos::Experimental::OpenMPTarget>
+{
+  public:
+    static constexpr int vector_length = 16;
```
dalg24, line 92: I would add a FIXME because this will depend on the architecture we are targeted.

core/unit_test/neighbor_unit_test.hpp
```diff
@@ -77,12 +77,14 @@ createParticles( const int num_particle, const double box_min,
...
-    using PoolType = Kokkos::Random_XorShift64_Pool<TEST_EXECSPACE>;
-    using RandomType = Kokkos::Random_XorShift64<TEST_EXECSPACE>;
+    using AoSoA_t_h = Cabana::AoSoA<DataTypes, Kokkos::HostSpace>;
+    AoSoA_t_h aosoa_host( "host_aosoa", num_particle );
+
+    auto position = Cabana::slice<0>( aosoa_host );
+    using PoolType =
+        Kokkos::Random_XorShift64_Pool<Kokkos::DefaultHostExecutionSpace>;
+    using RandomType =
+        Kokkos::Random_XorShift64<Kokkos::DefaultHostExecutionSpace>;
```
dalg24, line 87: I would use a preprocessor directive because I assuming we want it back on the device once it is supported?
I don't feel strongly about this.

core/unit_test/tstNeighborList.hpp
```diff
@@ -263,37 +263,49 @@ TEST( TEST_CATEGORY, linked_cell_stencil_test ) { testLinkedCellStencil(); }
 //---------------------------------------------------------------------------//
 TEST( TEST_CATEGORY, verlet_list_full_test )
 {
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
     testVerletListFull<Cabana::VerletLayoutCSR, Cabana::TeamOpTag>();
+#endif
     testVerletListFull<Cabana::VerletLayout2D, Cabana::TeamOpTag>();
 
-#ifndef KOKKOS_ENABLE_HIP // FIXME_HIP
+#ifndef KOKKOS_ENABLE_HIP          // FIXME_HIP
```
dalg24, line 271: We should not address it here but I did not realize this was still disabled.

streeve: I forgot about it. It's removed in #312 though

cajita/src/Cajita_Parallel.hpp
streeve: Done

core/src/impl/Cabana_PerformanceTraits.hpp
streeve: Done

core/unit_test/neighbor_unit_test.hpp
streeve: Done - I assume we'll have to do this for SYCL as well

dalg24 (approved):

dalg24:
Retest this please

cajita/src/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ set(HEADERS_PUBLIC
   Cajita_UniformDimPartitioner.hpp
   )
 
+if(Kokkos_ENABLE_OPENMPTARGET) #FIXME_OPENMPTARGET
+  list(REMOVE_ITEM HEADERS_PUBLIC Cajita_Parallel.hpp)
```
sslattery, line 27: Is the issue here the multidimensional parallelism?

cajita/src/Cajita.hpp
```diff
@@ -26,7 +26,11 @@
 #include <Cajita_LocalMesh.hpp>
 #include <Cajita_ManualPartitioner.hpp>
 #include <Cajita_MpiTraits.hpp>
+
+#ifdef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
```
sslattery, line 30: Based on the logic in `Cajita/src/CMakeLists.txt` don't you want to *not* include this if OpenMP Target is on?

cajita/unit_test/CMakeLists.txt
```diff
@@ -81,6 +81,10 @@ set(SERIAL_TESTS
   Splines
   )
 
+if(NOT Kokkos_ENABLE_OPENMPTARGET) #FIXME_OPENMPTARGET
```
sslattery, line 84: Logic here is also different than the original cmake file. Is this on or off when OpenMP Target is on?

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -59,6 +59,13 @@ struct CountSendsAndCreateSteeringAlgorithm<Kokkos::Experimental::HIP>
     using type = CountSendsAndCreateSteeringAtomic;
 };
 #endif // end KOKKOS_ENABLE_HIP
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+template <>
+struct CountSendsAndCreateSteeringAlgorithm<Kokkos::Experimental::OpenMPTarget>
```
sslattery, line 64: No duplication because ostensibly this is for a GPU which should have good atomics?

cajita/src/Cajita_Array.hpp
```diff
@@ -655,6 +659,7 @@ struct Norm1Functor
   \param norms The norms for each degree-of-freedom in the array. This vector
   should be pre-sized to the number of degrees-of-freedom per entity.
 */
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
```
sslattery, line 662: In general, why do the vector reductions not work?

core/src/impl/Cabana_PerformanceTraits.hpp
```diff
@@ -83,6 +83,16 @@ class PerformanceTraits<Kokkos::Experimental::HIP>
 };
 #endif
 
+//---------------------------------------------------------------------------//
+#if defined( KOKKOS_ENABLE_OPENMPTARGET )
+template <>
+class PerformanceTraits<Kokkos::Experimental::OpenMPTarget>
+{
+  public:
+    static constexpr int vector_length = 16; // FIXME_OPENMPTARGET
```
sslattery, line 92: @dalg24 Will the Kokkos interface have something like the warp size parameter based on the device this was compiled for?

core/unit_test/neighbor_unit_test.hpp
```diff
@@ -76,13 +76,23 @@ Cabana::AoSoA<Cabana::MemberTypes<double[3]>, TEST_MEMSPACE>
 createParticles( const int num_particle, const double box_min,
                  const double box_max )
 {
-    using DataTypes = Cabana::MemberTypes<double[3]>;
-    using AoSoA_t = Cabana::AoSoA<DataTypes, TEST_MEMSPACE>;
-    AoSoA_t aosoa( "aosoa", num_particle );
+    #ifdef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
```
sslattery, line 79: Will this not make the test spaces always use the host when OpenMP Target is used even when another backend may also be enabled in the build? Maybe just disable this test for now when the backend is OpenMP Target? You would need to use some type traits to do that effectively.

sslattery: Or is the issue random number generation isn't implemented? You could just build them on the host and copy them to whatever device type then to get this test to work.

sslattery (changes requested):

cajita/src/CMakeLists.txt
streeve: Yeah. So that affects quite a bit of Cajita, but this will at least enable what CabanaMD needs

cajita/src/Cajita.hpp
streeve: My bad

core/src/Cabana_CommunicationPlan.hpp
streeve: That was my assumption, but worth discussing

core/unit_test/neighbor_unit_test.hpp
streeve: Yeah, this is just temporarily switched to the host. I switched all random generation to the host first - I can go back to that to make it cleaner

cajita/unit_test/CMakeLists.txt
streeve: Done

cajita/src/Cajita.hpp
streeve: Fixed

cajita/unit_test/CMakeLists.txt
streeve: Fixed

cajita/src/Cajita_Array.hpp
streeve: Missed this one - these are also MD reductions, right?

sslattery (approved):

PR #315, opened 2020-11-06
Fix empty migrate

### LINKED ISSUE #313: Distributor that does not actually send data segfaults
If we have a situation where the distributor is called but doesn't have to do anything, it dies by going out of bounds trying to access an empty `_neighbor` array. I found this by trying to pass -1 to all values in the distributor 

i.e we can get to here while leaving the `_neighbor` array at size 0:

https://github.com/ECP-copa/Cabana/blob/a502acf1c37932dcda52248f33d2b12f90b934f2/core/src/Cabana_CommunicationPlan.hpp#L699

and then later we call `distirbutor.neighborRank` to check against our own rank:

https://github.com/ECP-copa/Cabana/blob/a502acf1c37932dcda52248f33d2b12f90b934f2/core/src/Cabana_Distributor.hpp#L185

streeve:
Make sure there is something to migrate before assuming there is at least one rank neighbor. 

This fixes #313

### DIFF
--- a/core/src/Cabana_CommunicationPlan.hpp
+++ b/core/src/Cabana_CommunicationPlan.hpp
@@ -799,7 +799,6 @@ class CommunicationPlan
             Kokkos::ViewAllocateWithoutInitializing( "export_steering" ),
             _total_num_export );
         auto steer_vec = _export_steering;
-        Kokkos::View<std::size_t*, memory_space> counts( "counts", num_n );
         Kokkos::parallel_for(
             "Cabana::createSteering",
             Kokkos::RangePolicy<execution_space>( 0, _num_export_element ),
--- a/core/src/Cabana_Distributor.hpp
+++ b/core/src/Cabana_Distributor.hpp
@@ -182,9 +182,10 @@ void distributeData(
     // therefore can be directly copied. If any of the neighbor ranks are this
     // rank it will be stored in first position (i.e. the first neighbor in
     // the local list is always yourself if you are sending to yourself).
-    std::size_t num_stay = ( distributor.neighborRank( 0 ) == my_rank )
-                               ? distributor.numExport( 0 )
-                               : 0;
+    std::size_t num_stay =
+        ( num_n > 0 && distributor.neighborRank( 0 ) == my_rank )
+            ? distributor.numExport( 0 )
+            : 0;
 
     // Allocate a send buffer.
     std::size_t num_send = distributor.totalNumExport() - num_stay;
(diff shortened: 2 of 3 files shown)


sslattery (approved):

rfbird (approved):
Looks good to me! Nice fix

streeve:
Retest this please

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -798,18 +798,23 @@ class CommunicationPlan
...
-                               neighbor_ids( i ) ) =
-                        ( use_iota ) ? i : element_export_ids( i );
-            } );
-        Kokkos::fence();
+        // It is possible to have export elements, but not any to export (e.g.
+        // all elements to be removed).
+        if ( _total_num_export > 0 )
+        {
+            auto steer_vec = _export_steering;
+            Kokkos::View<std::size_t*, memory_space> counts( "counts", num_n );
```
dalg24, line 806: I see that it was it before but why do we have this array in the first place?  It is not used.  Also how comes we don't get a warning about it?

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -798,18 +798,23 @@ class CommunicationPlan
...
-            KOKKOS_LAMBDA( const int i ) {
-                if ( element_export_ranks( i ) >= 0 )
-                    steer_vec( rank_offsets( element_export_ranks( i ) ) +
-                               neighbor_ids( i ) ) =
-                        ( use_iota ) ? i : element_export_ids( i );
-            } );
-        Kokkos::fence();
+        // It is possible to have export elements, but not any to export (e.g.
+        // all elements to be removed).
+        if ( _total_num_export > 0 )
```
dalg24, line 803: What was the issue with a parallel_for on an empty range?

core/src/Cabana_CommunicationPlan.hpp
streeve: Yeah, I'm really not sure why - I removed it

core/src/Cabana_CommunicationPlan.hpp
streeve: It was possible to go out of bounds for the steering vector - I made this more general

streeve:
Retest this please

core/src/Cabana_CommunicationPlan.hpp
streeve: Nevermind - I went back and the empty case doesn't make it here. I removed this

dalg24 (approved):

core/src/Cabana_CommunicationPlan.hpp
dalg24: That was my point.  This code path will not execute :)

PR #320, opened 2020-11-12
GHA: don't format gtest files

junghans:


streeve (approved):
I noticed the GHA format updated the pointer style in example files - how was that not failing or caught before?

PR #325, opened 2020-11-27
CI: Travis -> Github Actions

junghans:
Initial step in migrating from travis to GHA. WIll continue with #345


junghans:
@streeve some cajita tests fail on `fedora:rawhide` any ideas?

streeve:
I'm trying to recreate locally, but no luck yet

sslattery:
> @streeve some cajita tests fail on `fedora:rawhide` any ideas?

Man `IndexConversion` is a weird test to fail. Perhaps there is an uninitialized value issue somewhere I missed because it is dumping out `0`?

junghans:
> I'm trying to recreate locally, but no luck yet

Did you try the docker container?
```
docker run -it ghcr.io/kokkos/ci-containers/fedora:rawhide /bin/bash
```

streeve:
> Did you try the docker container?

Turns out that's super helpful. 

`fedora:rawhide` tests seem fine - only failing because of mpi (passed with `-oversubscribe`)
I reproduced the error with `fedora:latest`; it only affects periodic cases, but no fix yet

streeve:
So `Cajita_Halo` failure is the issue (causing the periodic `Cajita_IndexConversion` failures

streeve:
I haven't found the bug that causes `Halo` to fail on some distributions, but this now at least runs all the tests. I had to reduce printing in those failing tests (not sure why GHA doesn't truncate) and remove `ubuntu:devel` for now as it hangs

.github/workflows/CI2.yml
```diff
@@ -0,0 +1,90 @@
...
+            -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }}
+          cmake --build build --parallel 2
+          cmake --install build
+      - name: Checkout Cabana
+        uses: actions/checkout@v2.2.0
+      - name: Build Cabana
+        run: |
+          cmake -B build \
+            -DCMAKE_INSTALL_PREFIX=$HOME/Cabana \
+            -DMPIEXEC_MAX_NUMPROCS=2 -DMPIEXEC_PREFLAGS="--oversubscribe" \
```
sslattery, line 76: Is 2 MPI ranks our limit here? Do we just get one physical core per build? Would be nice to do 4 if possible - even if we oversubscribe in terms of MPI.

cajita/unit_test/tstHalo.hpp
```diff
@@ -54,6 +54,7 @@ int haloPad( Edge<D>, int d )
 template <class Array>
 void checkGather( const int halo_width, const Array& array )
 {
+    int pass_gather_test = 1;
```
sslattery, line 57: Were these changes part of trying to address the failing tests?

sslattery (approved):
Just a few basic questions. I see the Travis still showing up in the branch protection even though the `.travis.yml` has been removed - is there something else in the repo settings we need to change?

.github/workflows/CI2.yml
```diff
@@ -0,0 +1,90 @@
...
+    continue-on-error: ${{ matrix.kokkos_ver == 'develop' || matrix.distro == 'fedora:latest' || matrix.distro == 'fedora:intel' || matrix.distro == 'ubuntu:devel' }}
+    strategy:
+      matrix:
+        distro: ['fedora:latest', 'fedora:rawhide', 'opensuse:latest', 'ubuntu:latest', 'ubuntu:rolling'] # 'ubuntu:devel'
+        cxx: ['g++', 'clang++']
+        openmp: ['ON']
+        cmake_build_type: ['Debug', 'Release']
+        kokkos_ver: ['3.2.01', 'master', 'develop']
+        include:
+          - distro: 'fedora:intel'
```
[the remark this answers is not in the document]
streeve: line 17 adds `distro` for the full matrix with all options; those in `include` are more of a spot check - we could change which distributions are in which

cajita/unit_test/tstHalo.hpp
streeve: Yes, just limiting the prints because it was hanging

streeve:
> I see the Travis still showing up in the branch protection even though the `.travis.yml` has been removed - is there something else in the repo settings we need to change?

I don't have those permissions, but yes I think there's a setting to require Travis

cajita/unit_test/tstHalo.hpp
streeve: Do you want a `fixme` for each?

.github/workflows/CI2.yml
streeve: It must be one if we couldn't run without this. Maybe change that separately?

sslattery:
> > I see the Travis still showing up in the branch protection even though the `.travis.yml` has been removed - is there something else in the repo settings we need to change?
> 
> I don't have those permissions, but yes I think there's a setting to require Travis

OK - I'll add you to the list. We don't wan't to disable anyway until this one is merged. All other open PRs will also need to be rebased to get the new CI

cajita/unit_test/tstHalo.hpp
sslattery: yes let's add this. Seeing as it was distribution dependent we will really want to dig.

.github/workflows/CI2.yml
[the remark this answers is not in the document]
junghans: Yeah, I didn't wanted to build the full matrix.

.github/workflows/CI2.yml
junghans: Yeah, I think github actions has 4 cores, but let's bump that in an extra PR.

.github/workflows/CI2.yml
```diff
@@ -0,0 +1,90 @@
...
+            cxx: 'icpc'
+            openmp: 'ON'
+            cmake_build_type: 'Debug'
+            kokkos_ver: 'master'
+          - distro: 'fedora:intel'
+            cxx: 'icpc'
+            cmake_build_type: 'Release'
+            openmp: 'OFF'
+            kokkos_ver: 'master'
+            cxx_flags: ' -parallel'
```
junghans, line 38: Why is that needed? Especially because openmp is `OFF`.

streeve: It's only for the `peakflops` tests. So something about vectorization

junghans: I see, might be something to document.

Issue #323, opened 2020-11-20
Sparse Grids Roadmap


sslattery:
Initial rectilinear partitioning path:
- [x] Initial sparse grid functionality #287 
- [x] Rename `SparseIndexSpace` -> `SparseMap` (Also the `TileIndexSpace` and similarly named classes). Note that a user will be inputting global `ijk` cell ids into this structure. Perhaps we should add a `createSparseMap()` function which takes a `GlobalMesh<SparseMesh>` to populate the number of global tiles including padding and returns a `SparseMap`. @YuxingQiu #326 
- [x] Partitioner needs to be generalized such that communicators with the given topology and block id information is returned by the partitioner. For example, the constructor implementation of `GlobalGrid` will largely be moved into the partitioners. @YuxingQiu  #383; #464 
- [x] #534
- [x] `SparseArrayLayout` and `SparseArray` @YuxingQiu #546 
- [x] `Halo` specialization for local grid which defines the relevant gather/scatter operations. May simply just need to return 0 at first when outside of sparse domain. @Yuxing #560 

`Hypre` structured solvers mapped to `SparseArray` (future work)
~`Spline` infrastructure for `p2g` and `g2p` mapped to sparse space (this will probably just work).~

~ Unite load balancing interface (initially developed in #444) #541 ~ (separate issue)

Additional steps for non-rectilinear path:
- Develop new `GlobalGrid<SparseMesh>` specialization that will take and hold a `SparseMap` object. The `blockRank` will no longer be `ijk` but rather just the MPI rank. This information will be provided by the partitioner. Communicators will no longer have Cartesian topology - they will be general. We could consider trying out the general MPI topology feature. All indices are linearized so functions like `dimNumBlock()` are no longer needed. @sslattery 
- Develop new `SparseIndexSpace` using `SparseMap` and information from the `EntityType`, `Decomposition`, and `IndexType` which would be used in a `LocalGrid`. @YuxingQiu 
- `LocalGrid` specialization which defines partitioning strategy and returns `SparseIndexSpace<3>` based on `EntityType`, `Decomposition`, and `IndexType`. Our `GlobalGrid` specialization for sparse grids will have a partitioner which will help inform what things are owned and what are ghosted. This specialization should take a `SparseMap` in the constructor. @YuxingQiu 
- `LocalMesh` specialization for sparse grids which describes the local geometry as a function of the dense local indices. @YuxingQiu 
[...]

PR #326, opened 2020-12-02
Sparsemap

YuxingQiu:
First TODO for #323. 

1. Rename SparseIndexSpace -> SparseMap (Also the TileIndexSpace and similarly named classes). 
2. Add to comment: the inputting global ijk cell ids are global. 
3. Add a createSparseMap() function which takes a GlobalMesh<SparseMesh> and returns a SparseMap. Corresponding tests are also changed.
4. Change the name to distinguish between hash table capacity with the pre-allocation-size info.

### DIFF
--- a/cajita/src/Cajita_SparseIndexSpace.hpp
+++ b/cajita/src/Cajita_SparseIndexSpace.hpp
@@ -516,7 +516,7 @@ class SparseMap
 
 //---------------------------------------------------------------------------//
 // Creation function for SparseMap from GlobalMesh<SparseMesh>
-template <class Scalar, typename MemorySpace,
+template <typename MemorySpace, class Scalar,
           unsigned long long CellPerTileDim = 4,
           HashTypes Hash = HashTypes::Naive, typename Key = uint64_t,
           typename Value = uint64_t>
--- a/cajita/src/Cajita_SparseIndexSpace.hpp
+++ b/cajita/src/Cajita_SparseIndexSpace.hpp
@@ -337,27 +337,27 @@ struct HashKey2TileID<Key, HashTypes::Morton>
 
 //---------------------------------------------------------------------------//
 // Hierarchical index spaces
-// SparseIndexSpace <- BlockIndexSpace <- TileIndexSpace
+// SparseMap <- BlockMap <- TileMap
 // Naming:
 //      Block = MPI Rank
 //      Tile = Sub-block with several cells
 //      Cell = Basic grid unit
 //---------------------------------------------------------------------------//
 
-//! Declaration of BlockIndexSpace
+//! Declaration of BlockMap
 template <typename MemorySpace, unsigned long long CBits,
           unsigned long long CNumPerDim, unsigned long long CNumPerTile,
           HashTypes Hash, typename Key, typename Value>
-class BlockIndexSpace;
+class BlockMap;
 
-//! Declaration of TileIndexSpace
+//! Declaration of TileMap
 template <int CBits, int CNumPerDim, int CNumPerTile>
-class TileIndexSpace;
+class TileMap;
 
 /*!
-  \class SparseIndexSpace
+  \class SparseMap
   \brief Sparse index space, with a hierarchical structure (cell->tile->block)
-  \tparam MemorySpace Memory space to store the IndexSpace(Hash Table)
+  \tparam MemorySpace Memory space to store the Map(Hash Table)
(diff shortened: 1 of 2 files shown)


cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -500,23 +503,39 @@ class SparseIndexSpace
...
+    TileMap<cell_bits_per_tile_dim, cell_num_per_tile_dim, cell_num_per_tile>
         _tile_id_space;
     //! space size (global), channel size
     Kokkos::Array<int, rank> _min;
     Kokkos::Array<int, rank> _max;
 };
 
+//---------------------------------------------------------------------------//
+// Creation function for SparseMap from GlobalMesh<SparseMesh>
+template <class Scalar, typename MemorySpace,
```
sslattery, line 519: Reorder these such that `MemorySpace` comes before `Scalar`. That way if a user wants to just use the defaults then only `MemorySpace` is needed and `Scalar` will be deduced from the `global_mesh` parameter. In the current order a user would still need to provide `Scalar` even though it could be deduced.

cajita/unit_test/tstSparseIndexSpace.hpp
```diff
@@ -237,15 +236,29 @@ void testBlockSpace()
...
+        global_low_corner[0] + cell_size * global_num_cell[0],
+        global_low_corner[1] + cell_size * global_num_cell[1],
+        global_low_corner[2] + cell_size * global_num_cell[2] };
+    auto global_mesh = createSparseGlobalMesh(
+        global_low_corner, global_high_corner, global_num_cell );
+
+    // Create Sparse Map
+    // SparseMap<TEST_EXECSPACE> sis( size, pre_alloc_size );
+    auto sis =
+        createSparseMap<double, TEST_EXECSPACE>( global_mesh, pre_alloc_size );
```
sslattery, line 261: Once you have updated this above per my comment the only template parameter needed will be `TEST_MEMSPACE`

sslattery (changes requested):
Just one minor change. Otherwise looks good!

cajita/unit_test/tstSparseIndexSpace.hpp
YuxingQiu: That's true. Sounds great!

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -427,10 +429,11 @@ class SparseIndexSpace
...
 
     /*!
-      \brief (Device) Insert a tile (to hash table)
+      \brief (Device) Insert a tile (to hash table); Note that the tile ijk
+      should be global
       \param tile_i tile id in dim-x
-      \param tile_j tile id in dim-y
-      \param tile_k tile id in dim-z
+      \param tile_j tile id in
+      dim-y \param tile_k tile id in dim-z
```
streeve, line 436: typo here

cajita/src/Cajita_SparseIndexSpace.hpp
```diff
@@ -500,23 +503,39 @@ class SparseIndexSpace
...
     Kokkos::Array<int, rank> _max;
 };
 
+//---------------------------------------------------------------------------//
+// Creation function for SparseMap from GlobalMesh<SparseMesh>
+template <typename MemorySpace, class Scalar,
+          unsigned long long CellPerTileDim = 4,
+          HashTypes Hash = HashTypes::Naive, typename Key = uint64_t,
+          typename Value = uint64_t>
+SparseMap<MemorySpace, CellPerTileDim, Hash, Key, Value> createSparseMap(
```
streeve, line 523: Could return `auto`

YuxingQiu: Yes, that should be a more clean way. 
But I notice that almost all creation functions in Cajita use this way rather than auto, even though they may require lines of code for the return type. 
So shall I follow this coding style here? Or shall I change it to auto?

cajita/src/Cajita_SparseIndexSpace.hpp
YuxingQiu: Okay I've changed that.

cajita/src/Cajita_SparseIndexSpace.hpp
sslattery: I think this is fine for now

sslattery (approved):

PR #327, opened 2020-12-04
Add HIP backend for heFFTe

sfogerty:


sfogerty:
@streeve Jenkins fails when heFFTe looks for rocfft CMake config files. I tried to guess the correct directory for that, but it's nothing more than a guess

streeve:
Closing for #328 - just a Jenkins permissions issue here on your fork, @sfogerty

PR #329, opened 2021-01-12
Update license for 2021

streeve:
Only two of the cmake files have the license block. Do we want to add or remove?


sslattery (approved):
I don't see why we can't add the license to all files.

streeve:
> I don't see why we can't add the license to all files.

Added to cmake files and a few that were missing (peakflops and sparsegrid)

PR #330, opened 2021-01-14
Return const reference types instead of value types from SoAs

sslattery:


core/src/Cabana_Tuple.hpp
```diff
@@ -59,7 +59,7 @@ get( Tuple_t& tp )
 
 // Rank-0 const
 template <std::size_t M, class Tuple_t>
-KOKKOS_FORCEINLINE_FUNCTION typename Tuple_t::template member_value_type<M>
+KOKKOS_FORCEINLINE_FUNCTION typename Tuple_t::template member_const_reference_type<M>
```
dalg24, line 62:
```suggestion
KOKKOS_FORCEINLINE_FUNCTION decltype(auto)
```

streeve (approved):

sslattery:
format this please

core/src/Cabana_Tuple.hpp
dalg24: ping...

dalg24:
Also I thought we wanted to avoid format only commits

PR #334, opened 2021-01-19
Fixed cartesian grid off by one error

### LINKED ISSUE #333: CartesianGrid returns incorrect indicies for boundary points
In `Cabana_CartesianGrid.hpp` [locatePoint](https://github.com/ECP-copa/Cabana/blob/d46804f0e002be4d9f7740fc94a34c4ff1fe6aa3/core/src/impl/Cabana_CartesianGrid.hpp#L101)  returns incorrect i,j,k indices when a point is on the boundary. This can lead to incorrect cardinal indices. For example a 2x2x2 grid of cell size 1x1x1 with` locatePoint(2.0, 2.0, 2.0, ic, jc, kc) `sets (ic = 2, jc = 2, kc = 2). This behaviour is only correct if we assume particles can be in the domain, but not on the boundary. The use of `floor` [cellsBetween](https://github.com/ECP-copa/Cabana/blob/d46804f0e002be4d9f7740fc94a34c4ff1fe6aa3/core/src/impl/Cabana_CartesianGrid.hpp#L148) leads to this off by one error.

ejm930:
A simple fix to the boundary error. The logic is a bit redundant in the sense that every particle is checked by it even though almost none will meet this condition. 

Fixes #333

### DIFF
--- a/core/performance_test/peakflops/common.h
+++ b/core/performance_test/peakflops/common.h
@@ -1,5 +1,7 @@
 /*
-  Note that the performance output on skylake can be about twice as the expected flops/cycle. This is because the rdtsp() is off by about a factor of two on this cpu.
+  Note that the performance output on skylake can be about twice as the expected
+  flops/cycle. This is because the rdtsp() is off by about a factor of two on
+  this cpu.
   TODO: remove rdtsp() and use flops/sec for measuring performance.
  */
 #ifndef PEAKFLOPS_COMMON_H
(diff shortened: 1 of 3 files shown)


sslattery:
reformat this please

sslattery:
@junghans Does our reformat auto-command still work?

streeve:
format this please

streeve:
Can you add a unit test check for this case?

ejm930:
Sorry about the format error, still trying to figure out clang-format. Added a simple unit test check.

sslattery (approved):

streeve (approved):

PR #328, opened 2020-12-08
Add HIP heFFTe, update FFT data types, and test with rocm 4.0

streeve:
Closes #319 

This PR does two things. First, to simplify the FFT interface, Kokkos Views with a pair of float/double is cast to `std::complex` when passed to heFFTe (rather than separate device complex types and `Kokkos::complex`). This also enabled adding FFTs with HIP, using the heFFTe rocm backend.

Testing with `rocm 4.0` was added (the `-complete` docker image comes with `rocff`)

PR #336, opened 2021-01-21
Change FFT complex datatypes

sfogerty:
Update heFFTe version used to 2.0.0

Instead of using a different complex data type for each heFFTe backend, simplify to using a Kokkos View of floats or doubles. This allows us to remove all the copy functions that are needed to shuffle between the specific types of complex arrays and Kokkos::complex<fp_type>

Note that directly using Kokkos::complex<fp_type> isn't supported by heFFTe because it may use architecture-dependent striding, with gaps that heFFTe doesn't expect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ before_script:
       popd && popd;
     fi
   - if [[ ${HEFFTE} ]]; then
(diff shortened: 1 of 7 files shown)


streeve:
Will close this once #328 passes

PR #339, opened 2021-01-28
Fix interpolation test 

sslattery:
The Cajita interpolation test had a bug where the wrong index space was used to loop over the test results. 

In addition, I was getting random failures for non-threaded backends on the test when `ScatterView` was used with the `Serial` backend. I believe that this was the result of threaded duplication still occurring so I updated to unit test harness to explicitly set `--kokkos-threads=1` when backends that are not threaded are used. That fixed my local test results. Curious if that will also fix the failures in #325

### DIFF
--- a/cajita/unit_test/tstInterpolation2d.hpp
+++ b/cajita/unit_test/tstInterpolation2d.hpp
@@ -73,6 +73,9 @@ void interpolationTest()
             points( pid, Dim::J ) = x[Dim::J];
         } );
 
+    // Node space.
+    auto node_space = local_grid->indexSpace( Own(), Node(), Local() );
+
     // Create a scalar field on the grid.
     auto scalar_layout = createArrayLayout( local_grid, 1, Node() );
     auto scalar_grid_field =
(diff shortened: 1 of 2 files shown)


streeve (approved):
Nice catch!

streeve:
Also, I checked and this doesn't seem to affect #325

streeve:
Were we going to add a `FIXME` comment for the single thread cases?

junghans (approved):

streeve:
Even without the explicit `--kokkos-threads=1` I don't have any issues locally. @sslattery are you okay with me removing that? I think that's the only thing that stopped us from merging in the first place

sslattery:
> Even without the explicit `--kokkos-threads=1` I don't have any issues locally. @sslattery are you okay with me removing that? I think that's the only thing that stopped us from merging in the first place

Yeah let's remove that and then I'll build locally and see how successful I am.

streeve (approved):
I updated 2D and everything looks okay on my end

Issue #340, opened 2021-01-28
Variable dimension


streeve:
Update Cajita to enable 2D (potentially 1D) particle simulations, where currently 3D is assumed throughout: 

- [ ] Global grids (uniform/non-uniform)
- [ ] Sparse grid
- [ ] Interpolation
- [ ] Halo

Eventually, update Cabana to also enable 2D/1D:

- VerletList
- ArborX Neighborlist
- Linked Cell List

PR #341, opened 2021-01-29
Mesh Tag Dimension

sslattery:
This PR adds a spatial dimension to the mesh tag as the first task in #340 to propagate general dimensions through the code. 

The new spatial dimension template is set to a default of `3` to allow for backwards-compatibility with existing implementations.

Note that I also removed the manual ETI for now on `GlobalMesh`, `GlobalGrid`, and `LocalGrid` as the type combinations were growing substantially with the addition of the new dimension template parameter in the mesh tag. We can revisit the choice of ETI as necessary if build types are significant in the future but this did not seem to affect my local build performance.


cajita/src/Cajita_GlobalMesh.cpp
```diff
@@ -14,13 +14,18 @@
 namespace Cajita
 {
 
-template class GlobalMesh<UniformMesh<float>>;
-template class GlobalMesh<UniformMesh<double>>;
+#define CAJITA_INST_GLOBALMESH( MESH, FP, NSD )                                \
```
streeve, line 17: Your description mentioned removing all explicit instantiation. Did you mean to keep this? Or keep `GlobalGrid`? `LocalGrid` seems to be the only one too large to manage

sslattery: Ah missed this - just deleted that file.

streeve (approved):

PR #342, opened 2021-01-29
Partitioner Variable Dimension

sslattery:
Adds a spatial dimension template to the partitioner base class and existing implementations. 

Note: This is actually a backwards compatibility break due to the addition of the template parameter where there was none previously. Hence `UniformDimPartitioner` is now `UniformDimPartitioner<3>` for 3D grids. I have updated the unit test to reflect this. This isn't a big change and easy to fix but it will break old code. I tried to find a way around this but we want that dimension template exposed in the virtual function that executes the partitioning and hence the base class could not be template-free. Furthermore, the derived classes definitely need that template argument.

This PR is part of #340 and depends on #341.


sslattery:
Ready for review

streeve (approved):
Looks good - the only issue is backwards compatibility which, as we've discussed, is not ideal, but not unreasonable

kwitaechong (approved):
Looks good!

dalg24 (commented):
Alternative is you give the templated version a new name and let the old one be an alias for 3D

sslattery:
> Alternative is you give the templated version a new name and let the old one be an alias for 3D

Great idea - I just made a push that should be backwards compatible. At some point we may want to deprecate the old implementations.

cajita/src/Cajita_Partitioner.hpp
```diff
@@ -19,20 +19,92 @@
...
+//
+// Use MPI to compute the most uniform block distribution possible (i.e. the
+// one that has the minimal number of neighbor communication messages in halo
+// exchange). This distribution is independent of mesh parameters - only the
+// size of the communicator is considered. Depending on the problem, this may
+// not be the optimal partitioning depending on cell-counts and workloads as
+// the reduced number of MPI messages may be overshadowed by the load
+// imbalance during computation.
+template <std::size_t NumSpaceDim>
+class DimBlockPartitioner : public BlockPartitioner<NumSpaceDim>
```
streeve, line 89: I like this since it often won't be uniform

streeve (approved):

PR #344, opened 2021-01-30
Local Grid Variable Dimension

sslattery:
Adds variable dimension capability to the local grid. Index spaces generated are now of size equal to the spatial dimension of the mesh. 3D and 2D interfaces are present for backwards compatibility but a general array-based interface is now also added where needed to enable a single implementation for all dimensions. K-faces and all edges are not supported in 2D and are 3D-only.

Overall the complexity of the `public` interface has been reduced dramatically. Most of the function overloading has moved into `private` implementation functions.

Part of #340. Depends on #343

### DIFF
--- a/cajita/src/Cajita_LocalGrid.hpp
+++ b/cajita/src/Cajita_LocalGrid.hpp
@@ -16,7 +16,9 @@
 #include <Cajita_IndexSpace.hpp>
 #include <Cajita_Types.hpp>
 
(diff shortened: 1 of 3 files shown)


sslattery:
@streeve This one is next for #340

streeve (approved):
Looks good - some nice public simplifications in the restructuring

PR #346, opened 2021-02-01
Add utils package

streeve:
Create a separate `utils` package to share capability across Cabana `core` and `cajita`. Currently contains `Version`, `ParameterPack` (this enables #296), and unit test backend files


cajita/src/Cajita_Halo.hpp
```diff
@@ -14,7 +14,8 @@
 
 #include <Cajita_Array.hpp>
 #include <Cajita_IndexSpace.hpp>
-#include <Cajita_ParameterPack.hpp>
+
+#include <Cabana_ParameterPack.hpp>
```
sslattery, line 18: Did you mean to include `<Cabana_Utils.hpp>` here?

cajita/src/Cajita_Halo.hpp
```diff
@@ -540,11 +541,12 @@ class Halo
                const Kokkos::View<int* [6], memory_space>& steering,
                const int element_idx,
                const std::integral_constant<std::size_t, 0>,
-               const ParameterPack<ArrayViews...>& array_views )
+               const Cabana::ParameterPack<ArrayViews...>& array_views )
```
sslattery, line 544: Note that this change of namespace for `ParameterPack` from `Cajita` to `Cabana` is a backwards compatibility break. We are already planning a bunch of small ones anyway elsewhere (e.g. in #342) so this doesn't really bother me that much.

utils/unit_test/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
+############################################################################
```
sslattery, line 1: We don't need to do it in this PR but we should use this new utils package as an opportunity to unify our test harness and have a single version sitting in `utils/`

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -12,7 +12,7 @@
 #ifndef CABANA_COMMUNICATIONPLAN_HPP
 #define CABANA_COMMUNICATIONPLAN_HPP
 
-#include <CabanaCore_config.hpp>
+#include <Cabana_config.hpp>
```
sslattery, line 15: Did you mean to include `Cabana_Utils.hpp`

sslattery (approved):
Just a few small things. Does this file need to be updated? https://github.com/ECP-copa/Cabana/blob/master/cmake/CabanaConfig.cmakein

cajita/src/Cajita_ParameterPack.hpp
```diff
@@ -12,169 +12,45 @@
...
-    static constexpr std::size_t size = sizeof...( Types );
-};
-
-//---------------------------------------------------------------------------//
 // Static type checker.
-template <class>
-struct is_parameter_pack_impl : public std::false_type
+template <class T>
+struct is_parameter_pack : public Cabana::is_parameter_pack_impl<
+                               typename std::remove_cv<T>::type>::type
```
dalg24, line 22: Already exists in `utils/src/Cabana_ParameterPack.hpp`

dalg24:
What else than the parameter pack utility do you want to move to that utils package?

Because these changes feel a bit silly if all you need if move parameter pack to Cabana.

cajita/src/Cajita_Halo.hpp
streeve: Fixed

cajita/src/Cajita_Halo.hpp
streeve: Added a wrapper to keep backwards compatibility

utils/unit_test/CMakeLists.txt
streeve: Done

core/src/Cabana_CommunicationPlan.hpp
streeve: Fixed

cajita/src/Cajita_ParameterPack.hpp
streeve: This is already used downstream

streeve:
> What else than the parameter pack utility do you want to move to that utils package?
> 
> Because these changes feel a bit silly if all you need if move parameter pack to Cabana.

Fair point, but we want to use it across both cajita and core. What would you suggest?

And we could do it separately, but it's a good place for the shared testing infrastructure

sslattery:
> > What else than the parameter pack utility do you want to move to that utils package?
> > Because these changes feel a bit silly if all you need if move parameter pack to Cabana.
> 
One thing to note is that I actually use the `ParameterPack` downstream in other libraries for meta-programming to capture variadic parameters on device. I could argue that if I wanted `ParameterPack` and nothing else in Cajita (e.g. I was doing particles only) then I would be building Cajita just for that one class. I can see some of the base tools in `core/src/impl` going in there or tools such as the `Version` code. Also a place to put the shared test harness.

dalg24:
> > What else than the parameter pack utility do you want to move to that utils package?
> > Because these changes feel a bit silly if all you need if move parameter pack to Cabana.
> 
> Fair point, but we want to use it across both cajita and core. What would you suggest?
> 
> And we could do it separately, but it's a good place for the shared testing infrastructure

You already have the Cajita dependency on Cabana, so I'd put it in Cabana.

streeve:
Closing in favor of #363 and #364

PR #347, opened 2021-02-01
Variable dimension local mesh

sslattery:
Adding variable dimensions (in particular 3d and 2d) to `Cajita::LocalMesh`. The interface implementation is backwards compatible. Again note, as in the other PRs in #340 that 3D meshes have cells, faces, edges, and nodes which 2D meshes have cells, faces, and nodes in only the `i`, and `j` dimensions.

I added a separate test for 2D which may seem like it largely duplicates the 3D test but in fact the interface functions used to get at data are largely dimension-dependent and therefore the majority of a combined 2d/3d test would have had much of the code in dimension-based if/else or enable_if statements.

Part of #340. Depends on #344

### DIFF
--- a/cajita/src/Cajita_LocalMesh.hpp
+++ b/cajita/src/Cajita_LocalMesh.hpp
@@ -37,7 +37,7 @@ class LocalMesh<Device, UniformMesh<Scalar, NumSpaceDim>>
     using scalar_type = Scalar;
 
     // Spatial dimension.
-    static constexpr int num_space_dim = NumSpaceDim;
+    static constexpr std::size_t num_space_dim = NumSpaceDim;
 
     // Device type.
     using device_type = Device;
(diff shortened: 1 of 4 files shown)


sslattery:
Note that the changes will look absurd until I rebase on upstream PRs after they are merged.

sslattery:
@streeve @kwitaechong This one is next for #340

kwitaechong:
Looks good to me

kwitaechong (approved):

streeve (approved):
I assume you split the test files because it got so large. Does that make sense for any others?

sslattery:
> I assume you split the test files because it got so large. Does that make sense for any others?

yeah I started doing that for a lot of them as it helped keep everything straight with the extensive interface overloads that needed to be tested for 2d vs. 3d. You will see in some of the upcoming PRs that I also did this splitting.

PR #349, opened 2021-02-01
Variable Dimension Array

sslattery:
Part of #340. Depends on #347

### DIFF
--- a/cajita/src/Cajita_Array.hpp
+++ b/cajita/src/Cajita_Array.hpp
@@ -21,6 +21,7 @@
 
 #include <cmath>
 #include <memory>
+#include <type_traits>
 #include <vector>
 
 #include <mpi.h>
@@ -39,6 +40,9 @@ class ArrayLayout
 
     // Mesh type.
     using mesh_type = MeshType;
+
+    // Spatial dimension.
+    static constexpr std::size_t num_space_dim = mesh_type::num_space_dim;
 
     /*!
       \brief Constructor.
@@ -65,8 +69,8 @@ class ArrayLayout
     // Get the index space of the array elements in the given
     // decomposition.
     template <class DecompositionTag, class IndexType>
-    IndexSpace<4> indexSpace( DecompositionTag decomposition_tag,
-                              IndexType index_type ) const
+    IndexSpace<num_space_dim + 1>
+    indexSpace( DecompositionTag decomposition_tag, IndexType index_type ) const
     {
         return appendDimension( _local_grid->indexSpace( decomposition_tag,
                                                          EntityType(),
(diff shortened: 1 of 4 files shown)


sslattery:

cajita/src/Cajita_Array.hpp
```diff
@@ -471,31 +545,59 @@ void update( Array_t& a, const typename Array_t::value_type alpha,
...
+        KOKKOS_LAMBDA( const long i, const long j, const long l ) {
+            a_view( i, j, l ) =
+                alpha * a_view( i, j, l ) + beta * b_view( i, j, l );
+        } );
+}
+
 //---------------------------------------------------------------------------//
 // Dot product
-template <class ViewType>
+template <class ViewType, std::size_t NumSpaceDim>
```
streeve, line 568: Should these structs have a default? Not sure if it makes sense to use directly in an application code

streeve (approved):

cajita/src/Cajita_Array.hpp
```diff
@@ -471,31 +545,59 @@ void update( Array_t& a, const typename Array_t::value_type alpha,
         } );
 }
 
+template <class Array_t, class DecompositionTag>
+std::enable_if_t<2 == Array_t::num_space_dim, void>
+update( Array_t& a, const typename Array_t::value_type alpha, const Array_t& b,
+        const typename Array_t::value_type beta, DecompositionTag tag )
+{
+    static_assert( is_array<Array_t>::value, "Cajita::Array required" );
```
kwitaechong, line 553: Do we assume that Array a and b are made from the same layout? do we need to check their extent of indexspace?

cajita/src/Cajita_Array.hpp
```diff
@@ -204,16 +233,11 @@ class Array
...
-                 layout->indexSpace( Ghost(), Local() ).extent( 1 ) ||
-             (long)view.extent( 2 ) !=
-                 layout->indexSpace( Ghost(), Local() ).extent( 2 ) ||
-             (long)view.extent( 3 ) !=
-                 layout->indexSpace( Ghost(), Local() ).extent( 3 ) )
-            throw std::runtime_error(
-                "Layout and view dimensions do not match" );
+        for ( std::size_t d = 0; d < num_space_dim + 1; ++d )
+            if ( (long)view.extent( d ) !=
+                 layout->indexSpace( Ghost(), Local() ).extent( d ) )
```
kwitaechong, line 238: It seems Array is designed over ghost index space but ArrayOp is over own index space. Any reason?

cajita/unit_test/tstArray2d.hpp
```diff
@@ -0,0 +1,388 @@
...
+        for ( long j = 0; j < sub_ghosted_space.extent( Dim::J ); ++j )
+            for ( long l = 0; l < sub_ghosted_space.extent( 2 ); ++l )
+                EXPECT_EQ( host_subview( i, j, l ), 3.0 * scales[l + 2] + 1.0 );
+
+#ifndef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
+    // Compute the dot product of the two arrays.
+    std::vector<double> dots( dofs_per_cell );
+    ArrayOp::dot( *array, *array_2, dots );
+    int total_num_node = global_grid->globalNumEntity( Node(), Dim::I ) *
+                         global_grid->globalNumEntity( Node(), Dim::J );
```
kwitaechong, line 311: it is valid since it is periodic condition. If it is non-periodic, it needs to be Cell()

kwitaechong (approved):

cajita/src/Cajita_Array.hpp
sslattery: I don't think a default is needed here - especially since `NumSpaceDim` can be deduced from the `Array` used to get the view to construct the functor.

cajita/src/Cajita_Array.hpp
sslattery: `a` and `b` are both `Array_t` and hence this won't compile if they are of the same type and have the same layout. I do agree though that they could technically have different sizes so we should check that. Let's make a new PR for that as this PR is just to add 2D/3D

cajita/src/Cajita_Array.hpp
sslattery: The idea here is that point-wise operations can happen on own or own+ghost entities. I wanted to give the user control of this in case they had a large halo region but only needed the owned values update. This is especially true in some PIC variants were we let the halo get pretty large.

cajita/unit_test/tstArray2d.hpp
sslattery: Actually you are right! This should be `Cell`. It is a typo.

Issue #345, opened 2021-02-01
Migrate to Github Actions


streeve:
Updates needed with the switch to GHA from Travis: 
- [x] Initial version (#325)
- [x] Add format (#351)
- [x] Clean up style (e.g. remove `cd build`) (#377)

GHA has shown that some distributions that fail (`fedora:latest` `fedora:intel` and `ubuntu:devel`)
- [x] Fix the `Cajita::Halo`

Features to implement in GHA:
- [x] HEFFTE FFTW build (#370)
- [x] HEFFTE MKL build (#370, disabled until fixed in heFFTe)
- [x] ARBORX build (#359)
- [x] Deploy doxygen (#396)
- [x] Enable ccache (#384)
- [x] HYPRE build (#385)

junghans:
Thanks for writing this up.

junghans:
We also need investigate why some tests fail on `fedora:latest` and `ubuntu:devel`.

streeve:
All distributions are now passing - unsure why they originally failed and why they all started passing at the same time

junghans:
All done.

PR #351, opened 2021-02-02
Add format step

streeve:


.github/workflows/CI2.yml
```diff
@@ -127,5 +127,10 @@ jobs:
           cmake --build build --parallel 2
           CTEST_OUTPUT_ON_FAILURE=1 cmake --build build --target test
           cmake --install build
+      - name: Format Cabana
```
sslattery, line 130: Curious if this will find clang format 10

streeve: We're getting the default: https://github.com/kokkos/ci-containers/blob/071c72d137a9fa8f203dc5bc2b4572be8cc210cc/ubuntu#L8 

Which looks to be working

sslattery (approved):

.github/workflows/CI2.yml
```diff
@@ -127,5 +127,11 @@ jobs:
...
+        if: ${{ matrix.distro == 'ubuntu:latest' }}
+        run: |
+             cd build
```
junghans, line 133: I would have used `working-directory: build` instead.

streeve:
#345

PR #353, opened 2021-02-03
Variable Dimension Spline

sslattery:
Adds general spatial dimension template to splines and spline data containers. A 2D test is also added.

Part of #340.

### DIFF
--- a/cajita/src/Cajita_Splines.hpp
+++ b/cajita/src/Cajita_Splines.hpp
@@ -492,71 +492,75 @@ struct has_spline_tag<T, SplineDataMemberTypes<T, Tags...>> : std::true_type
 };
 
 // Spline data members.
-template <typename Scalar, int Order, class Tag>
+template <typename Scalar, int Order, std::size_t NumSpaceDim, class Tag>
 struct SplineDataMember;
 
-template <typename Scalar, int Order>
-struct SplineDataMember<Scalar, Order, SplinePhysicalCellSize>
+template <typename Scalar, int Order, std::size_t NumSpaceDim>
+struct SplineDataMember<Scalar, Order, NumSpaceDim, SplinePhysicalCellSize>
 {
     // Physical cell size.
-    Scalar dx[3];
+    Scalar dx[NumSpaceDim];
 };
 
-template <typename Scalar, int Order>
-struct SplineDataMember<Scalar, Order, SplineLogicalPosition>
+template <typename Scalar, int Order, std::size_t NumSpaceDim>
+struct SplineDataMember<Scalar, Order, NumSpaceDim, SplineLogicalPosition>
 {
     // Logical position.
-    Scalar x[3];
+    Scalar x[NumSpaceDim];
 };
 
-template <typename Scalar, int Order>
-struct SplineDataMember<Scalar, Order, SplinePhysicalDistance>
+template <typename Scalar, int Order, std::size_t NumSpaceDim>
+struct SplineDataMember<Scalar, Order, NumSpaceDim, SplinePhysicalDistance>
 {
     using spline_type = Spline<Order>;
     static constexpr int num_knot = spline_type::num_knot;
 
     // Physical distance.
-    Scalar d[3][num_knot];
+    Scalar d[NumSpaceDim][num_knot];
 };
 
-template <typename Scalar, int Order>
-struct SplineDataMember<Scalar, Order, SplineWeightValues>
+template <typename Scalar, int Order, std::size_t NumSpaceDim>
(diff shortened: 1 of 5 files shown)


sslattery:

streeve:
Waiting on review here until we discuss more on the backwards compatibility

cajita/src/Cajita_Splines.hpp
```diff
@@ -593,172 +600,203 @@ struct SplineData<Scalar, Order, EntityType, SplineDataMemberTypes<Tags...>>
...
-KOKKOS_INLINE_FUNCTION void
-evaluateSpline( const LocalMesh<Device, UniformMesh<Scalar>>& local_mesh,
-                const Scalar p[3],
-                SplineData<Scalar, Order, EntityType, DataTags>& data )
+template <typename Scalar, int Order, std::size_t NumSpaceDim, class Device,
+          class EntityType, class DataTags>
+KOKKOS_INLINE_FUNCTION void evaluateSpline(
+    const LocalMesh<Device, UniformMesh<Scalar, NumSpaceDim>>& local_mesh,
+    const Scalar p[NumSpaceDim],
+    SplineData<Scalar, Order, NumSpaceDim, EntityType, DataTags>& data )
```
sslattery, line 756: I think the main issue here is the `NumSpaceDim` template in `SplineData` in terms of backwards compatibility - the rest of the objects here are implementation details. I don't think we can move this to the back of the list because `DataTags` is variadic and thus will have a variable length. I would argue the best way to use `SplineData` is to always have it be a generic template in your function. I think we need a `createSplineData` function that would return `SplineData` with the proper types via deduction and a user can just assign this to `auto`. 

Ultimately this doesn't fix the backwards compatibility issue so I think we might need to break it on this one.

@streeve comments?

streeve: Well we already missed one class for full backwards compatibility, so I think one more is fine. Unless you want to rename and wrap the existing with an explicit 3.

cajita/unit_test/tstSplineEvaluation2d.hpp
```diff
@@ -0,0 +1,448 @@
...
+
+    // Check the point set data.
+    EXPECT_EQ( point_set.num_point, num_point );
+    EXPECT_FLOAT_EQ( point_set.dx, cell_size );
+    EXPECT_FLOAT_EQ( point_set.rdx, 1.0 / cell_size );
+    double xn_low[2];
+    int idx_low[2] = { 0, 0 };
+    local_mesh.coordinates( Node(), idx_low, xn_low );
+    for ( int d = 0; d < 2; ++d )
+        EXPECT_EQ( point_set.low_corner[d], xn_low[d] );
```
streeve, line 315: Should this be `EXPECT_FLOAT_EQ`? There are a few in the 3D that are also still `EXPECT_EQ`

cajita/unit_test/tstSplineEvaluation2d.hpp
```diff
@@ -0,0 +1,448 @@
...
+
+    auto local_mesh = createLocalMesh<Kokkos::HostSpace>( local_grid );
+
+    int idx_low[2] = { 0, 0 };
+    int idx_low_p1[2] = { 1, 1 };
+    double lowx[2];
+    double lowxp1[2];
+    local_mesh.coordinates( Node(), idx_low, lowx );
+    local_mesh.coordinates( Node(), idx_low_p1, lowxp1 );
+    point_set.dx = lowxp1[0] - lowx[0];
```
streeve, line 251: Could this just use `measure`?

streeve (approved):
Looks good to me

streeve:
Also curious to see if this passes all tests with new commits/rebase

sslattery:
> Also curious to see if this passes all tests with new commits/rebase

Same tests fail here as previously.

cajita/unit_test/tstSplineEvaluation2d.hpp
sslattery: This was a way to do this in a dimensionally agnostic way. It would need to be for edges in 3d and faces in 2d. This way I know I always get the cell size. I guess we could also add a special function to the uniform mesh variation of local mesh which returns the cell size directly.

cajita/unit_test/tstSplineEvaluation2d.hpp
sslattery: yeah lets use float here - I will fix

cajita/src/Cajita_Splines.hpp
sslattery: we have discussed this and are going to move forward with the current structure.

PR #354, opened 2021-02-03
Variable Dimension Interpolation

sslattery:
Adds variable dimension interpolation for spline operations.

Part of #340. Depends on #353


sslattery:

streeve (approved):
Looks good to me

PR #355, opened 2021-02-03
Variable Dimension Structured Solvers

sslattery:
Adds variable dimensions to both Hypre linear solvers and the reference linear solver implementations.

Part of #340.


sslattery:
@kwitaechong @streeve This is the last one!

streeve (approved):
That's all of them!

PR #360, opened 2021-02-16
Add embedded neighbor iteration within parallel_for

### LINKED ISSUE #357: Device neighbor iterating
`neighbor_parallel_for` is a host function that asks for a functor, and thus enforcing that the whole kernel equals to neighbour interaction functor. 

In addition to `neighbor_parallel_for` I'd suggest a function for on-device iterating, e.g. called `neighbor_for`, that calls some device functor. This is far more common in my case then interaction-only kernels.

I don't know your coding style, but `neighbor_for` can even be in a struct containing all vars if you don't like the syntax that is passing template params and variables in a kernel.

```
template< ... >
struct NeighborIterating {
// NeighborListType& list;     ... other vars ...     ... constructor ...

template<class F>
KOKKOS_INLINE_FUNCTION void neighbor_for(index_type i, F functor) const {
  for ( index_type n = 0;  n < numNeighbor( list, i ); ++n )
    functor(getNeighbor(list, i, n));
}

};
```

Or simply I can just call by myself `numNeighbor` & `getNeighbor` inside the kernel :) ...

streeve:
Closes #357 

Only done for first nearest neighbors (so far)

### DIFF
--- a/core/src/Cabana_Parallel.hpp
+++ b/core/src/Cabana_Parallel.hpp
@@ -260,11 +260,11 @@ inline void neighbor_parallel_for(
   \brief Execute \c functor in parallel according to the execution \c policy
   with thread-local serial loops over particle first and second neighbors.
 
-  \tparam ExecutionSpace The execution space in which to execute the functor.
-
   \tparam FunctorType The functor type to execute.
 
   \tparam NeighborListType The neighbor list type.
+
+  \tparam ExecParameters The Kokkos execution policy parameters.
 
   \param exec_policy The policy over which to execute the functor.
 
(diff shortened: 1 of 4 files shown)


streeve:
@sslattery Ready for review

streeve:
@sslattery did you want me to rename?

sslattery:
> @sslattery did you want me to rename?

Yes - what do you think about `for_each_neighbor`?

streeve:
> Yes - what do you think about for_each_neighbor?

I like that - just couldn't remember

sslattery (approved):

Issue #361, opened 2021-02-18
Make library header-only


sslattery:
Steps to achieve this:

- [x] Remove `Cabana_Version.cpp` from `core/` and inline the implementations in `Cabana_Version.hpp`
- [x] Remove `Cabana_IndexSpace.cpp`
- [x] Update CMake install to be header-only. We would still create targets though.
- [x] Check build against CabanaMD downstream

PR #362, opened 2021-02-18
Let cabanacore and Cajita be header-only libraries

dalg24:
Partially addresses #361


core/src/Cabana_Version.hpp
```diff
@@ -19,9 +19,9 @@
 namespace Cabana
 {
 
-std::string version();
+std::string version() { return Cabana_VERSION_STRING; }
```
streeve, line 22: These need to be inlined to build downstream

dalg24: Good catch.  Multiple definition error at link time otherwise

streeve (approved):
Working with CabanaMD

dalg24:
> Working with CabanaMD

Thank you for checking

junghans (approved):

junghans:
Yeah to modern CMake!

sslattery (approved):

PR #364, opened 2021-02-19
Add Cajita dependency on core

streeve:
Move ParameterPack to core; keep wrapper in Cajita for backwards compatibility (could deprecate now if we want)

Enables #296


streeve:
I marked as draft because linking Cajita to core breaks downstream: 
```
CMake Error at CMakeLists.txt:31 (find_package):
  Found package configuration file:

    /Users/5t2/Codes/Cabana/build/install/lib/cmake/Cabana/CabanaConfig.cmake

  but it set Cabana_FOUND to FALSE so package "Cabana" is considered to be
  NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing:
  Cabana::cabanacore
```
Not sure what I'm doing wrong

streeve:
> I marked as draft because linking Cajita to core breaks downstream

Fixed with config ordering

cajita/src/Cajita_ParameterPack.hpp
```diff
@@ -12,169 +12,45 @@
...
+template <class T>
+struct is_parameter_pack : public Cabana::is_parameter_pack_impl<
+                               typename std::remove_cv<T>::type>::type
 {
 };
 
+// Create a parameter pack.
 template <typename... Types>
-struct is_parameter_pack_impl<ParameterPack<Types...>> : public std::true_type
+Cabana::ParameterPack<Types...> makeParameterPack( const Types&... ts )
```
sslattery, line 28: Why duplicate this function here?

cajita/src/Cajita_ParameterPack.hpp
```diff
@@ -12,169 +12,45 @@
...
-template <typename ParameterPack_t, std::size_t N, typename T,
-          typename... Types>
-void fillParameterPackImpl( ParameterPack_t& pp,
-                            const std::integral_constant<std::size_t, N>,
-                            const T& t, const Types&... ts )
-{
-    get<ParameterPack_t::size - 1 - N>( pp ) = t;
-    fillParameterPackImpl( pp, std::integral_constant<std::size_t, N - 1>(),
-                           ts... );
+    return Cabana::get<N, ParameterPack_t>( pp );
```
sslattery, line 51: I think we can probably remove this file altogether. This is a relatively new expert-level capability so I'm fine changing namespaces to avoid confusion.

sslattery (changes requested):
Looks good other than the possibility of removing `Cajita_ParameterPack.hpp` altogether.

cajita/src/Cajita_ParameterPack.hpp
streeve: It was only for full backwards compatibility. I can remove

sslattery (approved):

cajita/src/Cajita_Halo.hpp
```diff
@@ -14,7 +14,8 @@
 
 #include <Cajita_Array.hpp>
 #include <Cajita_IndexSpace.hpp>
-#include <Cajita_ParameterPack.hpp>
+
+#include <Cabana_Core.hpp>
```
dalg24, line 18: Any reason to include this and not just `<Cabana_ParameterPack.hpp>`?

core/unit_test/tstParameterPack.hpp
```diff
@@ -28,14 +26,14 @@ void captureTest()
...
-    auto pack = makeParameterPack( dbl_view, int_view );
+    auto pack = Cabana::makeParameterPack( dbl_view, int_view );
 
     // Update the pack in a kernel
     Kokkos::parallel_for(
         "fill_pack", Kokkos::RangePolicy<TEST_EXECSPACE>( 0, 1 ),
         KOKKOS_LAMBDA( const int ) {
-            auto dv = get<0>( pack );
-            auto iv = get<1>( pack );
+            auto dv = Cabana::get<0>( pack );
```
dalg24, line 35: (Not asking you to change just making sure you know) I expect you don't absolutely have to qualify the call, that the function would be found by ADL (argument-dependent lookup)

cajita/src/Cajita_Halo.hpp
streeve: Done

PR #368, opened 2021-02-24
Use external gtest

### LINKED ISSUE #365: Should we take GoogleTest as an external package instead of carrying our own copy?
I started wondering this while reviewing #363.
Instead of unconditionally creating the `cabana_gtest` library we could optionally find it as an external project.  Any opinion on this?

streeve:
Remove gtest snapshot and download externally. Fixes #365


streeve:
The current failure reminded me of the previous attempt: #195

cmake/test_harness/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+############################################################################
+
+##---------------------------------------------------------------------------##
+# Download and unpack googletest
+##---------------------------------------------------------------------------##
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG        release-1.10.0
```
junghans, line 19:
```suggestion
set(GTEST_URL "https://github.com/google/googletest/archive/release-1.10.0.tar.gz" CACHE STRING "URL for GTest tarball")
mark_as_advanced(GTEST_URL)
FetchContent_Declare(
  googletest
  URL ${GTEST_URL}
  URL_MD5         ecd1fa65e7de707cd5c00bdac56022cd
```
something like this.

streeve: Added

junghans (approved):

PR #370, opened 2021-03-03
GHA: add heffte builds

junghans:
Part of #345


junghans:
Needs https://github.com/kokkos/ci-containers/pull/8

.github/workflows/CI2.yml
```diff
@@ -21,107 +21,130 @@ jobs:
...
         coverage: ['OFF']
         include:
           - distro: 'fedora:intel'
             cxx: 'icpc'
             openmp: 'ON'
             cmake_build_type: 'Release'
             kokkos_ver: 'master'
-            arborx: ['NoArborX']
+            arborx: 'NoArborX'
+            heffte: 'NoheFFTe'
```
streeve, line 33: We should add to at least one intel build with MKL

junghans:
@streeve the containers works now, can you try your luck on the mkl builds?

.github/workflows/CI2.yml
```diff
@@ -166,14 +205,33 @@ jobs:
...
+        run: |
+          git clone --depth 1 --branch v2.0.0 https://bitbucket.org/icl/heffte.git heffte
+      - name: Build heffte
+        if: ${{ matrix.heffte != 'NoheFFTe' }}
+        working-directory: heffte
+        run: |
+          cmake -B build \
+            -DCMAKE_CXX_STANDARD="11" \
+            -DBUILD_SHARED_LIBS=ON \
+            -DHeffte_ENABLE_${{ matrix.heffte }}=ON \
```
junghans, line 220: Uh, that is ugly.

streeve: I don't love it. Would you prefer to build the list of optional flags like we used to with travis?

junghans: we could do the same in shell above
```
[[ ${{ matrix.heffte }} = "heFFTe_with_FFTW" ]] && cmake_opts+=(  -DHeffte_ENABLE_FFTW=ON )
[[ ${{ matrix.heffte }} = "heFFTe_with_MKL" ]] && cmake_opts+=(  -DHeffte_ENABLE_MKL=ON )
```

junghans (commented):
Looks good now!

junghans:
Intel fails with:
```
43 - Cabana_NeighborList_test_SERIAL (Failed)
```

junghans:
@streeve any idea?

streeve:
Here's what Miro said about the tests with heFFTe MKL build failures:
> The problem is with the newer versions of the Intel compiler and the way that they infer and convert template parameters. I will have to add a work-around and be explicit, which is not the most convenient thing. Clang, Gcc, and XL work fine, so I have to work-around an outlier compiler (frustrating).
If you can downgrade the version in the Intel compiler that will give you a temporary work-around.

We can discuss what we want to do with those tomorrow

streeve:
@sslattery ready for review, then we can reset the required tests again

sslattery (approved):

PR #372, opened 2021-03-04
Performance benchmarks: neighbor and linkedcell

streeve:
Add neighbor list and linked cell list performance tests matching bin/sort and using benchmark timers. This converts the the previous neighbor benchmark

Issue #335, opened 2021-01-21
LinkedCellList::build performance enhacements


j8asic:
**1. Extra deep copy**

In `LinkedCellList::build`, after computing offsets, the cell counters are reset [Kokkos::deep_copy( counts, 0 )](https://github.com/ECP-copa/Cabana/blob/d46804f0e002be4d9f7740fc94a34c4ff1fe6aa3/core/src/Cabana_LinkedCellList.hpp#L262)
But this is an overhead for the sake later usage of [Kokkos::atomic_fetch_add](https://github.com/ECP-copa/Cabana/blob/d46804f0e002be4d9f7740fc94a34c4ff1fe6aa3/core/src/Cabana_LinkedCellList.hpp#L271)
atomic_fetch_sub

Instead of the above, `counts` should not be reset, but instead it should be reused with the help of `Kokkos::atomic_fetch_sub`:
```
            int c = Kokkos::atomic_fetch_sub( &counts( cell_id ), 1 ) - 1;
            permute( offsets( cell_id ) + c ) = p;
```
EDIT: ignore this as I see below you need counts for BinningData

**2. Always allocating memory**

In unsteady simulations often construction of Views has significant impact. I advise `counts`, `offsets`, and `permute` to be member variables that are only resized when needed.

**3. Atomic exchange**

Linked lists can be built using atomic exchange [[1]](http://www.ce.uniroma2.it/publications/parco2013_uniformgrids.pdf), and maybe instead current scheme (atomic per particle + scan + atomic per particle), one can get (atomic + scan per linked list + permute per cell). But maybe this is not a great idea since the memory access would be more random.

Btw great project! I'll start making pull requests instead of issues asap when I get some time to contribute. 
Cheers,
Josip

sslattery:
@j8asic Thanks for the input! We really appreciate you checking out the code and providing feedback. I think you have some good ideas on performance here. We have a performance test that uses the linked cell list in the context of neighbor lists here: https://github.com/ECP-copa/Cabana/blob/master/core/performance_test/benchmark/md_neighbor_perf_test.cpp

We can use this to test your ideas. I think your memory allocation idea would be the easiest. Are you rebuilding the cell list every timestep or some subset of timesteps? Is this for molecular dynamics, PIC, or some other algorithm?

Thanks!

j8asic:
[...]

PR #373, opened 2021-03-04
LinkedCellList allocation

streeve:
Only allocate on create of the LinkedCellList, with the normal use case being create once, then rebuild (resizing if needed). Part of #335 

Also rearranges the unit test to reduce copied code. 

Will need to be rebased once #372 is merged. Using that performance test, I posted some results below

### DIFF
--- a/core/src/Cabana_LinkedCellList.hpp
+++ b/core/src/Cabana_LinkedCellList.hpp
@@ -35,6 +35,7 @@ class LinkedCellList
     using memory_space = typename device_type::memory_space;
     using execution_space = typename device_type::execution_space;
     using size_type = typename memory_space::size_type;
+    using CountView = Kokkos::View<int*, device_type>;
     using OffsetView = Kokkos::View<size_type*, device_type>;
 
     /*!
@@ -66,7 +67,9 @@ class LinkedCellList
                  grid_max[1], grid_max[2], grid_delta[0], grid_delta[1],
                  grid_delta[2] )
     {
-        build( positions, 0, positions.size() );
+        std::size_t np = positions.size();
+        allocate( totalBins(), np );
+        build( positions, 0, np );
     }
 
     /*!
@@ -98,6 +101,7 @@ class LinkedCellList
                  grid_max[1], grid_max[2], grid_delta[0], grid_delta[1],
                  grid_delta[2] )
     {
+        allocate( totalBins(), end - begin );
         build( positions, begin, end );
     }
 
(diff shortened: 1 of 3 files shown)


streeve:
Here's the performance comparison using #372:

```
Test 				Prev(ms) New(ms) Speedup(%)
serial_linkedcell_create_4  	 28.4  	 28.3  	 0.149
serial_linkedcell_sort_4  	 21.6  	 21.6  	 -0.0453
serial_linkedcell_create_8  	 28.2  	 28.1  	 0.178
serial_linkedcell_sort_8  	 21.6  	 21.6  	 -0.0598
cuda_linkedcell_create_4  	 0.586   0.157   73.2
cuda_linkedcell_sort_4 	 	 0.694   0.727   -4.71
cuda_linkedcell_create_8  	 0.713   0.23  	 67.7
cuda_linkedcell_sort_8  	 0.734   0.977   -33.1
```

Doesn't make a significant difference on the CPU. Getting a big speedup on the GPU creation, but I don't understand the permute slowdown. Also not sure why it's failing on jenkins - it's passing locally

sslattery:
You are seeing the Jenkins failure because you called a `parallel_for` in a constructor. It seems as though the compiler doesn't think it can take the address of the constructor. If you move the parallel implementation details in the construction into another function and call that function in the constructor it should work.

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -204,34 +208,45 @@ class LinkedCellList
...
-            ncell );
-        OffsetView offsets(
-            Kokkos::view_alloc( Kokkos::WithoutInitializing, "offsets" ),
-            ncell );
-        OffsetView permute(
-            Kokkos::view_alloc( Kokkos::WithoutInitializing, "permute" ),
-            end - begin );
+        if ( _counts.extent( 0 ) != ncell )
+        {
+            Kokkos::resize( _counts, ncell );
```
sslattery, line 231: I think this is fine for now because `Kokkos::resize` doesn't do anything if the grid size is the same. In the future if we have dynamic grid sizes we could keep a separate `size` and `capacity` here and only resize when we need more allocation.

sslattery (approved):
This looks good. I can't come up with a good reason for the slow-down on `permute` given that you haven't changed the data structure or the permutation algorithm.

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -204,34 +208,45 @@ class LinkedCellList
...
+            Kokkos::resize( _permute, nparticles );
+        }
 
-        // Get a local copy of the grid because it is class data and a lambda
-        // function will not capture it otherwise via CUDA.
+        // Get local copies of class data for lambda function capture.
         auto grid = _grid;
+        auto counts = _counts;
+        auto offsets = _offsets;
+        auto permute = _permute;
```
sslattery, line 244: actually this is aliasing a free function in the namespace

PR #378, opened 2021-03-17
Boundary Indices

sslattery:
Follow on to #374.

1. Adds `boundaryIndexSpace()` to the `LocalGrid` to allow for operations over indices in the boundary halo region for non-periodic boundaries.
2. Adds `inRange()` to the `IndexSpace` to check whether or not an index is within the range of the space. The will make implementing boundary conditions easier in some cases.

### DIFF
--- a/cajita/unit_test/tstLocalGrid.hpp
+++ b/cajita/unit_test/tstLocalGrid.hpp
@@ -327,13 +327,27 @@ void periodicTest3d()
 
     // Check that the boundary spaces are empty.
     auto boundary_cell_space =
-        local_grid->boundaryIndexSpace( Cell(), -1, 0, 1 );
+        local_grid->boundaryIndexSpace( Ghost{}, Cell(), -1, 0, 1 );
     EXPECT_EQ( 0, boundary_cell_space.size() );
 
-    boundary_cell_space = local_grid->boundaryIndexSpace( Cell(), 1, -1, 0 );
+    boundary_cell_space =
+        local_grid->boundaryIndexSpace( Ghost{}, Cell(), 1, -1, 0 );
     EXPECT_EQ( 0, boundary_cell_space.size() );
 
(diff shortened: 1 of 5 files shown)


cajita/unit_test/tstLocalGrid.hpp
```diff
@@ -1830,6 +1930,106 @@ void notPeriodicTest3d()
                 }
             }
 
+    // Check the boundary spaces.
+    IndexSpace<3> boundary_cell_space;
+    if ( -1 == local_grid->neighborRank( -1, 0, 1 ) )
+    {
+        boundary_cell_space =
+            local_grid->boundaryIndexSpace( Cell(), -1, 0, 1 );
+        EXPECT_EQ( boundary_cell_space.min( Dim::I ), 0 );
```
kwitaechong, line 1939: by the definition of `sharedIndexSpaceImpl` of off_ijk = -1,  owned_space.min( d ) - halo_width  = 0

kwitaechong:
I am wondering for the mixture case of periodic/non-periodic boundary condition, e.g., in 2D (True, False), how to compute boundaryIndexSpace( Cell(), 1, 1 )

kwitaechong (approved):

sslattery:
> I am wondering for the mixture case of periodic/non-periodic boundary condition, e.g., in 2D (True, False), how to compute boundaryIndexSpace( Cell(), 1, 1 ). In this case, it is a little confusing since upper-right block is periodic in x-dir but above y-non-periodic face.

Yes I thought about that a lot. If a dimensions is not periodic then *any* boundary region that includes that dimension has a non-zero sized `boundaryIndexSpace` because there is no way to decide which rank owns that particular region, even though it might be repeated on another rank. So to resolve this ambiguity we say that there is no unique owning rank and every rank gets its own copy

PR #383, opened 2021-03-31
Sparse Grid Partitioner

YuxingQiu:
Pull request for sparse grid partitioner; all local tests passed.

Part of #323


streeve:
This will need to be rebased on master to run CI and fix conflicts (switch to header only)

YuxingQiu:
I've just done the rebase stuff. I still have some debugging code not removed because I may need them for testing as my local computer cannot support tests with 12 MPI_ranks.

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,640 @@
...
+
+#include <array>
+#include <vector>
+
+#include <mpi.h>
+
+namespace Cajita
+{
+//---------------------------------------------------------------------------//
+template <typename MemorySpace, typename ExecSpace,
```
streeve, line 16: We should template on `DeviceType` instead, then extract the memory and execution space from it

streeve (changes requested):
Looking good! I'm wondering if we want this to be experimental for now

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,617 @@
...
+                    //     "compute_sub_workload to compute ave_workload - 0\n"
+                    //     );
+                    int j = static_cast<int>( jnk / rank_k );
+                    int k = static_cast<int>( jnk % rank_k );
+                    // printf(
+                    //     "in parallel_for, before calling "
+                    //     "compute_sub_workload to compute ave_workload - 1\n"
+                    //     );
+                    ave_workload( jnk ) =
+                        compute_sub_workload( dj, j, dk, k, rank, rec_partition,
```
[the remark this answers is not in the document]
YuxingQiu: `rank`, `rec_partition`, and `prefix_sum` are already local copies, otherwise, I will CUDA runtime errors; `ave_workload` is locally defined; other arguments are just numbers.

cajita/src/Cajita_SparseDimPartitioner.hpp
YuxingQiu: Sounds great. I'll change that.

cajita/src/Cajita_SparseDimPartitioner.hpp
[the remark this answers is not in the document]
streeve: Sorry, I should have been more clear - `compute_sub_workload` missing a local copy is the failure I see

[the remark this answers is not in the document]
YuxingQiu: Oh!! Well, even functions need local copies?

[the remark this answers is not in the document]
streeve: My mistake - I read too fast and thought this was a View.  Making it `public` will enable capture on the device

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+#include <gtest/gtest.h>
+
+#include <mpi.h>
+
+using namespace Cajita;
+
+namespace Test
+{
+void uniform_distribution_automatic_rank()
```
streeve, line 13: Can you add some more detail about the problem setup? Does the partitioning actually change?

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,293 @@
...
+        auto gt_tile = rec_partitions[d][cart_rank[d] + 1] -
+                       rec_partitions[d][cart_rank[d]];
+
+        EXPECT_EQ( owned_cells_per_dim[d], gt_tile * cell_per_tile_dim *
+                                               cell_per_tile_dim *
+                                               cell_per_tile_dim );
+    }
+}
+
+void biased_distribution_automatic_rank( std::array<int, 3> occupy_start,
```
streeve, line 143: How non-uniform is this test? I think I'll need to discuss with you to understand

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,293 @@
...
+    {
+        auto gt_tile = rec_partitions[d][cart_rank[d] + 1] -
+                       rec_partitions[d][cart_rank[d]];
+        EXPECT_EQ( owned_tiles_per_dim[d], gt_tile );
+        EXPECT_EQ( owned_cells_per_dim[d], gt_tile * cell_per_tile_dim *
+                                               cell_per_tile_dim *
+                                               cell_per_tile_dim );
+    }
+
+    // initialize sparseMap, register every tile on every MPI rank
```
streeve, line 101: You could share the sparsemap creation across tests

streeve (changes requested):

YuxingQiu:
I'm not sure how to fix the failed tests here. Both failed on the cases that are not related to SparsePartitioner.

streeve:
retest this please.

streeve:
> I'm not sure how to fix the failed tests here. Both failed on the cases that are not related to SparsePartitioner.

The Cuda failure was random, Intel failure I still need to fix

streeve (changes requested):
Couple quick things: 
- one warning in the CI to fix
- Needs to be added to `Cajita.hpp`

Couple bigger questions:
- Can you calculate the imbalance and check in the unit test?
- Should we have a unit test for a rebalancing a dense grid (either here or separate)?

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,369 @@
...
+    // check results (should be the same as the gt_partition)
+    auto part = partitioner.getCurrentPartition();
+    for ( int d = 0; d < 3; ++d )
+    {
+        for ( int id = 0; id < ranks_per_dim[d] + 1; id++ )
+            EXPECT_EQ( part[d][id], gt_partition[d][id] );
+    }
+
+    auto imbalance_factor = partitioner.computeImbalanceFactor( cart_comm );
+    EXPECT_EQ( imbalance_factor, 1.0f );
```
streeve, line 352: Can use `EXPECT_FLOAT_EQ` instead

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,687 @@
...
+    */
+    SparseDimPartitioner( MPI_Comm comm, float max_workload_coeff,
+                          int workload_num, int num_step_rebalance,
+                          const std::array<int, 3>& global_cells_per_dim,
+                          int max_optimize_iteration = 10 )
+        : _workload_threshold(
+              static_cast<int>( max_workload_coeff * workload_num ) )
+        , _num_step_rebalance( num_step_rebalance )
+        , _max_optimize_iteration( max_optimize_iteration )
+        , _workload_per_tile(
```
streeve, line 69: Could write an `allocate` function and call in both constructors

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,687 @@
...
+        int size_1 = prefix_sum.extent( 1 );
+        int size_2 = prefix_sum.extent( 2 );
+
+        int rank_0 = _ranks_per_dim[0];
+        int rank_1 = _ranks_per_dim[1];
+        int rank_2 = _ranks_per_dim[2];
+
+        Kokkos::parallel_for(
+            "compute_workload_record",
+            Kokkos::RangePolicy<execution_space>( 0, 2 ),
```
streeve, line 592: Probably not worth a parallel kernel here. Could copy what you need to the host instead

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,687 @@
...
+            Kokkos::deep_copy( _rectangle_partition_dev, rec_mirror );
+        } // end for (3 dimensions)
+    }
+
+    /*!
+      \brief compute the imbalance factor for the current partition
+      \param cart_comm MPI cartesian communicator
+      \return the imbalance factor = workload on current rank / ave_workload
+    */
+    float computeImbalanceFactor( MPI_Comm cart_comm )
```
streeve, line 564: I think we also want `computeImbalance` that just returns the current workload and `computeTotalImbalanceFactor` which sums the imbalance across ranks. That would be a good way to trigger a reoptimization step

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,687 @@
...
+
+    /*!
+      \brief compute the workload in the current MPI rank from particle
+      positions (each particle count for 1 workload value)
+      \param view particle positions view
+      \param particle_num total particle number
+      \param dx cell dx size
+    */
+    template <class ParticlePosViewType, typename CellUnit>
+    void computeLocalWorkLoad( ParticlePosViewType& view, int particle_num,
```
streeve, line 272: I don't think we're testing the particle workload codepath - is that relatively easy to add? Or should we leave that until later (if so maybe we remove this for now)?

streeve (changes requested):
This looks great! Few more small things, but I think it's almost ready to merge

YuxingQiu:
Could you please help retest the code? @streeve 
Lots of unrelated HIP tests failed. T^T 
Thank you so much!

streeve (commented):
As we discussed at the meeting let's add a test for the particle rebalancing and then this should be ready to merge

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,369 @@
...
+    for ( int d = 0; d < 3; ++d )
+    {
+        auto gt_tile = rec_partitions[d][cart_rank[d] + 1] -
+                       rec_partitions[d][cart_rank[d]];
+
+        EXPECT_EQ( owned_cells_per_dim[d], gt_tile * cell_per_tile_dim *
+                                               cell_per_tile_dim *
+                                               cell_per_tile_dim );
+    }
+}
```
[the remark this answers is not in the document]
YuxingQiu: I've added the imbalance factor test (after optimization) in `uniform_distribution_test`, with a manually computed ground-truth number. In this `uniform_distribution_test`, the imbalance_factor is not exactly 1.0f but depends on how many MPI_ranks we are using. (e.g. domain_size_in_dim_i = 32, rank_num_in_dim_i = 3, then the number of tiles is not perfectly averaged on each MPI_rank, so the imbalance factor in this case is not 1.0f).

I don't think we need to test the imbalance factor before optimization. On the one hand, the accuracy is ensured by the way we test the optimization results (ensuring the optimized output being the same as the initialized partition), which means we will get the same results calling the same interface to compute the imbalance factor. On the other hand, if we want to compute the imbalance factor before optimization, we have to manually compute the workload matrix and the prefix_sum matrix before computing the imbalanced factor, which is pretty time-consuming.

streeve (commented):
Looks good! Just one small thing, then a few places that need updated comments

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -256,22 +256,32 @@ class SparseDimPartitioner : public BlockPartitioner<3>
       \param particle_num total particle number
       \param dx cell dx size
     */
-    template <class ParticlePosViewType, typename CellUnit>
-    void computeLocalWorkLoad( ParticlePosViewType& view, int particle_num,
-                               CellUnit dx )
+    template <class ParticlePosViewType, typename ArrayType, typename CellUnit>
+    void computeLocalWorkLoad( const ParticlePosViewType& view,
```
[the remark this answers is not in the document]
YuxingQiu: Oh okay! sorry I always forget about the doc thing... T_T

streeve (approved):

streeve:
Ok, @YuxingQiu if you rebase on master everything should pass

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,473 @@
...
+        int ele = size_tile_per_dim / ranks_per_dim[d];
+        int part = 0;
+        for ( int i = 0; i < ranks_per_dim[d]; ++i )
+        {
+            rec_partitions[d].push_back( part );
+            part += ele;
+        }
+        rec_partitions[d].push_back( size_tile_per_dim );
+    }
+    partitioner.initializeRecPartition( rec_partitions[0], rec_partitions[1],
```
sslattery, line 82: Can the partitioning interface allocate the vectors inside of this array? It would prevent users needing to figure this out with a call to `ranksPerDimension`

cajita/unit_test/tstSparseDimPartitioner.hpp
```diff
@@ -0,0 +1,473 @@
...
+        Kokkos::parallel_for(
+            "insert_tile_to_sparse_map",
+            Kokkos::RangePolicy<TEST_EXECSPACE>( 0, tiles_view.extent( 0 ) ),
+            KOKKOS_LAMBDA( int id ) {
+                sis.insertTile( tiles_view( id, 0 ), tiles_view( id, 1 ),
+                                tiles_view( id, 2 ) );
+            } );
+
+        // compute workload from a sparseMap and do partition optimization
+        partitioner.optimizePartition( sis, MPI_COMM_WORLD );
```
sslattery, line 430: Does initialize have to be called before optimize? If so and we always want to optimize, can we create a version of these functions that fuses them together?

sslattery (approved):
Just a few small questions/comments and then we can merge

cajita/unit_test/tstSparseDimPartitioner.hpp
YuxingQiu: So the original design is like, the user may probably want to set up the ranks_per_dim manually and initialize the partition in some special way (not just the average way I used in the unit tests). 
But it's true here if the user wants to leave everything to us, this call would be tedious. So I think we can call this `initializeRecPartition` in the constructor to create an average partitioner as initialization. In this way, both `ranksPerDimension`, as well as the `initializeRecPartition`, are hidden from the user. 
How do like it if I do this before merging?

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -0,0 +1,752 @@
...
+      dimention
+      \param max_workload_coeff threshold factor for re-partition
+      \param workload_num total workload(particle/tile) number, used to compute
+      workload_threshold
+      \param num_step_rebalance the simulation step number after which one
+      should check if repartition is needed
+      \param global_cells_per_dim 3D array, global cells in each dimension
+      \param max_optimize_iteration max iteration number to run the optimization
+    */
+    SparseDimPartitioner( MPI_Comm comm, float max_workload_coeff,
```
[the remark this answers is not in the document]
YuxingQiu: Yeah sounds great! So for now even though I store the `max_workload_coeff` here, I don't really use this information. But I think I will make use of that in the future to create an automatic check for when or whether to do the optimizations. So we can set up default values for those parameters later based on our testings.

cajita/unit_test/tstSparseDimPartitioner.hpp
YuxingQiu: Only one time init before simulations start, and multiple times of optimization during the simulation. 
When optimizing, we will fix two dimensions and optimize the rest, so the partition information is needed in those two dimensions. That's why we do initialization before optimization, but only for once, as the results of the previous optimization can serve as the init state for the next round of optimization.

cajita/unit_test/tstSparseDimPartitioner.hpp
sslattery: Let's get this merged as-is and then do another pull request where we make API updates

PR #385, opened 2021-04-21
Add hypre to GHA

streeve:
Part of #345


sslattery (approved):

streeve:
@sslattery still fails with hypre `v2.19.0`

sslattery:
> @sslattery still fails with hypre `v2.19.0`

Yeah I just saw that :( I will need to look at that

streeve:
@sslattery this is passing with 2.18.2 (failure is unrelated)

PR #395, opened 2021-05-20
Particle-grid migrate

streeve:
This adds particle-grid migrate in Cajita, wrapping calls to `Cabana::Distributor`, `Cabana::migrate`

 - [x] Implementation (stolen from ExaMPM) determines new ranks destinations and wraps through periodic boundaries
 - [x] `particleGridMigrate` uses implementation functions and calls `migrate`
 - [x] 2d and 3d tests

This supersedes #263 and is part of the very old #77


streeve:
Ok @sslattery this is ready for review. I will keep working on the similar arbitrary dimension halo functionality and create a separate PR

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+        // Check within the halo width, within the ghosted domain.
+        const auto& local_mesh =
+            Cajita::createLocalMesh<Kokkos::HostSpace>( _local_grid );
+
+        Kokkos::Array<double, num_space_dim> local_low{};
+        Kokkos::Array<double, num_space_dim> local_high{};
+        for ( std::size_t d = 0; d < num_space_dim; ++d )
+        {
+            auto dx = _local_grid.globalGrid().globalMesh().cellSize( d );
```
sslattery, line 66: This implies that a uniform mesh is being used. In hindsight, given our curvilinear grid development elsewhere I wish that Cajita only did logically uniform grids but we need to support the other grids here. For now you could simply do a static assert that the user gave you a `LocalGrid` derived form a `UniformMesh`

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+        Kokkos::Array<double, num_space_dim> local_low{};
+        Kokkos::Array<double, num_space_dim> local_high{};
+        Kokkos::Array<bool, num_space_dim> periodic{};
+        Kokkos::Array<double, num_space_dim> global_low{};
+        Kokkos::Array<double, num_space_dim> global_high{};
+        Kokkos::Array<double, num_space_dim> global_extent{};
+
+        for ( std::size_t d = 0; d < num_space_dim; ++d )
+        {
+            local_low[d] = local_mesh.lowCorner( Cajita::Own(), d );
```
sslattery, line 150: We do this in a lot of places. No need to add in this PR but it would be nice to have functions for these that do not take a dimension argument and just return a `Kokkos::Array` for all dimensions.

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+                for ( std::size_t d = 0; d < num_space_dim; ++d )
+                {
+                    nid[d] = 1;
+                    if ( positions( p, d ) < local_low[d] )
+                        nid[d] = 0;
+                    else if ( positions( p, d ) > local_high[d] )
+                        nid[d] = 2;
+                }
+
+                // Compute the destination MPI rank [ni + 3*(nj + 3*nk) in 3d].
```
sslattery, line 173: this should also work in 2d right?

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+                {
+                    int npower = 1;
+                    for ( std::size_t dp = 1; dp <= d; ++dp )
+                        npower *= 3;
+                    neighbor_index += npower * nid[d];
+                }
+                destinations( p ) = neighbor_ranks( neighbor_index );
+
+                // Shift particles through periodic boundaries.
+                for ( std::size_t d = 0; d < num_space_dim; ++d )
```
sslattery, line 185: nice - good kernel fusion. I think we can also eventually find a way to fuse in the count kernel as well so we have only a single setup kernel.

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+//---------------------------------------------------------------------------//
+// Particle Grid Distributor
+//---------------------------------------------------------------------------//
+/*!
+  \brief Determine which particles need to be migrated using the grid
+
+  \tparam LocalGridType Cajita LocalGrid type.
+*/
+template <class LocalGridType>
+class ParticleGridDistributor
```
sslattery, line 39: Does this even need to be a class? Based on your implementations below the content of these could just be in those free functions.

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+  \param minimum_halo_width Number of halo mesh widths to include for
+  ghosting.
+*/
+template <class LocalGridType, class PositionSliceType>
+int migrateCount( const LocalGridType& local_grid,
+                  const PositionSliceType& positions,
+                  const int minimum_halo_width )
+{
+    ParticleGridDistributor<LocalGridType> pg_distributor( local_grid );
```
sslattery, line 223: I don't think `ParticleGridDistributor` needs to be an object - the code in `migrateCount` can just be added here instead as a free function. You should have some kind of `ParticleGridDistributor` namespace though for these free functions.

cajita/src/Cajita_ParticleGridDistributor.hpp
```diff
@@ -0,0 +1,390 @@
...
+        topology_host( topology.data(), topology.size() );
+    auto topology_mirror =
+        Kokkos::create_mirror_view_and_copy( device_type(), topology_host );
+    Kokkos::View<int*, device_type> destinations(
+        Kokkos::ViewAllocateWithoutInitializing( "destinations" ),
+        positions.size() );
+
+    // Determine destination ranks for all particles and wrap positions across
+    // periodic boundaries.
+    pg_distributor.getMigrateDestinations( topology_mirror, destinations,
```
sslattery, line 267: Same here - could just inject code into free function instead of having a class.

cajita/src/Cajita_ParticleGridDistributor.hpp
sslattery: Or instead of a namespace prefix the functions with `particleGridDistributor` as you have below.

cajita/unit_test/tstParticleGridDistributor2d.hpp
```diff
@@ -0,0 +1,364 @@
...
+        global_low_corner, global_high_corner, global_num_cell );
+    auto global_grid = Cajita::createGlobalGrid( MPI_COMM_WORLD, global_mesh,
+                                                 is_periodic, partitioner );
+    return global_grid;
+}
+
+//---------------------------------------------------------------------------//
+// RUN TESTS
+//---------------------------------------------------------------------------//
+TEST( TEST_CATEGORY, not_periodic_test )
```
sslattery, line 270: It seems like a lot of the code in the `not_periodic_test` and `periodic_test` could be combined and only a few boolean options needed to delineate them.

cajita/unit_test/tstParticleGridDistributor3d.hpp
```diff
@@ -0,0 +1,406 @@
...
+        global_low_corner, global_high_corner, global_num_cell );
+    auto global_grid = Cajita::createGlobalGrid( MPI_COMM_WORLD, global_mesh,
+                                                 is_periodic, partitioner );
+    return global_grid;
+}
+
+//---------------------------------------------------------------------------//
+// RUN TESTS
+//---------------------------------------------------------------------------//
+TEST( TEST_CATEGORY, not_periodic_test )
```
sslattery, line 288: Same here with overlap - these two could probably be combined mostly as a single function (still want separate functions at GTest level though for delineation)

sslattery (changes requested):
Just a few small changes

cajita/src/Cajita_ParticleGridDistributor.hpp
streeve: I converted this to a class to make the variable dimension easier. Is there a good way to do that otherwise? I'm okay with moving this to `Impl` if we keep it

cajita/src/Cajita_ParticleGridDistributor.hpp
streeve: I should definitely have commented to make this free function wrapper more clear, but this was again just to use the mesh spatial dimension without the user having to worry about it

cajita/src/Cajita_ParticleGridDistributor.hpp
streeve: Yes, I just left the comment because the code isn't as readable for 2d & 3d

cajita/unit_test/tstParticleGridDistributor2d.hpp
streeve: Fixed

cajita/unit_test/tstParticleGridDistributor3d.hpp
streeve: Fixed

cajita/src/Cajita_ParticleGridDistributor.hpp
streeve: Added, but I could add a non-uniform version later

cajita/src/Cajita_ParticleGridDistributor.hpp
streeve: I fixed this as you suggested

cajita/src/Cajita_ParticleGridDistributor.hpp
streeve: Removed

streeve:
@sslattery should be ready for final review

sslattery (approved):

PR #396, opened 2021-05-20
GHA: reenable doxygen upload

junghans:
Last item from #345

PR #394, opened 2021-05-20
Jenkins: update to rocm 4.2 and restrict HIP build to MI60

streeve:

PR #399, opened 2021-05-20
Fixup doxygen

streeve:
Currently errors if nothing new to commit to doxygen

Follow up to #396


streeve:
@junghans I tried to fix here, but we need a clean exit if there's nothing to commit for doxygen (#394 fails that way). Not sure why this gives a permissions error

junghans (approved):

PR #400, opened 2021-05-22
GHA: Treat Doxygen warnings as error

junghans:


.github/workflows/CI2.yml
```diff
@@ -290,7 +290,9 @@ jobs:
             -DCabana_ENABLE_PERFORMANCE_TESTING=ON \
             -DCabana_PERFORMANCE_EXPECTED_FLOPS=0 \
             -DCabana_ENABLE_COVERAGE_BUILD=${{ matrix.coverage }} \
-            -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }}
+            -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+            -DDOXYGEN_WARN_AS_ERROR=YES \
+            -DDOXYGEN_WARN_IF_UNDOCUMENTED=NO
```
junghans, line 295: Wouldn't it be better to document the missing items?

sslattery: Does this logic imply that the doxygen build will error out if something is undocumented? Also will this cause the CI to fail?

streeve: Is there a good way to only warn for things we intend to document (e.g. leave out some impl and explicit template instantiations)?

streeve: @sslattery with `WARN_IF_UNDOCUMENTED=NO`, `WARN_AS_ERROR` won't cause CI to fail if an entire class/member function is undocumented, but it will if it's partially documented. One other issue I fixed here was that it choked on the latex expression I escaped

streeve:
Rebased to check for errors in free functions. As discussed, we'll allow undocumented items for now and remove the `WARN_IF_UNDOCUMENTED` flag when documentation is updated in following PRs

junghans:
@streeve Great work!

@sslattery Ready to merge, please review.

sslattery (approved):

streeve (approved):

PR #402, opened 2021-05-25
Add atomic update for neighbor example

### LINKED ISSUE #401: Atomic access required in neighbor_parallel_for_example.cpp?
I am currently working through the tutorials. They are very clear and the code base is really nice. :+1: 

While reading through https://github.com/ECP-copa/Cabana/blob/master/example/tutorial/10_neighbor_parallel_for/neighbor_parallel_for_example.cpp I stumbled over a potential race condition. For `Cabana::TeamOpTag()` parallelization is done over particles and neighbors, right? This should result in a race condition here https://github.com/ECP-copa/Cabana/blob/master/example/tutorial/10_neighbor_parallel_for/neighbor_parallel_for_example.cpp#L124-L127, since access to `slice_i( i )` is not atomic. Did I miss something?

streeve:
Fixes #401 - thanks to @XzzX for pointing this out!


sslattery (approved):

junghans:
The IC failure is #398 again.

PR #406, opened 2021-06-01
Fix VerletList memory space

### LINKED ISSUE #405: something wrong with the first template parameter of VerletList
According to the documentation the first template parameter of `VerletList` should be a `DeviceType`
https://github.com/ECP-copa/Cabana/blob/d6505da281b23344952aec075905f21f759f5a8c/core/src/Cabana_VerletList.hpp#L611-L630

However, the example uses a `MemoryType`
https://github.com/ECP-copa/Cabana/blob/d6505da281b23344952aec075905f21f759f5a8c/example/core_tutorial/10_neighbor_parallel_for/neighbor_parallel_for_example.cpp#L112

and in fact passing a `DeviceType` will result in a `static_assert` during `neighbor_parallel_for`.

streeve:
`VerletList` used `DeviceType` as a template parameter which was inconsistent with the `NeighborList` bridge pattern. 

This PR: 
- Simplifies the `VerletListData` (which does not need an `ExecutionSpace`)
- converts `VerletList` to use `MemorySpace`
- adds a `VerletList` `build` function which can use a separate (accessible) `ExecutionSpace` which defaults to the Kokkos `MemorySpace` default

Fixes #405

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -37,23 +37,25 @@ struct VerletLayout2D
 //---------------------------------------------------------------------------//
 // Verlet List Data.
 //---------------------------------------------------------------------------//
-template <class DeviceType, class LayoutTag>
+template <class MemorySpace, class LayoutTag>
 struct VerletListData;
 
-template <class DeviceType>
-struct VerletListData<DeviceType, VerletLayoutCSR>
+template <class MemorySpace>
+struct VerletListData<MemorySpace, VerletLayoutCSR>
 {
-    // Device type.
-    using device_type = DeviceType;
+    // Kokkos memory space.
+    using memory_space = MemorySpace;
+    // Default Kokkos device type.
+    using device_type [[deprecated]] = typename memory_space::device_type;
 
     // Number of neighbors per particle.
-    Kokkos::View<int*, device_type> counts;
+    Kokkos::View<int*, memory_space> counts;
 
     // Offsets into the neighbor list.
-    Kokkos::View<int*, device_type> offsets;
+    Kokkos::View<int*, memory_space> offsets;
 
     // Neighbor list.
-    Kokkos::View<int*, device_type> neighbors;
+    Kokkos::View<int*, memory_space> neighbors;
 
     // Add a neighbor to the list.
     KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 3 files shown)


core/src/Cabana_VerletList.hpp
```diff
@@ -37,23 +37,23 @@ struct VerletLayout2D
...
 struct VerletListData;
 
-template <class DeviceType>
-struct VerletListData<DeviceType, VerletLayoutCSR>
+template <class MemorySpace>
+struct VerletListData<MemorySpace, VerletLayoutCSR>
 {
     // Device type.
-    using device_type = DeviceType;
+    using memory_space = MemorySpace;
```
dalg24, line 47: `device_type` was a public type member so this may break user code (just making sure you are aware of this).

You could do
```C++
using device_type [[deprecated]] = typename MemorySpace::device_type;
```
to be safe.

dalg24: Also you might want to do
```C++
static_assert(Kokkos::is_memory_space<MemorySpace>::value, "");

Comment (streeve) [reply core/src/Cabana_VerletList.hpp:47]: Good point - I re-added the `device_type` everywhere and added the `assert` below in the main class

**Comment by streeve** [comment]
@sslattery ready for review

File: core/src/Cabana_VerletList.hpp [hunk 2]
```diff
@@ -694,9 +699,46 @@ class VerletList
...
+    template <class PositionSlice, class ExecutionSpace>
+    void build( ExecutionSpace, PositionSlice x, const std::size_t begin,
+                const std::size_t end,
+                const typename PositionSlice::value_type neighborhood_radius,
+                const typename PositionSlice::value_type cell_size_ratio,
+                const typename PositionSlice::value_type grid_min[3],
+                const typename PositionSlice::value_type grid_max[3],
+                const std::size_t max_neigh = 0 )
+    {
+        static_assert( is_accessible_from<memory_space, ExecutionSpace>{}, "" );
```
sslattery, line 735: Should probably have a simple failure message here.

sslattery (approved):

core/src/Cabana_VerletList.hpp
streeve: Maybe as a separate PR? We do this in a number of places

PR #409, opened 2021-06-02
fix allocation/bounds error in sparse dim partitioner

sslattery:
A sanitizer build indicated an out of bounds access in the new partitioner code.

### DIFF
--- a/cajita/src/Cajita_SparseDimPartitioner.hpp
+++ b/cajita/src/Cajita_SparseDimPartitioner.hpp
@@ -230,7 +230,7 @@ class SparseDimPartitioner : public BlockPartitioner<3>
             Kokkos::HostSpace(), _rectangle_partition_dev );
         for ( int d = 0; d < 3; ++d )
         {
-            rec_part[d].resize( _ranks_per_dim[d] );
+            rec_part[d].resize( _ranks_per_dim[d] + 1 );
             for ( int id = 0; id < _ranks_per_dim[d] + 1; ++id )
             {
                 rec_part[d][id] = rec_mirror( id, d );


cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -231,7 +231,7 @@ class SparseDimPartitioner : public BlockPartitioner<3>
         for ( int d = 0; d < 3; ++d )
         {
             rec_part[d].resize( _ranks_per_dim[d] );
```
sslattery, line 233: @YuxingQiu I interpreted the size of this allocation to be correct. Therefore all of the `+ 1` operations in the loop upper bounds were creating data access of out of bounds. Can you review to ensure this was your intent?

YuxingQiu: Oh, looks like I didn't make it clear in the doc. But here we should change Line 233 to `rec_part[d].resize(_ranks_per_dim[d]+1)` without making other changes.

The size of `_rectangle_partition_dev` is `(ranks_per_dim[0]+1, ranks_per_dim[1]+1, ranks_per_dim[2]+1), with the first element equals to `0` and the last one equals to `ranks_per_dim[d]`. The first and the last element can make more general indexing in compute_local_workload. As you can see in `tstSparseDimPartitioner.hpp` line 80, there's an extra element pushed after `ranks_per_dim[d]` loops. Also in the allocation, the size is `max_size+1` (`Cajita_SparseIndexSpace.hpp` Line 207).

sslattery: OK I have updated - @YuxingQiu can you take another look?

YuxingQiu (approved):
Looks great!

streeve:
Retest this please

PR #410, opened 2021-06-02
Improve doxygen documentation

streeve:
Update doxygen to cover the full library. Conditionally exclude some implementation details with `\cond`

~~This is still WIP based on the total amount of missing doxygen, documentation, and the best way to reduce repeated comments~~


sslattery (approved):

sslattery:
@streeve This looks good - do you want to merge?

streeve:
@sslattery this should be finished and ready for review

PR #416, opened 2021-06-10
Sanitizer/Valgrind CI

### LINKED ISSUE #412: CI memory checks
GHA additions:
- [ ] Address/leak sanitizer
- [ ] valgrind

junghans:
First part of #412


junghans:
@sslattery vlagrind works and found a couple of things!

Memory and address sanitizer still needs to enabled a a build.

CMakeLists.txt
```diff
@@ -152,6 +152,26 @@ if(Cabana_ENABLE_COVERAGE_BUILD)
...
+endif()
+if(WITH_ASAN)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -fno-omit-frame-pointer")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=address")
+endif()
+
+if(WITH_MSAN)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=memory -fno-omit-frame-pointer")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=memory")
+endif()
```
dalg24, line 173: Why an option when all you need is CXX flags?

junghans: Only clang support `-fsanitize=memory` and some obscure compilers might not have `-fsanitize=address`.

junghans:
@sslattery @streeve I don't really like the fact that we are overwriting the existing tests, I think it would be better to create a second test for valgrind instead.

Any opinions?

junghans:
@streeve should be use enable `-fsanitize=address` all the time and add one build for `-fsanitize=memory`, or what do you think?

streeve:
So `-fsanitize=address` all the time and one additional `-fsanitize=memory` and one valgrind? 

That sounds good to me

streeve:
Couple of failures for the memory sanitizer to fix, but the address sanitizer looks good

streeve:
Okay @junghans this is passing with only address santizer @sslattery anything you want to change for these newoptions?

sslattery (approved):

PR #417, opened 2021-06-11
Only run GitHub Actions on main repo

streeve:
Exclude forks from running CI

### DIFF
--- a/.github/workflows/CI2.yml
+++ b/.github/workflows/CI2.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   CI:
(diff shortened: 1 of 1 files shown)


sslattery (commented):
Does this improve our turnaround times?

streeve:
> Does this improve our turnaround times?

Not sure. I just assumed we didn't need to run CI on forks

junghans:
Why do you want to do that? Testing commits on a fork before making a PR seems like a reasonable use case.

streeve:
> Why do you want to do that? Testing commits on a fork before making a PR seems like a reasonable use case.

Fair enough, I'll close

PR #419, opened 2021-06-17
Add neighbor_parallel functor tests

### LINKED ISSUE #414: Missing work_tag support for neighbor_parallel_for and neighbor_parallel_reduce
`Cabana::neighbor_parallel_for` and `Cabana::neighbor_parallel_reduce` do not support `function name tags`. 

The internal lambda has no `work_tag`
https://github.com/ECP-copa/Cabana/blob/575a692f1212a6be4059023ea75a81c25852f5c0/core/src/Cabana_Parallel.hpp#L243

However, this lambda is called via `Kokkos::parallel_for` which passes the `work_tag` as the first argument.
https://github.com/ECP-copa/Cabana/blob/575a692f1212a6be4059023ea75a81c25852f5c0/core/src/Cabana_Parallel.hpp#L253

streeve:
All previous neighbor parallel tests used lambdas. Our implementation works fine for team parallelism, but the explicit indexing pointed out by @XzzX breaks serial with a work tag

Fixes #414 
First part of #422

### DIFF
--- a/core/src/Cabana_Parallel.hpp
+++ b/core/src/Cabana_Parallel.hpp
@@ -237,6 +237,11 @@ inline void neighbor_parallel_for(
 
     using memory_space = typename neighbor_list_traits::memory_space;
 
+    auto begin = exec_policy.begin();
+    auto end = exec_policy.end();
+    using linear_policy_type = Kokkos::RangePolicy<execution_space, void, void>;
+    linear_policy_type linear_exec_policy( begin, end );
+
     static_assert( is_accessible_from<memory_space, execution_space>{}, "" );
 
     auto neigh_func = KOKKOS_LAMBDA( const index_type i )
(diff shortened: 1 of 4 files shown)


sslattery (approved):

XzzX:
Do you want to add a test for `neighbor_parallel_reduce` with work tags as well?

streeve:
> Do you want to add a test for `neighbor_parallel_reduce` with work tags as well?

Good point, I'll add

core/src/Cabana_Parallel.hpp
```diff
@@ -292,6 +297,11 @@ inline void neighbor_parallel_for(
 
     using memory_space = typename neighbor_list_traits::memory_space;
 
+    auto begin = exec_policy.begin();
+    auto end = exec_policy.end();
+    using linear_policy_type = Kokkos::RangePolicy<execution_space, void, void>;
```
dalg24, line 302: Sorry I know it is a bit late but what are these two trailing `void` template parameter there for?  Does the compile with Kokkos 3.4?

streeve: Yeah those are unnecessary. It does pass with Kokkos 3.2, master, and develop though

PR #426, opened 2021-06-29
Use new HYPRE cmake and add accelerator support

### LINKED ISSUE #388: HYPRE on GPU
Add support for calling HYPRE on the GPU either with GPU data or in an offload-mode with host data.

sslattery:
The pull request does the following:

1. Updates HYPRE to version 2.22.0
2. Uses the built-in HYPRE CMake with `find_package` instead of our own `FindHYPRE` file
3. Updates the HYPRE solver interface to use `MemorySpace` instead of the `DeviceType` template parameter to specify memory storage type. 
4. Adds accelerator support for HYPRE. The solver now checks that the memory space provided to the interface is compatible with HYPRE. I tested this on the Summit GPUs. No AMD testing yet.

NOTE: To get this to work I had to add `C` as one of the fundamental Cabana languages. The reason for this was that our `find_package( MPI )` was not finding the `MPI::MPI_C` component of the target which is a downstream requirement of HYPRE. HYPRE is setting that requirement in its CMake but perhaps we are overwriting this with our own `find_package(MPI)` call? I did attempt `find_package(MPI COMPONENTS MPI::MPI_C REQUIRED)` but that was not successful in finding `MPI::MPI_C` in our configuration until I added to `C` language support.

Closes #388


streeve:
Of course hypre tests pass on Cuda, but fail with serial...

sslattery:
I think we should either wait to merge this until https://github.com/hypre-space/hypre/issues/412 is resolved as this is the current source of our failure on Jenkins or attempt to turn on/off tests based on the HYPRE configuration.

sslattery:
OK for now I reworked logic so only one HYPRE backend is possible and only the tests for the backend will actually execute. If a user attempts to use a backend incompatible with the configured HYPRE then it will not compile. We can update later as needed if a run-time option for switching backends in HYPRE is provided.

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 #------------------------------------------------------------------------------#
 cmake_minimum_required(VERSION 3.9)
 
-project(Cabana LANGUAGES CXX)
+project(Cabana LANGUAGES CXX C)
```
junghans, line 17: Why do we need to enable `C`?

sslattery: See the PR description

junghans: Thanks maybe add a explicit `enable_language(C)` with a comment, where CMake looks for Hypre.

junghans: I made a PR to fix this in future hypre version https://github.com/hypre-space/hypre/pull/423

So let's use `enable_language(C)` below and add a comment to drop this in the future.

CMakeLists.txt
```diff
@@ -73,22 +73,24 @@ macro(Cabana_add_dependency)
   set(Cabana_ENABLE_${CABANA_DEPENDENCY_OPTION} ${${CABANA_DEPENDENCY_PACKAGE}_FOUND})
 endmacro()
 
+# FIXME: HYPRE workaround. Hypre is currently not exporting the find_package(
+# MPI ) to find the MPI::MPI_C targets and therefore we add this here for now
+# to ensure those targets are found for HYPRE.
+if ( Cabana_REQUIRE_HYPRE )
```
streeve, line 79: Should be ENABLE?

junghans: Add a link to  hypre-space/hypre#423 in the comment.

cajita/src/Cajita_HypreStructuredSolver.hpp
```diff
@@ -32,22 +33,78 @@
...
+//! Hypre device compatibility check.
+template <class MemorySpace>
+struct HypreIsCompatibleWithMemorySpace : std::false_type
+{
+};
+
+// FIXME: This is currently written in this structure because HYPRE only has
+// compile-time switches for backends and hence only one can be used at a
+// time. Once they have a run-time switch we can use that instead.
+#if defined HYPRE_USING_CUDA
```
streeve, line 55: Apparently it's `HYPRE_WITH_CUDA`

cajita/src/Cajita_HypreStructuredSolver.hpp
```diff
@@ -32,22 +33,78 @@
...
+};
+
+//! Hypre device compatibility check - CUDA UVM memory.
+template <>
+struct HypreIsCompatibleWithMemorySpace<Kokkos::CudaUVMSpace> : std::true_type
+{
+};
+#endif // end KOKKOS_ENABLE_CUDA
+
+#elif defined HYPRE_USING_HIP
```
streeve, line 71: Presumably `HYPRE_WITH_HIP`

cajita/src/Cajita_HypreStructuredSolver.hpp
sslattery: `HYPRE_config.h` has `HYPRE_USING_CUDA` and `HYPRE_USING_HIP`

CMakeLists.txt
sslattery: Fixed

CMakeLists.txt
sslattery: I just enabled it directly for now

junghans:
Problem: hypre spackage is building with autotools (not cmake) hence native `find_package(HYPRE)` won't work due to missing `HYPRE_config.h.cmake`.

see https://github.com/spack/spack/pull/24445#issuecomment-872535036

cajita/src/Cajita_HypreStructuredSolver.hpp
```diff
@@ -36,18 +37,72 @@
...
 //! Hypre structured solver interface for scalar fields.
-template <class Scalar, class EntityType, class DeviceType>
+template <class Scalar, class EntityType, class MemorySpace>
 class HypreStructuredSolver
 {
   public:
     //! Entity type.
     using entity_type = EntityType;
-    //! Kokkos device type.
-    using device_type = DeviceType;
```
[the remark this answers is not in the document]
sslattery: Good idea

sslattery:
> serial

How is that even compiling with the `enable_if` statements in the test? If the memory space isn't compatible it shouldn't even see the class get instantiated with a bad memory space.

streeve:
retest this please.

streeve (approved):

PR #432, opened 2021-07-13
CI: fix docker target repo

junghans:
Sorry, I wasn't able to test the final push!


streeve (approved):
I just noticed that the hypre version is out of date with master (only relevant for hypre with cuda I think)

junghans:
> I just noticed that the hypre version is out of date with master (only relevant for hypre with cuda I think)

Good point, let's make it consistent, bumped.

PR #433, opened 2021-07-14
CI: move sources into src/ in Docker

junghans:
CMake sporadically picks up the cmake files from sources (e.g. heffte) in `$HOME` as `$HOME/bin` is part of `$PATH`, which CMake searches automatically. Hence move all the sources to `$HOME/src`.

See https://gitlab.kitware.com/ecp/cmake-support/-/issues/16 for the full story.


streeve (approved):
I don't have access to that kitware repo, but this seems like a reasonable change to avoid cmake issues

PR #439, opened 2021-07-22
Preparation to include non equi block partitioning by liball.

aetx:
Very rough changes for now that are required by the corresponding `all_lb` branch of ExaMPM.

At the moment it allows one to get a mutable version of the global grid from the local grid. The global grid also allows to change the global cell offset and number of owned cells. Basically allowing to change the position and extent of the local grid.

(This also needs clean up, since not all changes are required and still remnants of tests)

### DIFF
--- a/cajita/src/Cajita_GlobalGrid.hpp
+++ b/cajita/src/Cajita_GlobalGrid.hpp
@@ -160,7 +160,7 @@ class GlobalGrid
     //! \brief Set the owned number of cells in a given dimension of this block.
     //! \param dim Spatial dimension.
     //! \param num_cell New number of owned cells.
-    void setOwnedNumCell( const int dim, const int num_cell );
+    void setOwnedNumCell( const std::array<int, num_space_dim>& num_cell );
 
     //! \brief Get the global offset in a given dimension. This is where our
     //! block starts in the global indexing scheme.
--- a/cajita/src/Cajita_GlobalGrid.hpp
+++ b/cajita/src/Cajita_GlobalGrid.hpp
@@ -183,6 +183,9 @@ class GlobalGrid
     std::array<int, num_space_dim> _global_cell_offset;
     std::array<bool, num_space_dim> _boundary_lo;
     std::array<bool, num_space_dim> _boundary_hi;
+
+    //! \brief Recompute global offset according to topology and _owned_num_cell
+    void computeGlobalOffset();
 };
 
 //---------------------------------------------------------------------------//
--- a/cajita/src/Cajita_GlobalGrid.hpp
+++ b/cajita/src/Cajita_GlobalGrid.hpp
@@ -54,22 +54,6 @@ class GlobalGrid
                 const std::array<bool, num_space_dim>& periodic,
                 const BlockPartitioner<num_space_dim>& partitioner );
 
-    /*!
-     \brief Constructor.
-     \param comm The communicator over which to define the grid.
-     \param global_mesh The global mesh data.
-     \param periodic Whether each logical dimension is periodic.
-     \param partitioner The grid partitioner.
-     \param local_offset The offset of the local grid
-     \param local_num_cell The number of owned cells of the local grid
-    */
-    GlobalGrid( MPI_Comm comm,
-                const std::shared_ptr<GlobalMesh<MeshType>>& global_mesh,
-                const std::array<bool, num_space_dim>& periodic,
-                const BlockPartitioner<num_space_dim>& partitioner,
-                const std::array<int, num_space_dim>& local_offset,
-                const std::array<int, num_space_dim>& local_num_cell );
-
     // Destructor.
     ~GlobalGrid();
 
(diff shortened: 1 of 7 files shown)


junghans:
format this please

junghans:
format this please

streeve (commented):
Not sure what @sslattery thinks on the mutable option here, but I think we need a few simple additions to the global grid unit test for set methods

aetx:
I am unsure what to test exactly. Just that `setOwnedNumCell(i, N); ownedNumCell(i) == N;`?

aetx:
I also don't particularly like the mutable, but am unsure how to set the two variables without it.

streeve:
> I am unsure what to test exactly. Just that `setOwnedNumCell(i, N); ownedNumCell(i) == N;`?

Yeah, that is a really simple test. But also testing that the mutable grid is in fact mutable

aetx:
I have added some tests, but am unsure how to test the mutability. Simply accessing the setter of `mutGlobalGrid` does not seem to trigger a failure, even though it does in ExaMPM.

junghans:
```
The following tests FAILED:
	 24 - Cajita_GlobalGrid_test_SERIAL_np_1 (Failed)
Errors while running CTest
	 25 - Cajita_GlobalGrid_test_SERIAL_np_2 (Failed)
```

aetx:
Yes, with b85213c the GlobalGrid test and LocalGrid are supposed to fail to test the test. However, LocalGrid does not seem to fail and I am unsure how to properly test the mutability of `mutGlobalGrid`.

junghans:
> Yes, with [b85213c](https://github.com/ECP-copa/Cabana/commit/b85213c0eb7524800f1942102386b966b30f052b) the GlobalGrid test and LocalGrid are supposed to fail to test the test. However, LocalGrid does not seem to fail and I am unsure how to properly test the mutability of `mutGlobalGrid`.

So what are the next step here?

aetx:
These tests should now check mutability of `mutGlobalGrid` and the setting of the global offset and owned number of cells.

If the tests should be more rigorous let me know. Otherwise this should be done.

cajita/src/Cajita_LocalGrid.hpp
```diff
@@ -55,6 +55,9 @@ class LocalGrid
     //! \brief Get the global grid that owns the local grid.
     const GlobalGrid<MeshType>& globalGrid() const;
 
+    //! \brief Get a mutable version of the global grid that own the local grid
+    GlobalGrid<MeshType>& mutGlobalGrid();
```
sslattery, line 59: you can call this function `globalGrid()`. Its OK to have a const and non-const version of the same accessor.

cajita/src/Cajita_GlobalGrid.hpp
```diff
@@ -157,11 +157,22 @@ class GlobalGrid
     //! \param dim Spatial dimension.
     int ownedNumCell( const int dim ) const;
 
+    //! \brief Set the owned number of cells in a given dimension of this block.
+    //! \param dim Spatial dimension.
+    //! \param num_cell New number of owned cells.
+    void setOwnedNumCell( const int dim, const int num_cell );
```
sslattery, line 163: I want this function to take all dimensions at once: `void setOwnedNumCell( const std::array<int,num_space_dim>& num_cell )`. In addition, I want a private function `computeGlobalOffset()` that automatically computes the global offsets given the owned number of cells. That way the partitioners don't need to compute that. This global offset compute function could also be called in the constructor then.

sslattery (changes requested):

cajita/src/Cajita_GlobalGrid.hpp
aetx: Should I also add the corresponding `ownedNumCell` that returns the array?

cajita/src/Cajita_LocalGrid.hpp
aetx: Ok, done.

cajita/src/Cajita_GlobalGrid.hpp
aetx: Regarding `computeGlobalOffset`: Are there already other helper functions present or should I just directly call `MPI_Exscan`?

aetx: With the added `computeGlobalOffset` I assume you also want that automatically called after setting the number of cells, so that everything is always up to date? Which would incur an additional (and for my load balancing case unnecessary) MPI communication.

junghans:
@aetx there is a warning now that something isn't documented in doxygen:
```
2021-07-29T15:23:06.2469163Z Generating docs for compound Cabana::CheckMemberTypes...
2021-07-29T15:23:06.3065749Z /__w/Cabana/Cabana/cajita/src/Cajita_GlobalGrid.hpp:160: warning: argument 'dim' of command @param is not found in the argument list of Cajita::GlobalGrid< MeshType >::setOwnedNumCell(const std::array< int, num_space_dim > &num_cell)
2021-07-29T15:23:06.3153716Z Generating docs for compound Cabana::CheckMemberTypes< MemberTypes< Types... > >...
```

junghans:
PS: @streeve doxygen Werror worked ;-)

cajita/src/Cajita_GlobalGrid.hpp
sslattery: yes - any time the number of cells is changed that will need to be called. it does incur communication but without it other objects that use this class will access it and it is not in a correct parallel consistent state.

aetx:
Ah yes, forgot to update the documentation for that.

(Working on the test case for `computeGlobalOffset`)

cajita/src/Cajita_GlobalGrid.hpp
aetx: As discussed this has changed to `setNumCellAndOffset` to force the update to either the number of cells or the offset to coincide. With that writing `computeGlobalOffset` is postponed.

streeve (approved):
Looks good to me!

junghans (approved):
More tests than actual code, I like it!

sslattery (approved):

PR #441, opened 2021-07-27
cmake: limit omp threads in tests

junghans:


aetx (approved):

streeve (approved):
I've never had any issue, but it looks fine to me. I wonder if this fixes anything related to what @sslattery saw with the interpolation tests quite a while ago

PR #443, opened 2021-08-05
Return particle-grid migration state

streeve:
For use in Picasso - return whether or not migration was performed to make further decisions

Also adds missing `const`


junghans (approved):

streeve:
@sslattery I think this matches Picasso use - do you agree?

PR #444, opened 2021-08-18
Cajita LoadBalancer interface to ALL load balancing library

aetx:
This pull request creates an interface to the ALL load balancing library, which generates a new, better balanced global grid decomposition.
https://gitlab.jsc.fz-juelich.de/SLMS/loadbalancing


streeve:
Retest this please.

cajita/unit_test/tstLoadBalancer.hpp
```diff
@@ -0,0 +1,96 @@
...
+    auto global_mesh = createUniformGlobalMesh(
+        global_low_corner, global_high_corner, global_num_cell );
+    std::array<bool, 3> is_dim_periodic = { false, false, false };
+
+    // Global grid
+    auto global_grid = createGlobalGrid( MPI_COMM_WORLD, global_mesh,
+                                         is_dim_periodic, partitioner );
+    int rank;
+    MPI_Comm_rank( MPI_COMM_WORLD, &rank );
+    LoadBalancer<Cajita::UniformMesh<double, 3>> lb( MPI_COMM_WORLD,
```
streeve, line 73: Could add a `createLoadBalancer` to match the others

streeve (changes requested):
We'll need to run these in the CI - I think updates to the GitHub Actions matrix is enough for now (1 new run with only ALL and adding ALL to the run with all the dependencies). Should be easy since already done in the proxies

cajita/unit_test/tstLoadBalancer.hpp
aetx: done

streeve (commented):
I linked a new issue to try to make sure the various dense/sparse load balancing options all work together. Otherwise just a few small comments

streeve:
Looks like the rerun failed here

cajita/unit_test/tstLoadBalancer.hpp
```diff
@@ -0,0 +1,519 @@
...
+    {
+        work = lbWork( vertices );
+        global_grid =
+            lb->createBalancedGlobalGrid( global_mesh, partitioner, work );
+        vertices = lb->getVertices();
+        internal_vertices = lb->getInternalVertices();
+        imbalance = lb->getImbalance();
+        // if ( rank == 0 )
+        //    printf( "LB Imbalance: %g\n", imbalance );
+        if ( imbalance < imbalance_limit )
```
streeve, line 111: Do I understand correctly that large numbers of ranks would require more steps to balance? @sslattery what do you think about a while loop until it hits the threshold vs a fixed maximum number of test steps the will fail for the largest cases?

streeve (approved):

cajita/unit_test/tstLoadBalancer.hpp
aetx: Exactly, without a minimum domain width the threshold will be reached, but it may take more steps. If a minimum domain width is set the same should be true in our setup as the number of cells per domain is constant and hence the loadbalancing resolution as well.

The `test_steps` limit is just there to avoid a possible infinite loop. But we can increase the limit, since it only affects large numbers of ranks anyway.

junghans:
@sslattery please do the final review! After that I will update the GitHub merge rules and merge.

junghans (approved):

sslattery (approved):

sslattery:
@aetx This is good to go - let's squash the history on this one and then we can merge. 1 commit will be fine.

PR #445, opened 2021-08-19
Add benchmark tests to CI

streeve:
Run performance tests with ctest on GHA and jenkins. We should set something up to make it easy to run these tests on CI/locally, as well as Summit etc.

### DIFF
--- a/benchmark/Cabana_CommPerformance.cpp
+++ b/benchmark/Cabana_CommPerformance.cpp
@@ -445,7 +445,7 @@ int main( int argc, char* argv[] )
 #ifdef KOKKOS_ENABLE_CUDA
     using CudaDevice = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>;
     using CudaUVMDevice = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaUVMSpace>;
-    using HostDevice = Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>;
+    using HostDevice = Kokkos::DefaultHostExecutionSpace::device_type;
 
     // Transfer GPU data to CPU, communication on CPU, and transfer back to
     // GPU.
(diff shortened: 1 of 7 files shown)


benchmark/Cabana_CommPerformance.cpp
```diff
@@ -445,14 +445,11 @@ int main( int argc, char* argv[] )
 #ifdef KOKKOS_ENABLE_CUDA
     using CudaDevice = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>;
     using CudaUVMDevice = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaUVMSpace>;
+    using HostDevice = Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>;
```
dalg24, line 448:
```suggestion
    using HostDevice = Kokkos::DefaultHostExecutionSpace::device_type;
```
so it works even in case the serial backend is not enabled

benchmark/Cabana_NeighborArborXPerformance.cpp
```diff
@@ -38,14 +38,14 @@ void performanceTest( std::ostream& stream, const std::string& test_prefix,
 
     // Declare problem sizes.
     double min_dist = 1.0;
-    std::vector<int> problem_sizes = { 1000, 10000, 100000, 1000000 };
+    std::vector<int> problem_sizes = { 100, 1000 };
```
dalg24, line 41: Are these taking too long?

streeve: These were all set up mainly for running on Summit, so too long for CI. Maybe I should set up defaults for single CPU, single GPU, scaling etc.

junghans (approved):

dalg24 (approved):

PR #448, opened 2021-08-26
Update to heFFTe 2.1 and enable MKL

streeve:
This adds FFT testing on the CPU for MKL and fixes cases with multiple host FFT backends

Previously added OneAPI/SYCL build, but will be moved to a separate PR (done on the host: it may be possible using Cuda SYCL, but probably not straightforward)

### DIFF
--- a/.github/workflows/CI2.yml
+++ b/.github/workflows/CI2.yml
@@ -68,6 +68,15 @@ jobs:
             heffte: 'NoheFFTe'
             hypre: 'NoHYPRE'
             coverage: 'OFF'
+          - distro: 'fedora:intel'
+            cxx: 'icpc'
+            openmp: 'OFF'
+            cmake_build_type: 'Debug'
+            kokkos_ver: 'master'
+            arborx: 'NoArborX'
+            heffte: 'heFFTe_MKL'
+            hypre: 'NoHYPRE'
+            coverage: 'OFF'
           - distro: 'fedora:latest'
             cxx: 'g++'
             openmp: 'ON'
@@ -235,7 +244,7 @@ jobs:
         if: ${{ matrix.heffte != 'NoheFFTe' }}
         # actions/checkout doesn't work for external repos yet
         run: |
-          git clone --depth 1 --branch v2.0.0 https://bitbucket.org/icl/heffte.git heffte
+          git clone --depth 1 --branch v2.1.0 https://bitbucket.org/icl/heffte.git heffte
       - name: Build heffte
         if: ${{ matrix.heffte != 'NoheFFTe' }}
         working-directory: heffte
@@ -245,11 +254,12 @@ jobs:
           cmake -B build \
             -DCMAKE_CXX_STANDARD="11" \
             -DBUILD_SHARED_LIBS=ON \
-            -DMKL_ROOT=/opt/intel/mkl \
             -DCMAKE_INSTALL_PREFIX=$HOME/heffte \
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+            -DMKL_ROOT=/opt/intel/oneapi/mkl/latest \
+            -DHeffte_MKL_IOMP5=/opt/intel/oneapi/compiler/latest/linux/compiler/lib/intel64_lin/libiomp5.so \
             ${heffte_cmake_opts}
           cmake --build build --parallel 2
           cmake --install build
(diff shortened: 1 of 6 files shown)


cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -364,13 +364,24 @@ struct HeffteBackendTraits<Kokkos::Cuda, Impl::FFTBackendDefault>
...
+#ifdef Heffte_ENABLE_ROCM
 #ifdef KOKKOS_ENABLE_HIP
 template <>
 struct HeffteBackendTraits<Kokkos::Experimental::HIP, Impl::FFTBackendDefault>
 {
     using backend_type = heffte::backend::rocfft;
 };
 #endif
+#endif
+#ifdef Heffte_ENABLE_ONEAPI
```
mkstoyanov, line 376: In previous versions of heFFTe, the backend type-tags were defined only if the corresponding preprocessor variable existed too.

In the latest master branch, all backend tags are always defined with the type-traits specialized structure `heffte::backend::is_enabled<type_tag>::value` to define whether it is available or not. This allows for a more C++-ish handing of these types.

At any rate, this is a F.Y.I.

.github/workflows/CI2.yml
```diff
@@ -203,13 +221,15 @@ jobs:
       - name: Build kokkos
         working-directory: kokkos
         run: |
+          [[ ${{ matrix.heffte }} == *"ONEAPI"* ]] && kokkos_cmake_opts=( "-DCMAKE_CXX_STANDARD=17 -DKokkos_ENABLE_SYCL=ON" )
```
mkstoyanov, line 224: The test gives an error on missing `libOpenCL.so`, which is part of OneAPI and should be installed.

Try sourcing the Intel setup script:
```
source /opt/intel/oneapi/setvars.sh
```
If things are still not working, I would try using `dpcpp` as the compiler in place of `icpc`

.github/workflows/CI2.yml
```diff
@@ -235,21 +258,25 @@ jobs:
...
         if: ${{ matrix.heffte != 'NoheFFTe' }}
         working-directory: heffte
         run: |
-          [[ ${{ matrix.heffte }} == "heFFTe_FFTW" ]] && heffte_cmake_opts+=( -DHeffte_ENABLE_FFTW=ON )
-          [[ ${{ matrix.heffte }} == "heFFTe_MKL" ]] && heffte_cmake_opts+=( -DHeffte_ENABLE_MKL=ON )
+          [[ ${{ matrix.heffte }} == *"heFFTe_FFTW"* ]] && heffte_cmake_opts+=" -DHeffte_ENABLE_FFTW=ON"
+          [[ ${{ matrix.heffte }} == *"heFFTe_MKL"* ]] && heffte_cmake_opts+=" -DHeffte_ENABLE_MKL=ON"
+          if [[ ${{ matrix.heffte }} == *"ONEAPI"* ]]; then
+            heffte_cmake_opts+=" -DHeffte_ENABLE_ONEAPI=ON"
+            source /opt/intel/oneapi/setvars.sh
```
junghans, line 270: Instead of doing that we should really set the right env variables in the container.....

mkstoyanov: > Instead of doing that we should really set the right env variables in the container.....

This script is part of Intel OneAPI and that's what they want people to use. You can source that inside the container build, but tracking each and every variable that Intel sets will be too much in my opinion.

junghans: The problem with sourcing it here is that it won't persist if we ever deploy the container. Plus if we fix it in the kokkos base container other consumers will profit from it as well.

Here are the lines where we setup the intel environment: 
https://github.com/kokkos/ci-containers/blob/main/fedora#L28-L29

junghans: Reading https://github.com/ECP-copa/Cabana/pull/448#discussion_r699419351, the solution might be as simple as adding the location of `libOpenCL.so` to `LD_LIBRARY_PATH`.

streeve: I was trying to get the SYCL build working before cleaning this up, but we're failing well before the FFT tests. We may need to separate that into a new PR (which also pushes off this sourcing issue)

junghans (approved):

.github/workflows/CI2.yml
```diff
@@ -245,11 +254,12 @@ jobs:
           cmake -B build \
             -DCMAKE_CXX_STANDARD="11" \
             -DBUILD_SHARED_LIBS=ON \
-            -DMKL_ROOT=/opt/intel/mkl \
             -DCMAKE_INSTALL_PREFIX=$HOME/heffte \
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+            -DMKL_ROOT=/opt/intel/oneapi/mkl/latest \
```
junghans, line 261: Add a note to clean this up later.

streeve:
Thanks for your help here, @mkstoyanov! I'll open a new PR for oneapi/sycl when I test more locally

mkstoyanov:
> Thanks for your help here, @mkstoyanov! I'll open a new PR for oneapi/sycl when I test more locally

Glad to help, keep me in the loop if you need more help @streeve

PR #450, opened 2021-09-07
CI: allow rawhide to fail

junghans:
Due to https://github.com/kokkos/ci-containers/issues/19 CI on rawhide is currently broken, so allow it to fail.


streeve:
How did rawhide pass? Is this an irregular failure?

junghans:
@streeve rawhide isn't getting updated anymore right now, hence it will ultimately fail at some point.

streeve (approved):

PR #451, opened 2021-09-09
WIP: Add FFT performance test

sfogerty:
Adding a performance test for 3d FFTs.


sfogerty:
@streeve here is my initial go at adding FFT performance testing. Could you take a look?

I'm thinking about adding 2D tests - leaning toward including them in Cajita_FastFourierTransformPerformance.cpp, rather than creating a separate test file

Another note - Most tests have `$device` options: `serial`, `openmp`, `cuda`... but we may want to split those up into other `${device}_${backend}` for this test?

streeve:
> I'm thinking about adding 2D tests - leaning toward including them in Cajita_FastFourierTransformPerformance.cpp, rather than creating a separate test file

I don't think that's a priority - can you post some example output for the backends we have working?

> Another note - Most tests have `$device` options: `serial`, `openmp`, `cuda`... but we may want to split those up into other `${device}_${backend}` for this test?

I agree

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,221 @@
...
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#include "Cabana_BenchmarkUtils.hpp"
+
+#include <Cajita_Array.hpp>
```
streeve, line 14: I would just include `Cajita.hpp`

streeve (changes requested):
We will definitely want Cuda and HIP performance results as well. Stuart mentioned that another team found that passing a work buffer to heffte can speed things up significantly (I guess by default it will allocate for each FFT). We can improve that here or in a follow-on PR

benchmark/Cajita_FastFourierTransformPerformance.cpp
sfogerty: Thanks @streeve for your helpful review
I made changes to address your comments, and can clean up commits if you think the PR is in good shape to merge.
I'll look at the allocation for FFTs and try to add a work buffer in a follow-on PR

streeve (commented):
Can you post results from Summit/Spock for this? We also probably want more system sizes to compare

sfogerty:
test this please

sfogerty:
Initial results for OpenMP on Broadwell, CUDA on V100, and HIP on MI25

Notes: the HIP performance on 2 nodes is shockingly bad in my testing. CUDA scaling is also poor. This could be linked to inefficient and unnecessary memory use. It may also be worth tuning how I am running these tests: binding processes, etc. to make best use of node resources.

Results split into time to set up FFT, and time to perform forward and reverse FFT 10 times (FFT size 200^3)

|   OpenMP   | setup | transform |
| ----------- |  ----------- | ----------- |
| 1 node      |   4.0e-5   | 9.53 |
| 2 nodes   |   4.3e-3  | 6.84 |
| 4 nodes   |  3.7e-3  | 4.15 |

|   CUDA   | setup | transform |
| ----------- |  ----------- | ----------- |
| 1 node      |   3.16e-3  |   0.29  |
| 2 nodes   |   3.64e-3  | 1.26 |
| 4 nodes   |  4.55e-3  |  0.89  |

|   HIP   | setup | transform |
| ----------- |  ----------- | ----------- |
| 1 node      |   3.16e-3  |   0.37  |
| 2 nodes   |   1.11e-3  |  14.82 |

streeve:
> Notes: the HIP performance on 2 nodes is shockingly bad in my testing. CUDA scaling is also poor. 

In part, this points to what we need to fix like you said. But I don't trust the network for the HIP example across nodes - 1 node is fine. I'll test on Spock when I get a chance.

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,232 @@
...
+                      const std::string& test_prefix )
+{
+    // Domain size setup
+    std::array<double, 3> global_low_corner = { 0.0, 0.0, 0.0 };
+    std::array<double, 3> global_high_corner = { 1.0, 1.0, 1.0 };
+    std::array<bool, 3> is_dim_periodic = { true, true, true };
+
+    // Declare the grid size per dimension
+    // currently, testing 3dims+symmetric
+    std::vector<double> grid_sizes_per_dim = { 10, 200 };
```
streeve, line 43: Let's modify to match the other benchmarks with a "small" (CI/default) test and "large" (Summit). I think that was changed after you started this test. Check the Actions output to make sure the CI run is ~1  sec

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,232 @@
...
+             << "\n";
+        file << "----------------------------------------------"
+             << "\n";
+        file << std::flush;
+    }
+
+    // Run the tests.
+#ifdef Heffte_ENABLE_FFTW
+#ifdef KOKKOS_ENABLE_SERIAL
+    using SerialDevice = Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>;
```
streeve, line 185: If you flip the ifdefs you can reuse the device types

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,232 @@
...
+    performanceTest<OpenMPDevice, Experimental::FFTBackendMKL>(
+        file, MPI_COMM_WORLD, "openmp_mkl_" );
+#endif
+#endif
+#ifdef Heffte_ENABLE_CUDA
+#ifdef KOKKOS_ENABLE_CUDA
+    using CudaDevice = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>;
+    performanceTest<CudaDevice, Experimental::Impl::FFTBackendDefault>(
+        file, MPI_COMM_WORLD, "openmp_cuda_" );
+
```
streeve, line 212: Remove blank

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,232 @@
...
+                }
+
+        // Copy to the device
+        Kokkos::deep_copy( lhs_view, lhs_host_view );
+
+        setup_timer.start( p );
+
+        // Create FFT options
+        Experimental::FastFourierTransformParams params;
+        params.setAllToAll( true );
```
streeve, line 111: Aren't these the defaults?

streeve (changes requested):
Close to ready to merge. Once you address comments can you rebase on master?

benchmark/Cajita_FastFourierTransformPerformance.cpp
sfogerty: @streeve thanks! I've changed the FFT sizes, since the old default was taking ~13 seconds.
In the last CI with new defaults, Jenkins ran the FFT performance test in 0.77s for HIP, and 2.26s for CUDA.

benchmark/Cajita_FastFourierTransformPerformance.cpp
sfogerty: Nice! Thanks I switched these

benchmark/Cajita_FastFourierTransformPerformance.cpp
sfogerty: Fixed

benchmark/Cajita_FastFourierTransformPerformance.cpp
sfogerty: Yes, these unnecessary lines are now removed.

sfogerty:
Thanks @streeve for looking over this. I've rebased and responded to comments above.

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,240 @@
...
+    performanceTest<OpenMPDevice, Experimental::FFTBackendMKL>(
+        file, grid_sizes_per_dim, MPI_COMM_WORLD, "openmp_mkl_" );
+#endif
+#endif
+
+#ifdef KOKKOS_ENABLE_CUDA
+    using CudaDevice = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>;
+#ifdef Heffte_ENABLE_CUDA
+    performanceTest<CudaDevice, Experimental::Impl::FFTBackendDefault>(
+        file, grid_sizes_per_dim, MPI_COMM_WORLD, "openmp_cuda_" );
```
streeve, line 220: Looks like a copy/paste label mistake here and for HIP

sfogerty: I'm not sure? I did switch the order of ifdefs here, just to match the ifdef ordering for other backends. It's not a necessary change, though. Very possible I'm just not seeing what is wrong though.

sfogerty: Ah, thanks! I need my vision checked

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,240 @@
...
+int main( int argc, char* argv[] )
+{
+    // Initialize environment
+    MPI_Init( &argc, &argv );
+    Kokkos::initialize( argc, argv );
+
+    // Check arguments.
+    if ( argc < 2 )
+        throw std::runtime_error( "Incorrect number of arguments. \n \
+             First argument - file name for output \n \
```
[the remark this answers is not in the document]
sfogerty: 👍

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,240 @@
...
+    {
+        run_type = argv[2];
+    }
+
+    // Declare the grid size per dimension
+    // currently, testing 3dims+symmetric
+    std::vector<double> grid_sizes_per_dim = { 10, 20 };
+    if ( run_type == "large" )
+    {
+        grid_sizes_per_dim = { 100, 200 };
```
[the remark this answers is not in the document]
sfogerty: This is per dimension, so should be particle numbers of {1e3, 8e3}, {1e6,8e6} for 3D

benchmark/Cajita_FastFourierTransformPerformance.cpp
```diff
@@ -0,0 +1,240 @@
...
+        auto vector_layout = createArrayLayout( local_grid, 2, Cell() );
+        auto lhs = createArray<double, Device>( "lhs", vector_layout );
+        auto lhs_view = lhs->view();
+        auto lhs_host =
+            createArray<double, typename decltype( lhs_view )::array_layout,
+                        Kokkos::HostSpace>( "lhs_host", vector_layout );
+        auto lhs_host_view = Kokkos::create_mirror_view( lhs_view );
+        uint64_t seed = global_grid->blockId() +
+                        ( 19383747 % ( global_grid->blockId() + 1 ) );
+        using rnd_type = Kokkos::Random_XorShift64_Pool<Kokkos::HostSpace>;
```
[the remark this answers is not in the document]
sfogerty: Good point. I don't think there is any issue - switched to using device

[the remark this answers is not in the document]
streeve: It's failing now because the loop below needs to be a Kokkos kernel and use the device views (shouldn't need any of the host views anymore). Also, pass the memory space from the device rather than the device itself.

benchmark/Cajita_FastFourierTransformPerformance.cpp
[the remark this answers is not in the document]
streeve: I was meaning to set grid sizes to end up with ~1e3, 1e4, 1e5, 1e6 total grid points, as we do with particles in other tests

streeve:
Once you address these last few comments, go ahead and squash into one commit and we should be ready to merge

benchmark/Cajita_FastFourierTransformPerformance.cpp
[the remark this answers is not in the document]
sfogerty: Changed to {10,20} and {50,100} which is {1e3,0.8e4} and {1.25e5,1e6}

benchmark/Cajita_FastFourierTransformPerformance.cpp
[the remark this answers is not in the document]
sfogerty: Ah, that's great we can remove the host view and copy. Fixed to use a Kokkos kernel on the device, and got rid of host view

streeve (approved):
Looks good

PR #464, opened 2021-10-26
Update GlobalGrid and Partitioner

YuxingQiu:
In this pull request, we update parts of the implementation in `GlobalGrid` and `Partitioner`, such that the `GlobalGrid` can support the `SparseMesh` mesh type as well.

**Changes made in:**

1. `GlobalGrid` constructor: all partition related operations are moved to the Partitioner class series
2. Partitioner series:
    1. Base class `BlockPartitioner`: add a virtual interface `getOwnedCellInfo` which will take in the MPI cartesian communicator and an array to record the number of global cells in each dimension, and output the owned number of cells and the global cell offset of the current MPI rank.
    2. Overriders in derived classes: `ManualBlockPartitioner`, `DimBlockPartitioner` and `SparseDimPartitioner`
        1. In `DimBlockPartitioner` and the `SparseDimParititoner`, the `owned_num_cell` and the `global_cell_offset` are dynamically computed from the input `global_cells_per_dim`.
        2. In `SparseDimPartitioner`, the `owned_num_cell` and the `global_cell_offset` are dynamically computed from the rec_partitioner, while the input `global_cells_per_dim` is not used anywhere. Please note, the unit of partition information is a tile but not a cell. Therefore, another interface called `getOwnedTileInfo` is added to support the implementation of `getOwnedCellInfo`.
3. Unit tests in `tstGlobalGrid` for the sparse grid case.

**User Doc:**
1. Users should take responsibility to initialize the partitioner (for sparse grid case); the owned cell information is meaningless if the partitioner is not initialized.
2. For dynamic partitioners, users are responsible for updating their `global_grid` variable (`owned_cell_num` and `global_cell_offset`) by using interface `setNumCellAndOffset` defined in `GlobalGrid`.
3. For dynamic partitioners, users can retrieve the `owned_cell_num` and `global_cell_offset` by calling `getOwnedCellInfo` defined in partitioners after partitions being updated. (Please see the last few lines in unit test `3d_sparse_grid_test` in `tstGlobalGrid.hpp` )

### DIFF
--- a/cajita/src/Cajita_Partitioner.hpp
+++ b/cajita/src/Cajita_Partitioner.hpp
@@ -27,7 +27,6 @@ namespace Cajita
 /*!
   \brief Block partitioner base class.
   \tparam NumSpaceDim Spatial dimension.
-
   Given global mesh parameters, the block partitioner computes how many MPI
   ranks are assigned to each logical dimension.
 */
@@ -46,16 +45,31 @@ class BlockPartitioner
       \param global_cells_per_dim The number of global cells in each dimension.
       \return The number of MPI ranks in each dimension of the grid.
     */
-    virtual std::array<int, NumSpaceDim> ranksPerDimension(
+    virtual std::array<int, num_space_dim> ranksPerDimension(
         MPI_Comm comm,
-        const std::array<int, NumSpaceDim>& global_cells_per_dim ) const = 0;
+        const std::array<int, num_space_dim>& global_cells_per_dim ) const = 0;
+
+    /*!
+      \brief Get the owned number of cells and global cell offset of the current
+      MPI rank.
+      \param cart_comm The MPI Cartesian communicator for the partitioning.
+      \param global_cells_per_dim The number of global cells in each dimension.
+      \param owned_num_cell (Return) The owned number of cells of the current
+      MPI rank in each dimension.
+      \param global_cell_offset (Return) The global cell offset of the current
+      MPI rank in each dimension
+    */
+    virtual void ownedCellInfo(
+        MPI_Comm cart_comm,
+        const std::array<int, num_space_dim>& global_cells_per_dim,
+        std::array<int, num_space_dim>& owned_num_cell,
+        std::array<int, num_space_dim>& global_cell_offset ) const = 0;
 };
 
 //---------------------------------------------------------------------------//
 /*!
   \brief Manual block partitioner.
   \tparam NumSpaceDim Spatial dimension.
-
   Assign MPI blocks from a fixed user input.
 */
 template <std::size_t NumSpaceDim>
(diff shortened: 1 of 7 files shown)


streeve (changes requested):

cajita/unit_test/tstGlobalGrid.hpp
```diff
@@ -390,6 +391,196 @@ void gridTest2d( const std::array<bool, 2>& is_dim_periodic )
...
+    auto part = partitioner.getCurrentPartition();
+    for ( int d = 0; d < 3; ++d )
+        for ( int id = 1; id < ranks_per_dim[d]; id++ )
+            part[d][id] += 1;
+
+    partitioner.initializeRecPartition( part[0], part[1], part[2] );
+
+    std::array<int, 3> new_owned_num_cell;
+    std::array<int, 3> new_global_cell_offset;
+    partitioner.getOwnedCellInfo( global_grid->comm(), global_num_cell,
```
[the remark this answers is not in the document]
YuxingQiu: I don't feel it quite necessary here for the dense grids for several reasons.
1) This function is already called in the constructor of GlobalGrid (in line 54 in `Cajita_ClobalGrid_impl.hpp`) and serves as the base of the rest of the code and checks. So basically other tests have already covered this function.
2) This unit test is for GlobalGrid but not for Partitioner, and the purpose of lines 554 - 581 is to check the behavior of function `setNumCellAndOffset` in `GlobalGrid`. 
Here's the reason why this function is called for the sparse grid but not for the dense grid here: 
Since we have a dynamic load balancer for the sparse grid, the sparse partition can tell the owned cell and the offset information to the global grid, providing `setNumCellAndOffset` a more reasonable source. But the dense grid partitioner will only provide the `ranks_per_dim` info and assume an average partition. So calling this function will only generate the same `ownedNumCell` and `globalCellOffset` as before. This doesn't fit the purpose of testing `setNumCellAndOffset` with new settings.

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -37,6 +37,9 @@ template <typename Device, unsigned long long CellPerTileDim = 4>
 class SparseDimPartitioner : public BlockPartitioner<3>
 {
   public:
+    //! dimension
+    static constexpr std::size_t num_space_dim = 3;
```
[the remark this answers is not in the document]
YuxingQiu: I will change it to a general setting for now. But we need to reimplement some member functions for the 2D case, which may require another PR.

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -187,17 +194,73 @@ class SparseDimPartitioner : public BlockPartitioner<3>
       \brief Get the cell number in each dimension owned by the current MPI rank
       \param cart_comm MPI cartesian communicator
     */
-    std::array<int, 3> ownedCellsPerDimension( MPI_Comm cart_comm ) const
+    std::array<int, num_space_dim>
+    ownedCellsPerDimension( MPI_Comm cart_comm ) const
```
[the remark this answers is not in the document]
YuxingQiu: Added a new interface in the dense partitioner.

cajita/src/Cajita_Partitioner.hpp
```diff
@@ -93,6 +109,61 @@ class ManualBlockPartitioner : public BlockPartitioner<NumSpaceDim>
...
+
+        // Compute the number of local cells in this rank in each dimension.
+        for ( std::size_t d = 0; d < num_space_dim; ++d )
+        {
+            owned_num_cell[d] = cells_per_dim[d];
+            if ( dim_remainder[d] > cart_rank[d] )
+                ++owned_num_cell[d];
+        }
+    }
+
```
streeve, line 166: Shouldn't we have `ownedCellsPerDimension` here as well?

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -187,17 +195,73 @@ class SparseDimPartitioner : public BlockPartitioner<3>
...
+    /*!
+      \brief Get the owned number of cells and the global cell offset of the
+      current MPI rank.
+      \param cart_comm The MPI Cartesian communicator for the partitioning.
+      \param owned_num_cell (Return) The owned number of cells of the current
+      MPI rank in each dimension.
+      \param global_cell_offset (Return) The global cell offset of the current
+      MPI rank in each dimension
+    */
+    void getOwnedCellInfo(
```
streeve, line 251: How about `ownedCellInfo` and `ownedTileInfo` - we don't return the array in these

cajita/src/Cajita_Partitioner.hpp
```diff
@@ -135,6 +206,94 @@ class DimBlockPartitioner : public BlockPartitioner<NumSpaceDim>
...
+      \brief Get the owned number of cells and the global cell offset of the
+      current MPI rank.
+      \param cart_comm The MPI Cartesian communicator for the partitioning.
+      \param global_cells_per_dim The number of global cells in each dimension.
+      \param owned_num_cell (Return) The owned number of cells of the current
+      MPI rank in each dimension.
+      \param global_cell_offset (Return) The global cell offset of the current
+      MPI rank in each dimension
+    */
+    void getOwnedCellInfo(
```
streeve, line 220: I'm sure you've considered this, but can we share more between `ownedCellInfo` and `ownedCellsPerDimension`? Maybe with some private helper functions

cajita/src/Cajita_SparseDimPartitioner.hpp
```diff
@@ -33,10 +33,14 @@ namespace Cajita
   \tparam Device Kokkos device type.
   \tparam CellPerTileDim Cells per tile per dimension.
 */
-template <typename Device, unsigned long long CellPerTileDim = 4>
-class SparseDimPartitioner : public BlockPartitioner<3>
+template <typename Device, unsigned long long CellPerTileDim = 4,
+          std::size_t NumSpaceDim = 3>
+class SparseDimPartitioner : public BlockPartitioner<NumSpaceDim>
```
streeve, line 38: The preparation for dimension generalization is nice, but I think we need a note in the doxygen that this only supports 3D

streeve (changes requested):
Looks really nice, tests should pass with doxygen update - happy to discuss on slack for the last few details

cajita/src/Cajita_Partitioner.hpp
```diff
@@ -93,6 +107,148 @@ class ManualBlockPartitioner : public BlockPartitioner<NumSpaceDim>
...
+      \brief Get the owned number of cells of the current MPI rank.
+      \param cart_comm The MPI Cartesian communicator for the partitioning.
+      \param global_cells_per_dim The number of global cells and the global
+      cell offset in each dimension.
+      \param owned_num_cell (Return) The owned number of cells of the current
+      MPI rank in each dimension.
+      \param global_cell_offset (Return) The global cell offset of the current
+      MPI rank in each dimension
+    */
+    void ownedCellInfo(
```
streeve, line 120: Can you add a very simple `tstPartitioner` file that tests these two interfaces for manual and average partitions? Since we moved it here and rely on it in the global grid test I think that would be best. If you don't have time one of us here can help

cajita/unit_test/tstPartitioner.hpp
```diff
@@ -0,0 +1,258 @@
...
+        global_low_corner, global_high_corner, global_num_cell );
+
+    auto global_grid = createGlobalGrid( MPI_COMM_WORLD, global_mesh,
+                                         is_dim_periodic, partitioner );
+
+    // Get the cells per dimension and the remainder with partitioner
+    std::array<int, 3> owned_num_cell;
+    std::array<int, 3> global_cell_offset;
+
+    partitioner.ownedCellInfo( global_grid->comm(), global_num_cell,
```
streeve, line 88: Let's remove the global mesh and grid entirely and just use COMM_WORLD here

streeve:
One small change, then we can squash commits and merge

cajita/unit_test/tstPartitioner.hpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+void testBlockPartitioner3d()
+{
+    DimBlockPartitioner<3> partitioner;
+    std::array<bool, 3> is_dim_periodic = { false, false, false };
+    owned_cell_info_test_3d( partitioner, is_dim_periodic );
+
+    for ( std::size_t d = 0; d < 3; ++d )
+        is_dim_periodic[d] = true;
```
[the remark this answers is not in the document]
YuxingQiu: No, it has nothing to do with the partitioning, but `MPI_Cart_create` requires it. Do you think it's okay to use all `true` for cart_comm creation?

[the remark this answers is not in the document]
streeve: Yes, let's pass true to MPI and remove it everywhere else

cajita/unit_test/tstPartitioner.hpp
```diff
@@ -0,0 +1,203 @@
...
+void testManualPartitioner3d()
+{
+    // Let MPI compute the partitioning for this test.
+    int comm_size;
+    MPI_Comm_size( MPI_COMM_WORLD, &comm_size );
+    std::array<int, 3> ranks_per_dim = { 0, 0, 0 };
+    MPI_Dims_create( comm_size, 3, ranks_per_dim.data() );
+
+    {
+        ManualPartitioner partitioner( ranks_per_dim );
```
streeve, line 160: Sorry I keep finding one more thing at a time, but we should use `ManualBlockPartitioner<3>` here and add `ManualBlockPartitioner<2>` as well (`ManualPartitioner` is a backwards compatibility wrapper)

streeve (approved):

PR #469, opened 2021-11-17
Add local mesh default constructor

streeve:
Useful in Picasso for storing the mesh on the device for boundary condition checks


streeve:
@sslattery ready to merge?

cajita/src/Cajita_LocalMesh.hpp
```diff
@@ -50,6 +50,9 @@ class LocalMesh<Device, UniformMesh<Scalar, NumSpaceDim>>
     //! Kokkos execution space.
     using execution_space = typename Device::execution_space;
 
+    //! Default constructor.
+    LocalMesh() {}
```
sslattery, line 54: `LocalMesh() = default;`

streeve: fixed

sslattery (approved):

PR #472, opened 2021-12-09
Require Kokkos minimum version 3.4

dalg24:
Per our discussion at the developer meeting we will only support the latest two Kokkos version.

I haven't updated the Wiki yet.


streeve:
@YuxingQiu HIP is failing on the sparse partitioner. Have you seen failures like this before?

streeve:
I believe `core/src/impl/Cabana_CartesianGrid.hpp` is the only place using the Kokkos math functions

YuxingQiu:
> @YuxingQiu HIP is failing on the sparse partitioner. Have you seen failures like this before?

Not quite sure. I have never seen this specific error before. But it's strange as the CUDA tests passed. 

I was once in a similar situation before when CUDA passed, but HIP failed (with a different error). At that time, it was because of the different Kokkos atomic_or behaviors on CUDA and HIP. It took time to find that error, and we added a unit test called `testAtomicOr` in `tstSparseIndexSpace.hpp`.

streeve:
@dalg24 this should now pass with the missing fences from #517

junghans (approved):

streeve (approved):
Waiting tests are duplicates - merging without them

Issue #474, opened 2021-12-13
Avoid using Kokkos::Impl


masterleinad:
@streeve and I discovered that unit tests are failing for `SYCL` when running with `Intel`  GPUs. I'm pretty sure that this is related to https://github.com/kokkos/kokkos/pull/4349 after which `Kokkos::Impl::DeepCopy` requires the call site to enforce a global fence for proper Kokkos semantics, i.e., the two-argument deep copy fences before and after the copy operation. Just using `Kokkos::deep_copy` with an unmanaged Kokkos::View would fix that issue.

This is a good reminder to just not use any functionality in the `Kokkos::Impl` namespace. We don't guarantee any backward-compatibility for that.

Current places we reach in to `Kokkos::Impl`:
- [x] `Kokkos::Impl::DeepCopy<Dst, Src>( dst, src, bytes ) );`
- [x] `Kokkos::Impl::PolicyTraits` (in creating the `SimdPolicy`)
- [x] `Kokkos::Impl::min_max_functor<KeyViewType>( keys )`
- [x] `Kokkos::Impl::CudaTraits::WarpSize`

streeve:
Tests are all now passing

PR #478, opened 2021-12-16
Do not use `Kokkos::Impl::DeepCopy`

dalg24:
Partial fix for #474


sslattery (approved):

streeve:
This should have fixed the fencing issue mentioned in #474. I will test

PR #480, opened 2021-12-16
Cleanup `SimdPolicy` avoid messing with underlying `TeamPolicy`

dalg24:
Do not mess with the properties of the `Kokkos::TeamPolicy`, let the functor accept non-void work tags as leading argument.

### DIFF
--- a/core/src/Cabana_Parallel.hpp
+++ b/core/src/Cabana_Parallel.hpp
@@ -74,6 +74,54 @@ KOKKOS_FORCEINLINE_FUNCTION
     functor( t, std::forward<IndexTypes>( indices )..., reduce_val );
 }
 
+template <class ExecutionPolicy, class Functor>
+struct ParallelFor;
+
+template <class Functor, int VectorLength, class... Properties>
+struct ParallelFor<SimdPolicy<VectorLength, Properties...>, Functor>
+{
+    using simd_policy = SimdPolicy<VectorLength, Properties...>;
+    using team_policy = typename simd_policy::base_type;
+    using work_tag = typename team_policy::work_tag;
+    using index_type = typename team_policy::index_type;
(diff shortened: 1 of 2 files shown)


dalg24:
This is ready for review.  I surely can't be blamed here for the coverage.

core/src/Cabana_Parallel.hpp
```diff
@@ -74,6 +74,54 @@ KOKKOS_FORCEINLINE_FUNCTION
...
+        else
+            Kokkos::parallel_for(
+                label, dynamic_cast<const team_policy&>( exec_policy_ ),
+                *this );
+    }
+
+    template <class WorkTag>
+    KOKKOS_FUNCTION std::enable_if_t<!std::is_void<WorkTag>::value &&
+                                     std::is_same<WorkTag, work_tag>::value>
+    operator()( WorkTag, member_type const& team ) const
```
streeve, line 108: So this is just adding flexibility if you want a different tag than the `SimdPolicy`?

streeve (approved):
This is great - likely the structure we want for updating the neighbor parallel

streeve:
> I surely can't be blamed here for the coverage.

Agreed - opened #483 to remind us to add

core/src/Cabana_Parallel.hpp
streeve: @dalg24 walked me through this - resolving

sslattery (approved):
Very nice! I like this solution. Much better way to do the tag dispatch.

dalg24:
What's going on with the check statuses?  It looks like they are some pending duplicates.
What is the policy?  Can we override and merge?

streeve:
It's because #444 updated the required tests. It's easier to rebase and rerun - otherwise we have to manually enable/disable all the tests again

PR #481, opened 2021-12-16
Use external install of GTest when suitable version is found

dalg24:
NOTE to reviewers:  make sure to hide whitespaces.

The changes I suggest are just
```CMake
find_package(GTest)
if(NOT GTest_FOUND)
  <download gtest and install indented>
endif() 
```
If a suitable install of googletest can be found on the system, use it instead of downloading and building our own.


streeve (approved):

streeve:
> make sure to hide whitespaces.
🙏

sslattery (approved):

PR #484, opened 2022-01-03
Add Silo output

streeve:
Move Silo particle output from Picasso (new optional dependency). Will be useful for direct particle output - could be moved to experimental depending on plans for future output options

Marked as draft because the core version is not tested in the CI yet (and will likely want Silo installed in ci-container first)

Includes
- Core version with user-specified comm and number of groups
- Cajita version using the global grid


CMakeLists.txt
```diff
@@ -102,6 +102,14 @@ Cabana_add_dependency( PACKAGE HYPRE VERSION 2.22.1 )
 # find heffte
 Cabana_add_dependency( PACKAGE Heffte VERSION 2.1.0 )
 
+# find Silo
+Cabana_add_dependency( PACKAGE SILO )
+if(SILO_FOUND)
+  install(FILES
+    ${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindSILO.cmake
```
dalg24, line 109: The development version of SILO seems to be exporting the package configuration with CMake.  https://github.com/LLNL/Silo/blob/a1e27d34e8b52196be2de3d4aae2d0cae21b3f36/CMakeLists.txt#L675-L706

Do we actually need to write our own find module?

junghans: @markcmiller86: Does silo have a CMake find module?

markcmiller86: @junghans I don't know but @biagas would. I know @biagas is just adding CMake build functionality to Silo but I honestly don't know if that included its own FindSilo functionality to be exported for other projects to use. @biagas, do you have any comments/thoughts here?

biagas: The cmake build and package config exports is new and not yet well tested on all platforms, so it is still considered experimental at this stage, especially the exported cmake configs.

That said, the current state of things regarding CMake and Silo:

The CMake build should work on linux and Windows.

There isn't a find module for Silo. CMake should be able to find it with "find_package"' if Silo_DIR is set to location of Silo's install cmake subdir (eg location of SiloConfig.cmake) before the call to find_package.  Alternatively, including the SiloConfig.cmake file directly should work.

If Silo was built with HDF5 support, path to HDF5 should also be specified before the call to find_package for Silo, or perform a Find for HDF5 before Silo.

Here's a sample CMakeLists.txt I used for testing find_package with Silo (on Windows)

```
cmake_minimum_required(VERSION 3.14)

project(TestImportingSilo)

# path to Silo
set(Silo_DIR "/path/to/installed_silo/cmake")

# for HDF5 built with CMake, SiloConfig.cmake will use this to find hdf5
set(HDF5_DIR "/path/to/installed_hdf5/cmake")

find_package(Silo REQUIRED CONFIG)

if(Silo_FOUND)
    message("Silo_FOUND")
else()
    message("NOT Silo_FOUND")
endif()

```

.github/workflows/CI2.yml
```diff
@@ -310,6 +321,19 @@ jobs:
...
+        with:
+          repository: LLNL/Silo
+          ref: 4.11
+          path: silo
+      - name: Build Silo
+        if: ${{ matrix.silo == 'Silo' }}
+        working-directory: silo
+        run: |
+          ./configure CC=mpicc CXX=mpicxx --enable-fortran=no --prefix=$HOME/silo
+          make install
```
junghans, line 336: we are really getting to the point, where we should build all deps, but kokkos (and arborX) in the CI-Container.
And then force CMake to not find it using `-DCMAKE_DISABLE_FIND_PACKAGE_<Foo>=ON`.

CMakeLists.txt
junghans: @biagas does it matter how we build silo or do we need to build silo with CMake to get `SiloConfig.cmake`.

markcmiller86: @junghans I just wanna make sure you are aware that the CMake work @biagas has completed so far is not in any released version of Silo yet. Its on our [`main` branch](https://github.com/LLNL/Silo/commits/main)

streeve:
@junghans I at least got this to fail with `REQUIRE_SILO`; any ideas on why it can't find the container installed Silo?

junghans:
> @junghans I at least got this to fail with `REQUIRE_SILO`; any ideas on why it can't find the container installed Silo?

Got it, under Ubuntu the library is called `siloh5`, no idea why?

@markcmiller86  @biagas any ideas on this?

junghans:
> > @junghans I at least got this to fail with `REQUIRE_SILO`; any ideas on why it can't find the container installed Silo?
> 
> Got it, under Ubuntu the library is called `siloh5`, no idea why?
> 
> @markcmiller86 @biagas any ideas on this?

Got it, Debian patches that in https://sources.debian.org/patches/silo-llnl/4.11-2/add-lib-version.patch/

streeve:
@sslattery @dalg24 @junghans this is ready for review unless anyone wants to wait for Silo CMake support in a release

.github/workflows/CI2.yml
streeve: Added - I checked that the non-Silo builds do not include Silo

CMakeLists.txt
streeve: For now we plan to support non-CMake builds and switch to CMake-only when released - thanks for all the details

sslattery (approved):

cmake/CabanaConfig.cmakein
```diff
@@ -38,3 +38,7 @@ set(Cabana_ENABLE_ALL @Cabana_ENABLE_ALL@)
 if(Cabana_ENABLE_ALL)
   find_dependency(ALL @ALL_VERSION@ REQUIRED)
 endif()
+set(Cabana_ENABLE_SILO @Cabana_ENABLE_SILO@)
+if(Cabana_ENABLE_SILO)
+  find_dependency(SILO @SILO_VERSION@ REQUIRED)
```
dalg24, line 43: Do we test against an installed Cabana.  Do we need to do anything to make sure the `FindSILO.cmake` module we installed is visible?

CMakeLists.txt
streeve: We currently add and link `SILO::silo` - @biagas will that break compatibility with the current/planned CMake support?

cmake/CabanaConfig.cmakein
streeve: We have one for the base library, but not for installed dependencies. @junghans do you think we should add one for each optional library?

CMakeLists.txt
biagas: We haven't namespaced silo in the CMake support yet, but it makes a lot of sense to do so. 

Shouldn't be a problem. I've created a [ticket](https://github.com/LLNL/Silo/issues/242) to ensure the exported targets are namespaced with 'SILO'.

streeve: Thanks!

cmake/CabanaConfig.cmakein
junghans: We have `list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}" )` above so that should add the folder that has `FindSILO.cmake`.

dalg24 (approved):

PR #488, opened 2022-01-20
CI: add missing issue link

junghans:


.github/workflows/CI2.yml
```diff
@@ -233,7 +233,7 @@ jobs:
           cmake --install build
       - name: Checkout heffte
         if: ${{ matrix.heffte != 'NoheFFTe' }}
-        # actions/checkout doesn't work for external repos yet
+        # actions/checkout doesn't work for external repos yet (actions/checkout#447)
```
streeve, line 236: This would apply to ALL as well, correct?

junghans: yeah

streeve (approved):

Issue #470, opened 2021-11-18
Efficient memory use in FFTs


sfogerty:
Simple improvements to memory use in the Cabana FFT implementation could have an outsize impact on performance. 

1) Get rid of the data copies. Instead of converting between various types of complex data, we could use `Kokkos::complex<Scalar>`. The issue with this was the alignment in `Kokkos::complex`, but could be resolved with `-DKOKKOS_ENABLE_COMPLEX_ALIGN=ON`

2) heFFTe may be allocating a work buffer for each FFT? If so we should pass it a work buffer to use. This could speed things up significantly.

sfogerty:
@streeve @sslattery I made an issue here to try capturing potential improvements for FFT performance.

streeve:
With #451 merged can you start on these @sfogerty? Ideally for each performance update you can run the test on each backend and show the improvement. I have a small python script to compare if that's useful

streeve:
@sfogerty is it relatively straightforward to add the other optimization here?

PR #495, opened 2022-02-02
Use workspaces in FFTs

sfogerty:
The goal is to create a work buffer for FFTs when a Cabana FFT is created. This will be re-used every time a forward or backward FFT is computed. heFFTe by default creates a new work buffer on each computation, hurting performance.

Part of #470

### DIFF
--- a/cajita/src/Cajita_FastFourierTransform.hpp
+++ b/cajita/src/Cajita_FastFourierTransform.hpp
@@ -474,6 +474,9 @@ class HeffteFastFourierTransform
         _fft_work = Kokkos::View<Scalar*, DeviceType>(
             Kokkos::ViewAllocateWithoutInitializing( "fft_work" ),
             2 * fftsize );
+        _workspace = Kokkos::View<Scalar* [2], DeviceType>(
+            Kokkos::ViewAllocateWithoutInitializing( "workspace" ),
+            4 * fftsize );
     }
 
     /*!
@@ -525,6 +528,7 @@ class HeffteFastFourierTransform
             _fft->forward(
                 reinterpret_cast<std::complex<Scalar>*>( _fft_work.data() ),
                 reinterpret_cast<std::complex<Scalar>*>( _fft_work.data() ),
+                reinterpret_cast<std::complex<Scalar>*>( _workspace.data() ),
                 scale );
         }
         else if ( flag == -1 )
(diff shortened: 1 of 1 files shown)


sfogerty:
@streeve here is my initial go at getting workspaces used for FFTs. While I had tried a few different things, what I pushed here is the most bare-bones attempt I could conjure. Included are some debugging printouts and comments with values. (Those values are for running the default FFT benchmark which starts with a 10x10x10 3D FFT)

I'm confused by some of the values I'm seeing, and probably making a dumb mistake. This really seems like it should be simple! If you have any tips or want to chat sometime soon, please put a comment here or in Slack.

streeve:
Changeset looks good; waiting to see performance results on Summit/Frontier-like systems @sfogerty

streeve:
Results look really good on HIP. Was this using the FFT benchmark? That would let you compare systems sizes from 1e3-1e6 on one rank, rather than across ranks and understand. Maybe also compare on one node, but not necessary

Then can you squash commits and rebase on master?

sfogerty:
Will update with numbers for 1 node, scaling the system size. I was using the FFT benchmark in `benchmarks/cajita/` so the data should already be stored in the output files. Then rebase + clean up commits.

sfogerty:
Tests with master branch and with workspaces (1 rank)

|   OpenMP   | setup | transform |  setup_workspaces | transform_workspaces |
| ----------- |  ----------- | ----------- | ----------- | ----------- |
| 200^3     |   6.3e-5   | 10.41 | 6.98e-5 | 7.97 |



|   CUDA   | setup | transform | setup_workspaces | transform_workspaces |
| ----------- |  ----------- | ----------- | ----------- | ----------- |
| 200^3     |   3.05e-3  |   0.29  | 1.30e-4  |   0.19  | 
| 200^3     |   3.05e-3  |   0.29  | 1.30e-4  |   0.19  | 
| 200^3     |   3.05e-3  |   0.29  | 1.30e-4  |   0.19  | 
| 200^3     |   3.05e-3  |   0.29  | 1.30e-4  |   0.19  | 


|   HIP   | setup | transform | setup_workspaces | transform_workspaces |
| ----------- |  ----------- | ----------- |----------- | ----------- |
| 10^3      |   6.49e-5  |   0.015  | 1.88e-4 | 0.019 |
| 25^3      |   8.61e-5  |   0.019  | 2.13e-4 | 0.013 |
| 50^3      |   1.38e-4  |   0.024  | 2.62e-4 | 0.019 |
| 100^3      |   3.48e-4  |   0.054  | 4.35e-4 | 0.045 |
| 200^3      |   1.64e-3  |   0.38  | 1.67e-3 | 0.36 |

0.00167e-6

streeve (approved):

streeve:
Looks like around 20% speedup for HIP and up to 2x for Cuda - looks good

Issue #496, opened 2022-02-04
documentation/examples for interfaces to third-party libraries


XzzX:
It is hard to figure out how to use the interfaces for `heFFTe`, `HYPRE`, `ALL` and `ArborX`. I think the best sources are currently the unit tests. Can you add short usage examples and/or a dedicated wiki page for these interfaces?

streeve:
For each dependency, let's add an in-repo example and add to/update the wiki page (linked from https://github.com/ECP-copa/Cabana/wiki/Programming-Guide):

- @sfogerty heFFTe
- @sslattery hypre
- @aetx ALL
- I'll do ArborX (very little to be done anyway)

PR #507, opened 2022-03-10
Loadbalancer example

aetx:
Add a more documented (and slightly changed) version of the loadbalancer unit test as an example.

The example number has to be adapted to the final position.

Part of #496


streeve:
@kwitaechong @abisner can you give a final review on this?

streeve (approved):

sslattery (approved):

PR #508, opened 2022-03-24
fix gtest not found issue

kwitaechong:


streeve:
Like we discussed at the call, let's plan to remove all internal gtest and replace with a top-level `find_package`

streeve:
Closing in favor of #510

PR #509, opened 2022-03-25
Removed value_type requirement for view functors in g2p vector gradient.

abisner:


streeve:
Retest this please.

streeve (commented):
`value_type` is still noted as required in a lot of comments - is that still the case?

abisner:
The use of `ViewType::value_type` in the `G2P::` and `P2G::` methods was a holdover from an earlier revision, according to @sslattery. But it does not appear to be functionally necessary. It was only explicitly present in the 2D/3D vector gradient `G2P::` operators, and nowhere else (not for the scalar gradient, divergence, etc.)

streeve (approved):
Nice catch

sslattery (approved):

PR #510, opened 2022-03-25
Remove gtest internal build

streeve:
Treat googletest as a normal dependency and require if testing turned on. This may be an issue for any users not realizing this will no longer be downloaded for them

Missing updates for CI


streeve:
@kwitaechong can you test this locally?

junghans (approved):

kwitaechong:
@streeve yes, they paased tests locally

junghans (approved):

dalg24 (approved):

junghans:
@streeve all fixed!

@dalg24 can you look why the jenkins build doesn't pull in gtest?

docker/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN apt-get update && apt-get install -y \
         ccache \
         wget \
         openssh-client \
+        libgtest-dev \
```
junghans, line 11: @dalg24 we added the new deps here....

dalg24:
> @streeve all fixed!
> 
> @dalg24 can you look why the jenkins build doesn't pull in gtest?

Because it does not export the configuration (not installed with CMake)

dalg24:
> > @streeve all fixed!
> > @dalg24 can you look why the jenkins build doesn't pull in gtest?
> 
> Because it does not export the configuration (not installed with CMake)

We determined that the base images were Ubuntu 18.04-based and came with a GTest version that was too old.  GTest is being found using the CMake find module since the exported config is not there but the versions check was failing

dalg24 (commented):
https://github.com/ECP-copa/Cabana/blob/20846a8fe58a059740eb31770c6350dae13fad95/.jenkins#L20

The versions are pinned in the jenkinsfile...

CMakeLists.txt
```diff
@@ -187,8 +187,17 @@ MESSAGE(STATUS "Cabana Revision = '${Cabana_GIT_COMMIT_HASH}'")
 # enable unit tests
 option(Cabana_ENABLE_TESTING "Build tests" OFF)
 if(Cabana_ENABLE_TESTING)
+  find_package(GTest 1.10 REQUIRED)
+  # Workaround for FindGTest module in CMake older than 3.20
```
junghans, line 191: Wouldn't it be better to use `if(CMAKE_VERSION VERSION_LESS 3.20)` here?

dalg24: I suppose you mean something like
```CMake
if(CMAKE_VERSION VERSION_LESS 3.20)
  set(gtest_target GTest:GTest)
else()
  set(gtest_target GTest::gtest)
endif()
```
but we still use a variable and I find it a bit more subtle (I wasn't even sure that would work, whether the basic signature first search using export mode or module mode) and it would potentially break if the user set [CMAKE_FIND_PACKAGE_PREFER_CONFIG](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html#variable:CMAKE_FIND_PACKAGE_PREFER_CONFIG).
I don't feel super strongly about it, I can change it if you want me to.

streeve: I'm happy with it as is - are we good to merge?

junghans: @dalg24 another option could be to do:
`
if(CMAKE_VERSION VERSION_LESS 3.20)
  add_library(GTest::gtest ALIAS GTest::GTest)
endif()
```
and then use `GTest::GTest` below.

dalg24: Did you actually try it?  I don't think you can alias an imported target.

junghans: Ah , ok! Thanks!

PR #512, opened 2022-03-28
Allow any integer type for LocalMesh coordinates

streeve:
It is not currently possible to use get `coordinates` with a `sharedIndexSpace` because it returns `long` instead of `int`

Needed for the grid halo for peridynamics (eventually MD as well)


dalg24 (commented):
Do you plan on adding a test for it?

kwitaechong (approved):
looks good

streeve:
> Do you plan on adding a test for it?

Yeah I should for the local grid issue I found at least

abisner (approved):
Looks good with the Integer template parameter.

streeve:
Added a test for the specific issue I was fixing

streeve:
Added a test for the specific issue I was fixed

PR #513, opened 2022-03-29
Make benchmark subdirectories

streeve:
Separate core and Cajita


abisner (approved):
This split looks good.

PR #520, opened 2022-04-11
Fixup reference solver example

streeve:
Boundary values causing occasional convergence failures - blocking #510 right now


kwitaechong (approved):
it seems with this previous implementation, that we don't have error anymore.

PR #525, opened 2022-04-18
Remove unnecessary gtest for performance tests

streeve:
Benchmarks do not use gtest

### DIFF
--- a/benchmark/cajita/CMakeLists.txt
+++ b/benchmark/cajita/CMakeLists.txt
@@ -22,14 +22,11 @@ if(Kokkos_ENABLE_SERIAL OR Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_OPENMP OR Kokkos_
     endif()
 
   if(Cabana_ENABLE_TESTING)
-    target_link_libraries(SparseMapPerformance gtest)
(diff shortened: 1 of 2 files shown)


dalg24 (approved):

benchmark/cajita/CMakeLists.txt
```diff
@@ -22,14 +22,11 @@ if(Kokkos_ENABLE_SERIAL OR Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_OPENMP OR Kokkos_
     endif()
 
   if(Cabana_ENABLE_TESTING)
```
dalg24, line 24: By the way why are we guarding with ` if(Cabana_ENABLE_TESTING)`?

junghans (approved):

benchmark/cajita/CMakeLists.txt
junghans: You mean we should use `BUILD_TESTING` instead?

dalg24: I wouldn't put any condition (so yes rely on `BUILD_TESTING`)

streeve: I usually either want to build the benchmark only or build tests for CI. And we only enable tests with `Cabana_ENABLE_TESTING` on

PR #526, opened 2022-04-18
Deprecate Cajita backwards compatibility wrappers

streeve:
Plan to remove halo and partitioner wrappers in the next release

Spacing is a little odd with the deprecation tags

### DIFF
--- a/cajita/src/Cajita_ManualPartitioner.hpp
+++ b/cajita/src/Cajita_ManualPartitioner.hpp
@@ -28,17 +28,14 @@ namespace Cajita
 /*!
   \brief Backwards compatibility wrapper for 3D ManualBlockPartitioner
 */
-class ManualPartitioner : public ManualBlockPartitioner<3>
-{
-  public:
-    /*!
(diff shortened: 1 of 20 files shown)


cajita/src/Cajita_ManualPartitioner.hpp
```diff
@@ -28,17 +28,14 @@ namespace Cajita
...
-  public:
-    /*!
-      \brief Constructor
-      \param ranks_per_dim MPI ranks per dimension.
-    */
-    ManualPartitioner( const std::array<int, 3>& ranks_per_dim )
-        : ManualBlockPartitioner<3>( ranks_per_dim )
-    {
-    }
+class [[deprecated]] ManualPartitioner : public ManualBlockPartitioner<3>{
```
dalg24, line 31: Format looks wrong

streeve: Agreed, but I think it's the deprecation tag messing it up - not sure if we can fix it easily in the format options

PR #537, opened 2022-06-02
Add THREADS backend build to CI

streeve:
Renames backends in GHA to make this easier

### DIFF
--- a/.github/workflows/CI2.yml
+++ b/.github/workflows/CI2.yml
@@ -39,6 +39,16 @@ jobs:
         silo: ['NoSilo']
         coverage: ['OFF']
         include:
+          - distro: 'ubuntu:latest'
+            cxx: 'g++'
+            backend: 'THREADS'
+            cmake_build_type: 'Debug'
+            kokkos_ver: 'master'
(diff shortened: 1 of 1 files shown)


junghans (approved):

streeve:
Needs fix: https://github.com/kokkos/kokkos/pull/5109 Once that's merged we can run with develop only

dalg24:
How is it related to this CI build?

streeve:
Our Threads failures are not just in CI, but any run with multiple threads with a `parallel_reduce`. @masterleinad mentioned a fix, but maybe I linked the wrong PR

dalg24:
No that would be the one you are correct
But unless you have hwloc or specify the number of threads we were using only one thread

masterleinad:
> Develop now works with my local build of core library only, but now Threads compilation fails for Cajita parallel reductions with tags

I'll have a look.

masterleinad:
https://github.com/kokkos/kokkos/pull/5123 provides a fix.

streeve:
Thanks, @masterleinad

PR #538, opened 2022-06-08
Add heFFTe SYCL support

streeve:
Run in the CI with CPU

### DIFF
--- a/cajita/src/Cajita_FastFourierTransform.hpp
+++ b/cajita/src/Cajita_FastFourierTransform.hpp
@@ -421,6 +421,7 @@ struct HeffteScalingTraits<FFTScaleSymmetric>
     static const auto scaling_type = heffte::scale::symmetric;
 };
 
+#ifdef KOKKOS_ENABLE_SYCL
 // Overload for SYCL.
 template <class HeffteBackendType>
 auto createHeffte(
@@ -437,6 +438,7 @@ auto createHeffte(
     return std::make_shared<heffte::fft3d<HeffteBackendType>>( q, inbox, outbox,
                                                                comm, params );
 }
+#endif
 
 template <class HeffteBackendType>
 auto createHeffte(
--- a/cajita/src/Cajita_FastFourierTransform.hpp
+++ b/cajita/src/Cajita_FastFourierTransform.hpp
@@ -101,7 +101,6 @@ class FastFourierTransformParams
     bool alltoall = true;
     bool pencils = true;
     bool reorder = true;
-    bool queue = false;
 
   public:
     /*!
@@ -121,13 +120,8 @@ class FastFourierTransformParams
     */
     void setReorder( bool value ) { reorder = value; }
     /*!
-      \brief setQueue Set SYCL queue.
-      \param value Use SYCL queue.
-    */
-    void setQueue( bool value ) { queue = value; }
-    /*!
-      \brief getAllToAll Get SYCL queue.
-      \return Using SYCL queue or not.
+      \brief Get MPI communication strategy.
+      \return Using AllToAll or not.
     */
     bool getAllToAll() const { return alltoall; }
     /*!
@@ -141,11 +135,6 @@ class FastFourierTransformParams
       Contiguous layout requires tensor transposition; strided layout does not.
     */
     bool getReorder() const { return reorder; }
-    /*!
-      \brief getQueue Get SYCL queue.
-      \return Using SYCL queue or not.
-    */
-    bool getQueue() const { return queue; }
 };
 
 //---------------------------------------------------------------------------//
@@ -355,6 +344,13 @@ template <class ExecutionSpace, class BackendType>
 struct HeffteBackendTraits
 {
 };
+#ifdef Heffte_ENABLE_MKL
+template <class ExecutionSpace>
+struct HeffteBackendTraits<ExecutionSpace, FFTBackendMKL>
+{
+    using backend_type = heffte::backend::mkl;
+};
+#endif
 #ifdef Heffte_ENABLE_FFTW
 template <class ExecutionSpace>
 struct HeffteBackendTraits<ExecutionSpace, FFTBackendFFTW>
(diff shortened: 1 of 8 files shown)


streeve:
@mkstoyanov I haven't looked into getting this working for a while - should we update the heFFTe version from 2.1 for oneapi support?

mkstoyanov:
I see one old bug that came from an old SYCL update. I had to redefine some kernel templates ... update to 2.1, that problem should be fixed.

streeve:
Thanks @mkstoyanov updating to 2.2 worked. @sfogerty thanks, let's merge #540 before this one

streeve:
@junghans any ideas here on the `unrecognized argument oversubscribe` MPI errors?

junghans:
> @junghans any ideas here on the `unrecognized argument oversubscribe` MPI errors?

It looks like for some reason the mpi coming with the dpcpp doesn't support that argument...

streeve:
@junghans I switched to only run dpcpp with one rank and now it can't even find the exe..

sfogerty:
@streeve I have been having a tough time with SYCL compiles for Cabana - I think the compiler install on our nodes doesn't include some needed tools. (Fails even when I include those tools in $PATH...) 

Here I have pushed an untested fix. Basically, heFFTe recommends using a `sycl::queue` with this backend. (see https://bitbucket.org/icl/heffte/src/master/examples/heffte_example_sycl.cpp)

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -471,11 +481,22 @@ class HeffteFastFourierTransform
...
 
-        // Set FFT options from given parameters
-        // heFFTe correctly handles 2D or 3D FFTs within "fft3d"
-        _fft = std::make_shared<heffte::fft3d<heffte_backend_type>>(
-            inbox, outbox, layout.localGrid()->globalGrid().comm(),
-            heffte_params );
+        // heffte recommends a sycl queue to avoid errors with SYCL backend
+        bool queue = params.getQueue();
+        // TODO: just one call here, no if/else
+        if ( queue )
```
streeve, line 488: This will need to be an ifdef

mkstoyanov: Alternatively, you can pass in a null pntr when not dealing with SYCL:
```
_fft = std::make_shared<heffte::fft3d<heffte_backend_type>>(
              (std::is_same<heffte_backend_type, heffte::backend::onemkl>::value) ?  q : nullptr, 
              inbox, outbox, layout.localGrid()->globalGrid().comm(), heffte_params );
```

mkstoyanov: On another note, why use `shared_ptr` which has more overhead than the simpler `unique_ptr`.

streeve:
@steverangel similar idea here, but different failures with the CPU SYCL build

sfogerty:
@streeve Is it intentional that the CI is enabling both `Heffte_ENABLE_ONEAPI=ON` and `Heffte_ENABLE_MKL=ON` for the  CI matrix value `Heffte_MKL_ONEAPI`? 

In any case, the CI has found an issue in our FFT defaults when both MKL and ONEAPI are enabled. I'm trying to thing how we would want to handle that - which backend we should set as default, or whether we can easily support compiling for both options at the same time.

streeve:
> Is it intentional that the CI is enabling both `Heffte_ENABLE_ONEAPI=ON` and `Heffte_ENABLE_MKL=ON` for the CI matrix value `Heffte_MKL_ONEAPI`?

As I recall the intent was to enable both host and device FFTs if the right backends were enabled. I don't think we've ever tested that or have a clear case that needs it though. If you disable MKL and everything works okay, then that's okay with me. Either way we need to be clear in the documentation

mkstoyanov:
> > Is it intentional that the CI is enabling both `Heffte_ENABLE_ONEAPI=ON` and `Heffte_ENABLE_MKL=ON` for the CI matrix value `Heffte_MKL_ONEAPI`?
> 
> As I recall the intent was to enable both host and device FFTs if the right backends were enabled. I don't think we've ever tested that or have a clear case that needs it though. If you disable MKL and everything works okay, then that's okay with me. Either way we need to be clear in the documentation

Quirks of OneMKL, since SYCL code is not exclusively GPU, you need both the CPU and GPU versions of the MKL to run. There are other dependencies and ultimately, you cannot use heffte OneAPI without also enabling CPU MKL. If you want to pick a default backend, you can do that using type-trait templates:
```
heffte::backend::is_enabled<backend-tag>::value
heffte::backend::uses_gpu<backend-tag>::value
```

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -452,15 +482,38 @@ class HeffteFastFourierTransform
...
+        // heffte recommends a sycl queue to avoid errors with SYCL backend
+        bool queue = params.getQueue();
+        // TODO: just one call here, no if/else
+
+#ifdef Heffte_ENABLE_ONEAPI
+#ifdef KOKKOS_ENABLE_SYCL
+        sycl::queue q;
+        // Set FFT options from given parameters
+        // heFFTe correctly handles 2D or 3D FFTs within "fft3d"
+        _fft = std::make_shared<heffte::fft3d<heffte_backend_type>>(
```
mkstoyanov, line 503: I think the resting environment is trying to compile with one with `heffte_backend_type = mkl`

The issue is that `onemkl` takes a `sycl::queue`, but `mkl` takes either nothing or `nullptr`. This may be the issue.

streeve: I need to take this PR over and get it up to date with master, then I'll try this suggestion. Thanks!

cajita/src/Cajita_FastFourierTransform.hpp
```diff
@@ -407,6 +420,40 @@ struct HeffteScalingTraits<FFTScaleSymmetric>
...
+    HeffteBackendType, heffte::box3d<> inbox, heffte::box3d<> outbox,
+    MPI_Comm comm, heffte::plan_options params,
+    typename std::enable_if<
+        std::is_same<HeffteBackendType, heffte::backend::onemkl>::value,
+        int>::type* = 0 )
+{
+    // Set FFT options from given parameters
+    // heFFTe correctly handles 2D or 3D FFTs within "fft3d"
+    // FIXME: get SYCL queue from Kokkos.
+    sycl::queue q;
```
mkstoyanov, line 437: This is illegal!

The queue has local scope, but heFFTe needs to keep a reference to the queue for all operations. This is needed for synchronization and management of devices on platforms with multiple GPUs. The correct course of action is to use the Kokkos global sycl queue.

streeve: Thanks! Updated locally - @masterleinad am I correct that we need Kokkos >=3.7 in order to get the SYCL queue?

masterleinad: > Thanks! Updated locally - @masterleinad am I correct that we need Kokkos >=3.7 in order to get the SYCL queue?

Yes, for
```C++
sycl::queue& q = space.sycl_queue();
```
but you could do
```C++
 Kokkos::Experimental::Impl::SYCLInternal& instance =
        *space.impl_internal_space_instance();
    sycl::queue& q = *instance.m_queue;
```
(which is of course discouraged) even before.

PR #539, opened 2022-06-09
Enable partial range for Silo output

streeve:
This is important for not outputting ghosted particles.

### DIFF
--- a/core/src/Cabana_SiloParticleOutput.hpp
+++ b/core/src/Cabana_SiloParticleOutput.hpp
@@ -30,6 +30,8 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <Cabana_Slice.hpp>
+
 #include <silo.h>
 
 #include <mpi.h>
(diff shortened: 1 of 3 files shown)


core/src/Cabana_SiloParticleOutput.hpp
```diff
@@ -86,17 +86,18 @@ namespace Impl
...
     // Reorder in a contiguous blocked format.
     Kokkos::View<typename SliceType::value_type*,
                  typename SliceType::device_type>
-        view( "field", slice.size() );
+        view( "field", num_particle );
     Kokkos::parallel_for(
         "Cabana::SiloParticleOutput::writeFieldRank0",
         Kokkos::RangePolicy<typename SliceType::execution_space>(
-            0, slice.size() ),
+            0, num_particle ),
```
kwitaechong, line 100: When there is 10 particle, for example, we can only write from `0` to `6`? Can we also write from `2` to `8`?

streeve: Good point - added your suggestion

kwitaechong (approved):

PR #540, opened 2022-06-09
Update heffte to 2.2.0

sfogerty:
Update heffte version to 2.2.0

needed a slight change to the `alltoall` option. We had that as a bool and heffte now has three options. Future change should allow choosing any of those three options, but for now it's true -> all-to-all, and false -> point-to-point. (heffte also allows p2p_plined)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Checkout heffte
         # actions/checkout doesn't work for external repos yet
         run: |
-          git clone --depth 1 --branch v2.0.0 https://bitbucket.org/icl/heffte.git heffte
+          git clone --depth 1 --branch v2.2.0 https://bitbucket.org/icl/heffte.git heffte
       - name: Checkout hypre
         uses: actions/checkout@v2.2.0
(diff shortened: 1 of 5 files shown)


.github/workflows/docker.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Checkout heffte
         # actions/checkout doesn't work for external repos yet
         run: |
-          git clone --depth 1 --branch v2.0.0 https://bitbucket.org/icl/heffte.git heffte
+          git clone --depth 1 --branch v2.2.0 https://bitbucket.org/icl/heffte.git heffte
```
dalg24, line 37: Was that used at all?  We were requiring v2.1.0 so I am not sure what was going on.

sfogerty: I am not sure that Cabana_REQUIRE_HEFFTE is being set in the CI everywhere it should be

streeve: You may be right, but @junghans any idea why all the Cajita runs are failing on the docker build?

streeve:
Closing in favor of #538

PR #541, opened 2022-06-14
Refactor Load Balancing

squarefk:
The goal of this pull request is to generalize SparseDimPartitioner into more scenarios, to support:
1. Both 2D and 3D
2. SparseMap, DenseMap, ParticlesView
3. Performance comparison with ALL load balancing

### DIFF
--- a/benchmark/cajita/Cajita_DynamicPartitionerPerformance.cpp
+++ b/benchmark/cajita/Cajita_DynamicPartitionerPerformance.cpp
@@ -12,7 +12,7 @@
 #include "../Cabana_BenchmarkUtils.hpp"
 #include "Cabana_ParticleInit.hpp"
 
-#include <Cajita_SparseDimPartitioner.hpp>
+#include <Cajita_DynamicPartitioner.hpp>
 #include <Cajita_SparseIndexSpace.hpp>
 
 #include <Kokkos_Core.hpp>
@@ -142,7 +142,7 @@ void performanceTest( ParticleWorkloadTag, std::ostream& stream, MPI_Comm comm,
         int num_tiles_per_dim = num_cells_per_dim[c] >> cell_bits_per_tile_dim;
 
         // set up partitioner
-        Cajita::SparseDimPartitioner<Device, cell_num_per_tile_dim> partitioner(
+        Cajita::DynamicPartitioner<Device, cell_num_per_tile_dim> partitioner(
             comm, max_workload_coeff, max_par_num, num_step_rebalance,
             global_num_cell, max_optimize_iteration );
         auto ranks_per_dim =
@@ -273,7 +273,7 @@ void performanceTest( SparseMapTag, std::ostream& stream, MPI_Comm comm,
         // set up partitioner
         auto total_num =
             num_tiles_per_dim * num_tiles_per_dim * num_tiles_per_dim;
-        Cajita::SparseDimPartitioner<Device, cell_num_per_tile_dim> partitioner(
+        Cajita::DynamicPartitioner<Device, cell_num_per_tile_dim> partitioner(
             comm, max_workload_coeff, total_num, num_step_rebalance,
             global_num_cell, max_optimize_iteration );
         auto ranks_per_dim =
--- a/benchmark/cajita/Cajita_DynamicPartitionerPerformance.cpp
+++ b/benchmark/cajita/Cajita_DynamicPartitionerPerformance.cpp
@@ -189,9 +189,10 @@ void performanceTest( ParticleWorkloadTag, std::ostream& stream, MPI_Comm comm,
 
                 // compute local workload
                 local_workload_timer.start( p );
-                partitioner.computeLocalWorkLoad( pos_view, par_num,
-                                                  global_low_corner,
-                                                  1.0f / num_cells_per_dim[c] );
+                partitioner.setLocalWorkloadByParticles( pos_view, par_num,
+                                                         global_low_corner,
+                                                         1.0f / num_cells_per_dim[c],
+                                                         comm );
                 local_workload_timer.stop( p );
 
                 // compute prefix sum matrix
@@ -205,8 +206,8 @@ void performanceTest( ParticleWorkloadTag, std::ostream& stream, MPI_Comm comm,
                 total_optimize_timer.start( p );
                 for ( int i = 0; i < max_optimize_iteration; ++i )
                 {
-                    partitioner.optimizePartition( is_changed,
-                                                   std::rand() % 3 );
+                    partitioner.optimizePartitionAlongDim( is_changed,
+                                                           std::rand() % 3 );
                     if ( !is_changed )
                         break;
                 }
@@ -326,7 +327,7 @@ void performanceTest( SparseMapTag, std::ostream& stream, MPI_Comm comm,
 
                 // compute local workload
                 local_workload_timer.start( frac );
-                partitioner.computeLocalWorkLoad( sis );
+                partitioner.setLocalWorkloadBySparseMap( sis, comm );
                 local_workload_timer.stop( frac );
 
                 // compute prefix sum matrix
@@ -340,8 +341,8 @@ void performanceTest( SparseMapTag, std::ostream& stream, MPI_Comm comm,
                 total_optimize_timer.start( frac );
                 for ( int i = 0; i < max_optimize_iteration; ++i )
                 {
-                    partitioner.optimizePartition( is_changed,
-                                                   std::rand() % 3 );
+                    partitioner.optimizePartitionAlongDim( is_changed,
+                                                           std::rand() % 3 );
                     if ( !is_changed )
                         break;
                 }
--- a/cajita/src/Cajita_DynamicPartitioner.hpp
+++ b/cajita/src/Cajita_DynamicPartitioner.hpp
@@ -343,12 +343,13 @@ class DynamicPartitioner : public BlockPartitioner<NumSpaceDim>
       \param global_lower_corner the coordinate of the domain global lower
       corner
       \param dx cell dx size
+      \param comm MPI communicator used for workload reduction
     */
     template <class ParticlePosViewType, typename ArrayType, typename CellUnit>
-    void computeLocalWorkLoad( const ParticlePosViewType& view,
-                               int particle_num,
-                               const ArrayType& global_lower_corner,
-                               const CellUnit dx )
+    void setLocalWorkloadByParticles( const ParticlePosViewType& view,
+                           int particle_num,
+                           const ArrayType& global_lower_corner,
+                           const CellUnit dx, MPI_Comm comm )
     {
         resetWorkload();
         // make a local copy
@@ -375,15 +376,18 @@ class DynamicPartitioner : public BlockPartitioner<NumSpaceDim>
                 Kokkos::atomic_increment( &workload( ti + 1, tj + 1, tz + 1 ) );
             } );
         Kokkos::fence();
+        // Wait for other ranks' workload to be ready
+        MPI_Barrier( comm );
     }
 
     /*!
       \brief compute the workload in the current MPI rank from sparseMap
       (the workload of a tile is 1 if the tile is occupied, 0 otherwise)
       \param sparseMap sparseMap in the current rank
+      \param comm MPI communicator used for workload reduction
     */
     template <class SparseMapType>
-    void computeLocalWorkLoad( const SparseMapType& sparseMap )
+    void setLocalWorkloadBySparseMap( const SparseMapType& sparseMap, MPI_Comm comm )
     {
         resetWorkload();
         // make a local copy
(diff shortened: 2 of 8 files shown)


streeve:
Jenkins is failing because of the benchmark CMakeLists - I can start the other tests after that fix. 

This is a great start! I think we still want some separation of the sparse partitioning from the particle version, maybe a derived class. Particularly because the tile concept does not exist in the dense grid

I like the `DynamicPartitioner` naming - we could eventually rename the existing as `StaticPartitioner` and `ALLDynamicPartitioner`

squarefk:
> Jenkins is failing because of the benchmark CMakeLists - I can start the other tests after that fix.
> 
> This is a great start! I think we still want some separation of the sparse partitioning from the particle version, maybe a derived class. Particularly because the tile concept does not exist in the dense grid
> 
> I like the `DynamicPartitioner` naming - we could eventually rename the existing as `StaticPartitioner` and `ALLDynamicPartitioner`

@streeve  That makes sense. Since particles and sparse map have quite different parameters to pass in the function, I am not sure how the derived class works here to address. Neither constructor does the job I guess.

cajita/unit_test/tstDynamicPartitioner.hpp
```diff
@@ -286,11 +287,9 @@ auto generate_random_particles(
   truth partition )
   \param occupy_num_per_rank the tile number that will be registered on each MPI
   rank
-  \param use_tile2workload indicate the source to compute the workload on MPI
-  ranks, true if using tile occupation while false if using particle positions
 */
-void random_distribution_automatic_rank( int occupy_num_per_rank,
-                                         bool use_tile2workload = true )
+template <bool use_tile2workload>
+void random_distribution_automatic_rank( int occupy_num_per_rank )
```
streeve, line 292: From yesterday, maybe we'll want to just pass in the partitioner here instead

cajita/unit_test/tstDynamicPartitioner.hpp
```diff
@@ -464,11 +475,11 @@ TEST( sparse_dim_partitioner, sparse_dim_partitioner_uniform_test )
 }
 TEST( sparse_dim_partitioner, sparse_dim_partitioner_random_tile_test )
```
streeve, line 476: Since this file only tests the sparse version, let's call it `tstSparseDynamicPartitioner.hpp`. And then also create `tstDynamicPartitioner.hpp` for the dense case

benchmark/cajita/Cajita_DynamicPartitionerPerformance.cpp
```diff
@@ -273,9 +274,10 @@ void performanceTest( SparseMapTag, std::ostream& stream, MPI_Comm comm,
         // set up partitioner
         auto total_num =
             num_tiles_per_dim * num_tiles_per_dim * num_tiles_per_dim;
-        Cajita::SparseDimPartitioner<Device, cell_num_per_tile_dim> partitioner(
-            comm, max_workload_coeff, total_num, num_step_rebalance,
-            global_num_cell, max_optimize_iteration );
+        Cajita::SparseMapDynamicPartitioner<Device, cell_num_per_tile_dim>
```
streeve, line 277: Same as the unit tests, we'll want separate dense and sparse versions here

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -903,6 +762,225 @@ class SparseDimPartitioner : public BlockPartitioner<NumSpaceDim>
...
+  Dynamic mesh block partitioner. (Current Version: Support 3D only) Workload
+  are computed from sparse map occupancy.
+
+  \tparam Device Kokkos device type.
+  \tparam CellPerTileDim Cells per tile per dimension.
+  \tparam NumSpaceDim Dimemsion (The current version support 3D only)
+*/
+template <typename Device, unsigned long long CellPerTileDim = 4,
+          std::size_t NumSpaceDim = 3>
+class SparseMapDynamicPartitioner
```
streeve, line 891: Let's move the sparse version into a separate `Cajita_SparseDynamicPartitioner.hpp`

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -29,14 +29,15 @@ namespace Cajita
...
-  \tparam Device Kokkos device type.
-  \tparam CellPerTileDim Cells per tile per dimension.
-  \tparam NumSpaceDim Dimemsion (The current version support 3D only)
+  Dynamic mesh block partitioner. (Current Version: Support 3D only) There
+  should be no instantiation for this class without implementing any workload
+  computation. \tparam Device Kokkos device type. \tparam CellPerTileDim Cells
+  per tile per dimension. \tparam NumSpaceDim Dimemsion (The current version
+  support 3D only)
 */
 template <typename Device, unsigned long long CellPerTileDim = 4,
```
streeve, line 38: I don't think the `CellPerTileDim` should exist outside the sparse derived version

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -903,6 +762,225 @@ class SparseDimPartitioner : public BlockPartitioner<NumSpaceDim>
...
+      \param comm MPI communicator to decide the rank nums in each dimension
+      \param max_workload_coeff threshold factor for re-partition
+      \param workload_num total workload(particle/tile) number, used to compute
+      workload_threshold
+      \param num_step_rebalance the simulation step number after which one
+      should check if repartition is needed
+      \param global_cells_per_dim 3D array, global cells in each dimension
+      \param max_optimize_iteration max iteration number to run the optimization
+    */
+    ParticleDynamicPartitioner(
```
streeve, line 801: Can simplify with just `using base::base;` since we're not doing anything except calling the base constructor

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -903,6 +762,225 @@ class SparseDimPartitioner : public BlockPartitioner<NumSpaceDim>
...
+      \brief compute the workload in the current MPI rank from particle
+      positions (each particle count for 1 workload value). This function must
+      be called before running optimizePartition() \param view particle
+      positions view \param particle_num total particle number \param
+      global_lower_corner the coordinate of the domain global lower corner
+      \param dx cell dx size
+      \param comm MPI communicator used for workload reduction
+    */
+    template <class ParticlePosViewType, typename ArrayType, typename CellUnit>
+    void setLocalWorkloadByParticles( const ParticlePosViewType& view,
```
streeve, line 846: Just `setLocalWorkload`?

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -29,14 +29,15 @@ namespace Cajita
 {
 //---------------------------------------------------------------------------//
 /*!
-  Sparse mesh block partitioner. (Current Version: Support 3D only)
-  \tparam Device Kokkos device type.
-  \tparam CellPerTileDim Cells per tile per dimension.
-  \tparam NumSpaceDim Dimemsion (The current version support 3D only)
+  Dynamic mesh block partitioner. (Current Version: Support 3D only) There
+  should be no instantiation for this class without implementing any workload
```
streeve, line 33: This should be virtual then, right?

streeve (changes requested):

cajita/src/Cajita_DynamicPartitioner.hpp
squarefk: Do you mean the function `setLocalWorkload()` should be virtual function? However, `ParticleDynamicPartitioner.setLocalWorkload()` and `SparseMapDynamicPartitioner.setLocalWorkload()` have different input parameters now. I have no idea how to deal with it.

cajita/src/Cajita_DynamicPartitioner.hpp
squarefk: There is a class member called `_workload_per_tile` in the base class which the core partitioning algorithm is based on. It records the minimal dividable (unit) area when partitioning.

If we removed the `tile` concept, the whole partitioning algorithm existing in the base class need to be re-implement. My suggestion here is to enforce `CellPerTileDim=1` for the dense case.

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -575,11 +477,11 @@ class SparseDimPartitioner : public BlockPartitioner<NumSpaceDim>
 
     /*!
       \brief optimize the partition in three dimensions seperately
-      \param is_changed label if the partition is changed after the optimization
       \param iter_seed seed number to choose the starting dimension of the
       optimization
+      \param is_changed label if the partition is changed after the optimization
     */
-    void optimizePartition( bool& is_changed, int iter_seed )
+    void optimizePartitionAlongDim( int iter_seed, bool& is_changed )
```
streeve, line 484: `updatePartition`? This does one update in the optimization loop along all dimensions, right?

cajita/src/Cajita_ParticleDynamicPartitioner.hpp
```diff
@@ -0,0 +1,112 @@
...
+        {
+            lower_corner[d] = global_lower_corner[d];
+        }
+
+        Kokkos::parallel_for(
+            "compute_local_workload_parpos",
+            Kokkos::RangePolicy<execution_space>( 0, particle_num ),
+            KOKKOS_LAMBDA( const int i ) {
+                int ti = static_cast<int>(
+                             ( view( i, 0 ) - lower_corner[0] ) / dx - 0.5 ) >>
```
streeve, line 72: We'll either need to make this a class functor or create a local function copies of the member variables - capturing members in a lambda doesn't work in many cases. Either is fine with me

benchmark/cajita/Cajita_ParticleDynamicPartitionerPerformance.cpp
```diff
@@ -0,0 +1,303 @@
...
+#include <set>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include <mpi.h>
+
+//---------------------------------------------------------------------------//
+// Helper functions.
+struct ParticleWorkloadTag
```
streeve, line 33: I think these are unneeded

benchmark/cajita/Cajita_SparseMapDynamicPartitionerPerformance.cpp
```diff
@@ -30,9 +30,6 @@
 
 //---------------------------------------------------------------------------//
 // Helper functions.
-struct ParticleWorkloadTag
-{
-};
 struct SparseMapTag
```
streeve, line 33: Also unneeded

cajita/src/Cajita_SparseMapDynamicPartitioner.hpp
```diff
@@ -0,0 +1,97 @@
...
+namespace Cajita
+{
+
+//---------------------------------------------------------------------------//
+/*!
+  \brief Helper class to set workload for DynamicPartitioner with sparse map.
+  \tparam Sparse map type
+  \tparam Partitioner's device type
+*/
+template <class SparseMapType, typename Device>
```
streeve, line 37: Instead you can extract the device type from the sparse map

cajita/src/Cajita_DynamicPartitioner.hpp
```diff
@@ -27,16 +27,31 @@
...
+
   \tparam Device Kokkos device type.
-  \tparam CellPerTileDim Cells per tile per dimension.
-  \tparam NumSpaceDim Dimemsion (The current version support 3D only)
+  \tparam CellPerTileDim Cells
+  per tile per dimension.
+  \tparam NumSpaceDim Dimemsion (The current version
+  support 3D only)
 */
 template <typename Device, unsigned long long CellPerTileDim = 4,
```
streeve, line 52: This really should be templated on the memory space instead of device type, but I'm hesitant to add another change in this PR

cajita/src/Cajita_SparseMapDynamicPartitioner.hpp
```diff
@@ -0,0 +1,97 @@
...
+                                                 MPI_Comm comm )
+        : sparseMap( sparseMap )
+        , comm( comm )
+    {
+    }
+
+    //! \brief Called by DynamicPartitioner to compute workload
+    void compute( Kokkos::View<int***, memory_space>& workload ) override
+    {
+        const SparseMapType& sparseMap_proxy = sparseMap;
```
streeve, line 63: `sparse_map_copy`

streeve (changes requested):
You can run `make doxygen` from your build to check for missing docs

cajita/src/Cajita_SparseMapDynamicPartitioner.hpp
```diff
@@ -0,0 +1,97 @@
...
+    {
+    }
+
+    //! \brief Called by DynamicPartitioner to compute workload
+    void compute( Kokkos::View<int***, memory_space>& workload ) override
+    {
+        const SparseMapType& sparse_map_copy = sparseMap;
+        Kokkos::parallel_for(
+            "compute_local_workload_sparsmap",
+            Kokkos::RangePolicy<execution_space>( 0, sparseMap.capacity() ),
```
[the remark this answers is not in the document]
streeve: Nope, I'm wrong

PR #548, opened 2022-07-05
Fix arraySize for fully occupied AoSoA

YuxingQiu:
Fix `arraySize` implementation in `Cabana::AoSoA` for cases when all SoA sturctures are fully occupied.
(when size = numSoA * vector_lenght)

### DIFF
--- a/core/src/Cabana_AoSoA.hpp
+++ b/core/src/Cabana_AoSoA.hpp
@@ -442,8 +442,10 @@ class AoSoA
     KOKKOS_INLINE_FUNCTION
     size_type arraySize( const size_type s ) const
     {
-        return ( (size_type)s < _num_soa - 1 ) ? vector_length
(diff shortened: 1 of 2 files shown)


core/src/Cabana_AoSoA.hpp
```diff
@@ -442,8 +442,10 @@ class AoSoA
     KOKKOS_INLINE_FUNCTION
     size_type arraySize( const size_type s ) const
     {
-        return ( (size_type)s < _num_soa - 1 ) ? vector_length
-                                               : ( _size % vector_length );
+        return ( ( (size_type)s < _num_soa - 1 ) ||
```
streeve, line 445: Can you add a comment for this edge case?

YuxingQiu: Sure

streeve (approved):

PR #554, opened 2022-07-13
Test for removing particles outside domain

### LINKED ISSUE #519: Test removing particles with ParticleGridDistributor
Does not currently check for particles outside the global mesh

streeve:
Particle-grid migrate feature naturally removes particles outside the global domain (when non-periodic). This adds a test to explicitly check this

Closes #519

### DIFF
--- a/cajita/unit_test/tstParticleGridDistributor2d.hpp
+++ b/cajita/unit_test/tstParticleGridDistributor2d.hpp
@@ -164,7 +164,7 @@ void migrateTest( const GridType global_grid, const double cell_size,
                 EXPECT_DOUBLE_EQ( coords( p, d ), coords_initial( p, d ) );
         }
     }
-} // namespace Test
+}
 
 //---------------------------------------------------------------------------//
 // The objective of this test is to check how the redistribution works when we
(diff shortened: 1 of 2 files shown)


cajita/unit_test/tstParticleGridDistributor2d.hpp
```diff
@@ -245,6 +245,75 @@ void localOnlyTest( const GridType global_grid, const double cell_size )
...
+    auto owned_cell_space =
+        block->indexSpace( Cajita::Own(), Cajita::Cell(), Cajita::Local() );
+    int num_particle = owned_cell_space.size();
+    using MemberTypes = Cabana::MemberTypes<double[2], int>;
+    using ParticleContainer = Cabana::AoSoA<MemberTypes, Kokkos::HostSpace>;
+    ParticleContainer particles( "particles", num_particle );
+    auto coords = Cabana::slice<0>( particles, "coords" );
+    auto linear_ids = Cabana::slice<1>( particles, "linear_ids" );
+
+    // Put particles in the center of every local cell.
```
sslattery, line 271: This is a confusing way to set up this test. Why not just make a single particle on every rank that is outside the global mesh bounds?

cajita/unit_test/tstParticleGridDistributor3d.hpp
```diff
@@ -262,6 +262,90 @@ void localOnlyTest( const GridType global_grid, const double cell_size )
...
+    auto owned_cell_space =
+        block->indexSpace( Cajita::Own(), Cajita::Cell(), Cajita::Local() );
+    int num_particle = owned_cell_space.size();
+    using MemberTypes = Cabana::MemberTypes<double[3], int>;
+    using ParticleContainer = Cabana::AoSoA<MemberTypes, Kokkos::HostSpace>;
+    ParticleContainer particles( "particles", num_particle );
+    auto coords = Cabana::slice<0>( particles, "coords" );
+    auto linear_ids = Cabana::slice<1>( particles, "linear_ids" );
+
+    // Put particles in the center of every local cell.
```
sslattery, line 288: Same comment here on setup. I think this can be simplified dramatically and still achieve the same effect.

sslattery (changes requested):

cajita/unit_test/tstParticleGridDistributor2d.hpp
```diff
@@ -245,6 +245,72 @@ void localOnlyTest( const GridType global_grid, const double cell_size )
...
+            bool high_i = global_grid->onHighBoundary( Dim::I );
+            bool high_j = global_grid->onHighBoundary( Dim::J );
+            if ( low_i || low_j || high_i || high_j )
+            {
+                double i_corner = local_mesh.lowCorner( Cajita::Own(), Dim::I );
+                double i_extent = local_mesh.extent( Cajita::Own(), Dim::I );
+                double j_corner = local_mesh.lowCorner( Cajita::Own(), Dim::J );
+                double j_extent = local_mesh.extent( Cajita::Own(), Dim::J );
+                if ( low_i )
+                    coords( pid, Dim::I ) = i_corner - 2 * i_extent;
```
sslattery, line 287: You could avoid having to do this by looping over the ghosted cell space. If the domain is on a low boundary, use the low corner coordinate for the particle, if it is on a high boundary, use the high corner coordinate for the particle where the corner coordinates are for the ghosted local mesh. That avoids this logic and you only make a few particles on the domain corners.

streeve: Yeah I agree. I used the boundary indices to create the particles and I think it's more clear

streeve:
Created boundary particles (with slice plane)
boundary

streeve:
Ignoring HIP and SYCL failures as unrelated

PR #556, opened 2022-07-14
Add particle list with field tags

streeve:
Wraps the AoSoA with more user friendly type tags and enables a thread-level particle interface. Adapted from Picasso

Currently single AoSoA only, with a separate version in Cajita that stores the mesh as well.

Should add an example before merging

### DIFF
--- a/core/src/Cabana_Fields.hpp
+++ b/core/src/Cabana_Fields.hpp
@@ -1,0 +1,100 @@
+/****************************************************************************
+ * Copyright (c) 2018-2022 by the Cabana authors                            *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef CABANA_FIELDS_HPP
+#define CABANA_FIELDS_HPP
+
+namespace Cabana
+{
+namespace Field
+{
+//---------------------------------------------------------------------------//
+// General field types.
+//---------------------------------------------------------------------------//
+// Forward declarations.
+template <class T>
+struct Scalar;
+
+template <class T, int D>
+struct Vector;
+
+template <class T, int D0, int D1>
+struct Matrix;
+
+template <class T>
+struct Scalar
+{
+    using value_type = T;
+    static constexpr int rank = 0;
+    static constexpr int size = 1;
+    using data_type = value_type;
+};
+
+template <class T, int D>
+struct Vector
+{
+    using value_type = T;
+    static constexpr int rank = 1;
+    static constexpr int size = D;
+    static constexpr int dim0 = D;
+    using data_type = value_type[D];
+};
+
+template <class T, int D0, int D1>
+struct Matrix
+{
+    using value_type = T;
+    static constexpr int rank = 2;
+    static constexpr int size = D0 * D1;
+    static constexpr int dim0 = D0;
+    static constexpr int dim1 = D1;
+    using data_type = value_type[D0][D1];
+};
+
+//---------------------------------------------------------------------------//
+// Common field types.
+//---------------------------------------------------------------------------//
+template <std::size_t NumSpaceDim>
+struct Position : Vector<double, NumSpaceDim>
+{
+    static std::string label() { return "position"; }
+};
+
+} // namespace Field
+
+//---------------------------------------------------------------------------//
+// General type indexer.
+//---------------------------------------------------------------------------//
+template <class T, int Size, int N, class Type, class... Types>
+struct TypeIndexerImpl
+{
+    static constexpr std::size_t value =
+        TypeIndexerImpl<T, Size, N - 1, Types...>::value *
+        ( std::is_same<T, Type>::value ? Size - 1 - N : 1 );
+};
+
+template <class T, int Size, class Type, class... Types>
+struct TypeIndexerImpl<T, Size, 0, Type, Types...>
+{
+    static constexpr std::size_t value =
+        std::is_same<T, Type>::value ? Size - 1 : 1;
+};
+
+template <class T, class... Types>
+struct TypeIndexer
+{
+    static constexpr std::size_t index =
+        TypeIndexerImpl<T, sizeof...( Types ), sizeof...( Types ) - 1,
+                        Types...>::value;
+};
+
+} // namespace Cabana
+#endif
--- a/core/src/Cabana_Fields.hpp
+++ b/core/src/Cabana_Fields.hpp
@@ -9,6 +9,10 @@
  * SPDX-License-Identifier: BSD-3-Clause                                    *
  ****************************************************************************/
 
+/*!
+  \file Cabana_Fields.hpp
+  \brief Particle field types and common field examples.
+*/
 #ifndef CABANA_FIELDS_HPP
 #define CABANA_FIELDS_HPP
 
@@ -20,6 +24,7 @@ namespace Field
 // General field types.
 //---------------------------------------------------------------------------//
 // Forward declarations.
+//! \cond Impl
 template <class T>
 struct Scalar;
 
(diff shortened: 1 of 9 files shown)


core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,249 @@
...
+    typename Particle<FieldTags...>::tuple_type::template member_reference_type<
+        TypeIndexer<FieldTag, FieldTags...>::index>>::type
+get( Particle<FieldTags...>& particle, FieldTag, IndexTypes... indices )
+{
+    return Cabana::get<TypeIndexer<FieldTag, FieldTags...>::index>(
+        particle.tuple(), indices... );
+}
+
+//---------------------------------------------------------------------------//
+//! Return a single SoA for a single particle.
```
sslattery, line 145: This returns a single element of a single field via `indices`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,249 @@
...
+    typename Particle<FieldTags...>::tuple_type::
+        template member_const_reference_type<
+            TypeIndexer<FieldTag, FieldTags...>::index>>::type
+get( const Particle<FieldTags...>& particle, FieldTag, IndexTypes... indices )
+{
+    return Cabana::get<TypeIndexer<FieldTag, FieldTags...>::index>(
+        particle.tuple(), indices... );
+}
+
+//! Return a single field for a single particle.
```
sslattery, line 132: This returns a single element of a single field via `indices`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,249 @@
...
+        template member_const_reference_type<
+            TypeIndexer<FieldTag, FieldTags...>::index>>::type
+get( const ParticleView<VectorLength, FieldTags...>& particle, FieldTag,
+     IndexTypes... indices )
+{
+    return Cabana::get<TypeIndexer<FieldTag, FieldTags...>::index>(
+        particle.soa(), particle.vectorIndex(), indices... );
+}
+
+//! Return a single SoA for a single particle.
```
sslattery, line 160: This returns a single element of a single field via `indices`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,249 @@
...
+    soa_type& _soa;
+
+    //! The local vector index of the particle.
+    int _vector_index;
+};
+
+//---------------------------------------------------------------------------//
+// Particle accessors.
+//---------------------------------------------------------------------------//
+//! Return a single field for a single particle.
```
sslattery, line 119: This returns a single element of a single field via `indices`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,249 @@
...
+    //! Get the number of particles in the list.
+    std::size_t size() const { return _aosoa.size(); }
+
+    //! Get the AoSoA.
+    aosoa_type& aosoa() { return _aosoa; }
+    //! Get the AoSoA (const).
+    const aosoa_type& aosoa() const { return _aosoa; }
+
+    //! Get the label.
+    const std::string& label() { return _label; }
```
sslattery, line 218: should only need to `const` version of this function because you always return a `const &`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,249 @@
...
+        : _aosoa( label )
+        , _label( label )
+    {
+    }
+
+    //! Get the number of particles in the list.
+    std::size_t size() const { return _aosoa.size(); }
+
+    //! Get the AoSoA.
+    aosoa_type& aosoa() { return _aosoa; }
```
sslattery, line 213: Did I have some functions that also allowed one to get a `Particle` or `ParticleView` on device given a 1D particle index?

core/unit_test/tstParticleList.hpp
```diff
@@ -0,0 +1,227 @@
...
+        for ( int i = 0; i < 3; ++i )
+            for ( int j = 0; j < 3; ++j )
+                EXPECT_DOUBLE_EQ( pf_h( p, i, j ), -1.2 );
+    }
+
+    // Locally modify.
+    Kokkos::parallel_for(
+        "modify", Kokkos::RangePolicy<TEST_EXECSPACE>( 0, num_p ),
+        KOKKOS_LAMBDA( const int p ) {
+            typename list_type::particle_type particle( aosoa.getTuple( p ) );
```
sslattery, line 91: Yeah I'd like some helper functions to do this on device. We should be able to pass the particle list to the device and get the AoSoA out of it there.

core/unit_test/tstParticleList.hpp
```diff
@@ -0,0 +1,227 @@
...
+        for ( int i = 0; i < 3; ++i )
+            for ( int j = 0; j < 3; ++j )
+                EXPECT_DOUBLE_EQ( pf_h( p, i, j ), -1.2 );
+    }
+
+    // Locally modify.
+    Kokkos::parallel_for(
+        "modify", Kokkos::RangePolicy<TEST_EXECSPACE>( 0, num_p ),
+        KOKKOS_LAMBDA( const int p ) {
+            auto s = Cabana::Impl::Index<
```
sslattery, line 182: Definitely want a helper function for this

sslattery (changes requested):

core/src/Cabana_Fields.hpp
```diff
@@ -0,0 +1,167 @@
...
+    //! Scalar type.
+    using data_type = value_type[D0][D1];
+};
+
+//---------------------------------------------------------------------------//
+// Common field types.
+//---------------------------------------------------------------------------//
+//! Particle position field type.
+template <std::size_t NumSpaceDim>
+struct Position : Vector<double, NumSpaceDim>
```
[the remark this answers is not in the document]
streeve: Removed most, but left `Position` to discuss if we want to use it in algorithms

core/src/Cabana_ParticleList.hpp
streeve: Fixed

core/src/Cabana_ParticleList.hpp
streeve: Fixed

core/src/Cabana_ParticleList.hpp
streeve: I added another, but maybe `get`/`set` is confusing with the existing free functions

streeve: Discussed in the next comments

core/unit_test/tstParticleList.hpp
streeve: I added these, but maybe `get`/`set` is confusing with the existing free functions

core/unit_test/tstParticleList.hpp
streeve: I added, but maybe `access` isn't the best name

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,270 @@
...
+    //! Get the AoSoA (const).
+    const aosoa_type& aosoa() const { return _aosoa; }
+
+    //! Get a single particle.
+    template <class IndexType>
+    KOKKOS_INLINE_FUNCTION auto get( IndexType p ) const
```
sslattery, line 219: let's call this `getParticle()`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,270 @@
...
+    {
+        return particle_type( _aosoa.getTuple( p ) );
+    }
+
+    //! Set a single particle.
+    template <class ParticleType, class IndexType>
+    KOKKOS_INLINE_FUNCTION void set( ParticleType particle, IndexType p ) const
```
sslattery, line 226: lets call this `setParticle()`

core/src/Cabana_ParticleList.hpp
```diff
@@ -0,0 +1,270 @@
...
+    {
+        _aosoa.setTuple( p, particle.tuple() );
+    }
+
+    //! Set a single particle with the struct+array indexing.
+    template <class IndexType>
+    KOKKOS_INLINE_FUNCTION auto access( IndexType p ) const
```
sslattery, line 233: lets call this `getParticleView()`

sslattery (changes requested):
Just rename those functions and this is good to merge.

streeve:
@sslattery renamed functions, cleaned up history, and removed the internal field examples like you previously suggested (we can add if we ever want them in an internal algorithm)

Tested with HIP manually

sslattery (approved):

Issue #460, opened 2021-10-14
Neighbor list roadmap


streeve:
Driven by discussion with about the HACCabana proxy app, we need new functionality for improved neighbor list performance:

`LinkedCellList` #712 (replaced #557)
- [x] `NeighborList` interface
- [x] new `neighbor_parallel_for`
- [x] new `linkedcell_parallel`

Add particle interaction functor overload to neighbor lists to compute directly rather than store neighbors:
- [ ] `VerletList`
- [ ] ArborX list

Add distributed neighbor search
- [ ] ArborX

PR #557, opened 2022-07-14
WIP: Add LinkedCellList parallel_for

streeve:
Add an interface for looping through a linked cell list as if it were a neighbor list. Part of #460. Still WIP:

- [ ] Probably makes more sense to rename `neighbor_parallel_for`
- [ ] Would prefer to store a stencil inside the LCL instead of passing them separately 
- [ ] Started adding the neighbor list traits, but not using them yet
- [ ] Needs more robust testing

### DIFF
--- a/core/src/Cabana_LinkedCellList.hpp
+++ b/core/src/Cabana_LinkedCellList.hpp
@@ -16,6 +16,7 @@
 #ifndef CABANA_LINKEDCELLLIST_HPP
 #define CABANA_LINKEDCELLLIST_HPP
 
+#include <Cabana_NeighborList.hpp>
 #include <Cabana_Slice.hpp>
 #include <Cabana_Sort.hpp>
 #include <impl/Cabana_CartesianGrid.hpp>
@@ -115,6 +116,10 @@ class LinkedCellList
         allocate( totalBins(), end - begin );
         build( positions, begin, end );
     }
+
+    //! Number of binned particles.
+    KOKKOS_INLINE_FUNCTION
+    int numParticles() const { return _permutes.extent( 0 ); }
 
     /*!
       \brief Get the total number of bins.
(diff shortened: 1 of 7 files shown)


core/src/Cabana_Parallel.hpp
```diff
@@ -1059,6 +1061,217 @@ for_each_neighbor( const IndexType i, const TeamMemberType team,
...
+
+  In the above example, \c Index is a Cabana index to a given AoSoA element
+  for a particle and its neighbor. Its <tt>operator()</tt> method defines the
+  operation to parallelize, over the range of indices
+  <tt>idx=[begin,end]</tt>.  This compares to a single iteration \c idx of a
+  \c for loop.
+*/
+template <class FunctorType, class LinkedCellType, class StencilType,
+          class... ExecParameters>
+inline void linkedcell_parallel_for(
```
sslattery, line 1109: I would just use the same `neighbor_parallel_for` interface here and overload on a `NeighborList<LinkedCellList>`

core/src/Cabana_Parallel.hpp
```diff
@@ -1059,6 +1061,217 @@ for_each_neighbor( const IndexType i, const TeamMemberType team,
...
+
+    using execution_space =
+        typename Kokkos::RangePolicy<ExecParameters...>::execution_space;
+
+    using index_type =
+        typename Kokkos::RangePolicy<ExecParameters...>::index_type;
+
+    using memory_space = typename LinkedCellType::device_type::memory_space;
+
+    NeighborDiscriminator<FullNeighborTag> discriminator;
```
sslattery, line 1125: Does the half list ever make sense in this context or always just full?

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -410,6 +439,115 @@ void permute(
...
+                {
+                    total_count += list.binSize( i, j, k );
+                }
+        return total_count;
+    }
+
+    //! Get the id for a neighbor for a given particle index and the index of
+    //! the neighbor relative to the particle.
+    KOKKOS_INLINE_FUNCTION
+    static std::size_t getNeighbor( const list_type& list, const std::size_t,
```
sslattery, line 541: We are leaving it up to the user to determine if this neighbor is in the radius right? The way this is structured, a particle is a neighbor if it is in the bin of a given particle or in an adjacent bin.

core/src/Cabana_Parallel.hpp
```diff
@@ -1059,6 +1061,217 @@ for_each_neighbor( const IndexType i, const TeamMemberType team,
...
+        cell_stencil.getCells( cell( i ), imin, imax, jmin, jmax, kmin, kmax );
+
+        // Loop over the cell stencil.
+        for ( int gi = imin; gi < imax; ++gi )
+            for ( int gj = jmin; gj < jmax; ++gj )
+                for ( int gk = kmin; gk < kmax; ++gk )
+                {
+                    // Check the particles in this bin to see if they are
+                    // neighbors.
+                    auto offset = list.binOffset( gi, gj, gk );
```
sslattery, line 1147: I think you could avoid all this if you use at the neighbor list interface directly. You can still call the discriminator but I think we would honestly want to leave that up to user still because this could be used for MD, etc. Based on that I'm not sure you need this overload at all.

sslattery (changes requested):

core/src/Cabana_LinkedCellList.hpp
streeve: Right, this requires a user to ignore based on radius, but we should consider adding new discriminators

core/src/Cabana_Parallel.hpp
streeve: Agreed

core/src/Cabana_Parallel.hpp
streeve: I think based on the way you wrote the half discriminator it could work, but I would prefer to leave that until last

core/src/Cabana_Parallel.hpp
streeve: Planning to test this and compare performance (doing all this extra indexing for each thread may be a lot)

streeve:
Replaced by #712 - @lebuller I commented here to make sure we address everything Stuart mentioned

PR #562, opened 2022-08-10
Silo writer multi-block writer fix

abisner:
Fixed silo writer output for multi-block datasets with a custom prefix string. Remove hard-coded "particles_" naming.


streeve (approved):
So we just weren't using the existing prefix correctly?

PR #575, opened 2022-09-19
Generate fully random particles in benchmarks

streeve:
Creating systems with minimum distance between particles is extremely slow as compared to running the actual benchmarks. Performance profiles with or without minimum separation are extremely similar


abisner (approved):

streeve:
HIP failure unrelated - machines are down right now

PR #580, opened 2022-10-19
Add Kokkos profiling regions

streeve:
Focus on parallel iteration and communication first; should also add to main algorithms

### DIFF
--- a/core/src/Cabana_Parallel.hpp
+++ b/core/src/Cabana_Parallel.hpp
@@ -172,8 +172,12 @@ inline void simd_parallel_for(
     const SimdPolicy<VectorLength, ExecParameters...>& exec_policy,
     const FunctorType& functor, const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::simd_parallel_for" );
+
     Impl::ParallelFor<SimdPolicy<VectorLength, ExecParameters...>, FunctorType>(
         str, exec_policy, functor );
+
+    Kokkos::Profiling::popRegion();
 }
 
 //---------------------------------------------------------------------------//
@@ -249,6 +253,8 @@ inline void neighbor_parallel_for(
     const FunctorType& functor, const NeighborListType& list,
     const FirstNeighborsTag, const SerialOpTag, const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
+
     using work_tag = typename Kokkos::RangePolicy<ExecParameters...>::work_tag;
 
     using execution_space =
@@ -281,6 +287,8 @@ inline void neighbor_parallel_for(
         Kokkos::parallel_for( linear_exec_policy, neigh_func );
     else
         Kokkos::parallel_for( str, linear_exec_policy, neigh_func );
+
+    Kokkos::Profiling::popRegion();
 }
 
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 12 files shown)


core/src/Cabana_Parallel.hpp
```diff
@@ -249,6 +253,8 @@ inline void neighbor_parallel_for(
     const FunctorType& functor, const NeighborListType& list,
     const FirstNeighborsTag, const SerialOpTag, const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
```
dalg24, line 256: Are these `neighbor_parallel_for ` regions really useful?  These seem to only a single kernel launch with nothing other than construct the policy.  Did you look at some profiles that made you think that we should add them?
A possible alternative would be to give a default value not empty to the `str` argument.  Presumably when provided by the user that label is good enough and we don't really need the region.

streeve: That's a fair question. It is more useful for the Cabana algorithms/communication, but I do like showing where Cabana is being used (even when it effectively duplicates the user kernel in the profile)

abisner:
Can we add regions for `Cabana::binByKey`?

abisner (commented):
Can we add regions for `Cabana::binByKey()`? It is used downstream.

streeve:
@abisner added for `binByKey`. I left the (simd/neighbor/grid) parallel regions for now, but can switch to appending-to or replacing-empty parallel labels if you prefer as well

core/src/Cabana_Parallel.hpp
```diff
@@ -684,6 +712,8 @@ inline void neighbor_parallel_reduce(
     const SecondNeighborsTag, const SerialOpTag, ReduceType& reduce_val,
     const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
```
abisner, line 715: `Cabana::neighbor_parallel_reduce`

core/src/Cabana_Parallel.hpp
```diff
@@ -757,6 +789,8 @@ inline void neighbor_parallel_reduce(
     const FirstNeighborsTag, const TeamOpTag, ReduceType& reduce_val,
     const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
```
abisner, line 792: `Cabana::neighbor_parallel_reduce`

core/src/Cabana_Parallel.hpp
```diff
@@ -834,6 +870,8 @@ inline void neighbor_parallel_reduce(
     const SecondNeighborsTag, const TeamOpTag, ReduceType& reduce_val,
     const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
```
abisner, line 873: `Cabana::neighbor_parallel_reduce`

core/src/Cabana_Parallel.hpp
```diff
@@ -916,6 +956,8 @@ inline void neighbor_parallel_reduce(
     const SecondNeighborsTag, const TeamVectorOpTag, ReduceType& reduce_val,
     const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
```
abisner, line 959: `Cabana::neighbor_parallel_reduce`

abisner (commented):
A few naming typos in Cabana_Parallel regions.

core/src/Cabana_Parallel.hpp
```diff
@@ -618,6 +642,8 @@ inline void neighbor_parallel_reduce(
     const FirstNeighborsTag, const SerialOpTag, ReduceType& reduce_val,
     const std::string& str = "" )
 {
+    Kokkos::Profiling::pushRegion( "Cabana::neighbor_parallel_for" );
```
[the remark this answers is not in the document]
streeve: fixed

abisner (approved):

PR #581, opened 2022-10-19
Only include Silo header if MPI enabled

streeve:
Add CI build without MPI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -165,16 +165,14 @@ jobs:
             heffte: 'NoheFFTe'
             hypre: 'NoHYPRE'
             coverage: 'OFF'
-          - distro: 'ubuntu:latest'
+          - distro: 'fedora:nompi'
             cxx: 'g++'
             backend: 'OPENMP'
             cmake_build_type: 'Debug'
             kokkos_ver: '3.5.00'
             arborx: 'ArborX'
-            silo: 'Silo'
             heffte: 'NoheFFTe'
             hypre: 'NoHYPRE'
-            mpi: 'NoMPI'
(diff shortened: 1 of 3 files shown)


junghans:
I think we should create a container without mpi for that! I found cases before where we include mpi.h unconditionally!

streeve:
Makes sense - reopening only to make sure we get this fix in relatively soon

junghans:
> I think we should create a container without mpi for that! I found cases before where we include mpi.h unconditionally!

See https://github.com/ECP-copa/ci-containers/pull/15

.github/workflows/CI.yml
```diff
@@ -165,6 +165,17 @@ jobs:
             heffte: 'NoheFFTe'
             hypre: 'NoHYPRE'
             coverage: 'OFF'
+          - distro: 'ubuntu:latest'
```
junghans, line 168: With https://github.com/ECP-copa/ci-containers/pull/15 merged, we could use
```suggestion
          - distro: 'fedora:nompi'
```
here

.github/workflows/CI.yml
```diff
@@ -365,7 +380,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
             -DVALGRIND_EXECUTABLE=False \
             -DDOXYGEN_WARN_AS_ERROR=FAIL_ON_WARNINGS \
-            ${cabana_cmake_opts}
+            ${cabana_cmake_opts[@]}
```
junghans, line 383: Good catch!

streeve:
@junghans this should be ready

PR #582, opened 2022-10-20
CI: allow extra cmake args in Nightly

junghans:


streeve (approved):

streeve:
Cuda and HIP failures unrelated

PR #584, opened 2022-10-27
Run valgrind in CI instead of nightly

streeve:
Part of #412


streeve:
HIP failure unrelated

PR #585, opened 2022-10-27
Convert class templates to Kokkos memory spaces

### LINKED ISSUE #407: Separate memory and execution spaces
Wherever possible, template classes on `MemorySpace` and functions on `ExecutionSpace`. 

This was being done inconsistently for `VerletList`, but should be improved elsewhere:

- [x] `VerletList` #406
- [x] `LinkedCellList`
- [x] `Sort`
- [x] `CommunicationPlan`, `Halo`, `Distributor`
- [x] AoSoA
- [x] Cajita LocalMesh, Array, FFT, Interpolation, StructuredSolver

streeve:
Convert all class template parameters to memory space rather than device type to close #407. 

- Convert classes
- Add explicit execution space class methods (pass default exec space for existing methods)
- Default device typedefs deprecated. 
- Deprecation warning emitted when device type passed
- Examples and tests updated accordingly (Note only the comm plan used the device type explicitly in tests - the rest already used memory spaces)

### DIFF
--- a/cajita/src/Cajita_BovWriter.hpp
+++ b/cajita/src/Cajita_BovWriter.hpp
@@ -175,9 +175,10 @@ reorderView( TargetView& target, const SourceView& source,
   \param gather_array Gather the array before writing to make parallel
   consistent.
 */
-template <class Array_t>
-void writeTimeStep( const int time_step_index, const double time,
-                    const Array_t& array, const bool gather_array = true )
+template <class ExecutionSpace, class Array_t>
+void writeTimeStep( ExecutionSpace, const int time_step_index,
+                    const double time, const Array_t& array,
+                    const bool gather_array = true )
 {
     static_assert( isUniformMesh<typename Array_t::mesh_type>::value,
                    "ViSIT BOV writer can only be used with uniform mesh" );
@@ -185,8 +186,7 @@ void writeTimeStep( const int time_step_index, const double time,
     // Types
     using entity_type = typename Array_t::entity_type;
     using value_type = typename Array_t::value_type;
-    using device_type = typename Array_t::device_type;
-    using execution_space = typename device_type::execution_space;
+    using memory_space = typename Array_t::memory_space;
     const std::size_t num_space_dim = Array_t::num_space_dim;
 
     // Get the global grid.
@@ -244,7 +244,7 @@ void writeTimeStep( const int time_step_index, const double time,
             {
                 auto halo =
                     createHalo( NodeHaloPattern<num_space_dim>(), 0, array );
-                halo->gather( execution_space(), array );
+                halo->gather( ExecutionSpace(), array );
                 break;
             }
         }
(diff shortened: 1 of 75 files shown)


core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -588,9 +588,10 @@ class CommunicationPlan
       this rank, just use this rank as the export destination and the data
       will be efficiently migrated.
     */
-    template <class ViewType>
-    Kokkos::View<size_type*, device_type>
-    createFromExportsAndTopology( const ViewType& element_export_ranks,
+    template <class ExecutionSpace, class ViewType>
+    Kokkos::View<size_type*, memory_space>
+    createFromExportsAndTopology( ExecutionSpace,
```
dalg24, line 593: When we refactor further, we need to use this execution space to construct execution policies

dalg24 (approved):

core/src/Cabana_CommunicationPlan.hpp
dalg24, line 593:
```C++
static_assert( is_accessible_from<memory_space, execution_space>{}, "" );

Comment (streeve) [reply core/src/Cabana_CommunicationPlan.hpp]: Added throughout

Comment (streeve) [reply core/src/Cabana_CommunicationPlan.hpp]: Created an issue

File: core/src/Cabana_AoSoA.hpp [hunk 2]
```diff
@@ -131,14 +132,14 @@ class AoSoA
...
-    using device_type = DeviceType;
-
+    // FIXME: extracting the self type for backwards compatibility with previous
+    // template on DeviceType. Should simply be MemorySpace after next release.
     //! Memory space.
-    using memory_space = typename device_type::memory_space;
-
-    //! Execution space.
-    using execution_space = typename device_type::execution_space;
+    using memory_space = typename MemorySpace::memory_space;
```
streeve, line 138: @dalg24 I wanted to check with you about this new change - without this we lose backwards compatibility with the previous `DeviceType` template because of the `size_type` below (and a few similar issues in other classes)

dalg24: So with this hack the change you propose support both a device type and a memory type?

dalg24: Do you want to attempt to emit a warning if someone is using a device type?

streeve: Yes, since we've had device type in all of our examples up to now it seemed best to allow both. I think we should warn, but don't know a good way to do that

dalg24:
```C++
constexpr bool warn_deprecated( std::false_type ) { return true; }
[[deprecated( "some helpful msg" )]] constexpr bool warn_deprecated( std::true_type ) { return true; }

template<...>
class AoSoA {
...
static_assert(warn_deprecated(Kokkos::is_device<MemorySpace>()));
};

**Review by kwitaechong — APPROVED** [review]
can you please explain why template classes on `Memoryspace` and function on `ExecutiionSpace`?

File: cajita/src/Cajita_BovWriter.hpp [hunk 3]
```diff
@@ -175,18 +175,18 @@ reorderView( TargetView& target, const SourceView& source,
   \param gather_array Gather the array before writing to make parallel
   consistent.
 */
-template <class Array_t>
-void writeTimeStep( const int time_step_index, const double time,
-                    const Array_t& array, const bool gather_array = true )
+template <class ExecutionSpace, class Array_t>
+void writeTimeStep( ExecutionSpace, const int time_step_index,
```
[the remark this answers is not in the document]
streeve: That gives the default execution space for a given memory space, but there are always multiple options (e.g. OpenMP vs Serial on host)

streeve:
Overall, we only need the memory space to store the class Views, then the execution space to execute the member functions (except in a few very specific instances)

streeve:
Failures unrelated

PR #588, opened 2022-11-03
Fixup particle list

streeve:
Overall, these issues stem from not testing the Cajita extension of the core particle list - the initial Caijta version did not compile

- Private class member needs to be protected
- Workaround for device access of string member
~Local grid needs to be stored instead of local mesh~

### DIFF
--- a/core/src/Cabana_ParticleList.hpp
+++ b/core/src/Cabana_ParticleList.hpp
@@ -202,7 +202,6 @@ class ParticleList
     //! Default constructor.
     ParticleList( const std::string& label )
         : _aosoa( label )
-        , _label( label )
     {
     }
 
@@ -238,8 +237,8 @@ class ParticleList
         return particle_view_type( _aosoa.access( s ), a );
     }
 
-    //! Get the label.
-    const std::string& label() const { return _label; }
+    //! Get the AoSoA label.
+    const std::string& label() const { return _aosoa.label(); }
 
     //! Get a slice of a given field.
     template <class FieldTag>
@@ -253,7 +252,6 @@ class ParticleList
   protected:
     //! Particle AoSoA.
     aosoa_type _aosoa;
-    std::string _label;
 };
 
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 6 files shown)


dalg24 (approved):

streeve:
retest this please.

core/src/Cabana_ParticleList.hpp
```diff
@@ -238,8 +237,8 @@ class ParticleList
         return particle_view_type( _aosoa.access( s ), a );
     }
 
-    //! Get the label.
-    const std::string& label() const { return _label; }
```
abisner, line 242: Why the removal of `const` here?

abisner (approved):

core/src/Cabana_ParticleList.hpp
streeve: Mistake in copy paste - fixed

streeve:
@abisner can you take one more look - all I've changed is running the same test for the Cajita variant

abisner (commented):
It might be cleaner to refactor the common setup code for host view and particle list into a common function to be called in both `ParticleListTest` and `ParticleViewTest`.

streeve:
> It might be cleaner to refactor the common setup code for host view and particle list into a common function to be called in both `ParticleListTest` and `ParticleViewTest`.

Fair point. I've found some bigger issues that need to be fixed in order to use it, but I'll add that too

streeve:
> Fair point. I've found some bigger issues that need to be fixed in order to use it

Fixed issue from local grid not storing a memory space (and renamed variables in the Cajita version for clarity)

streeve:
@abisner one more update to get a clean cuda build (stored std::shared_ptr caused issues)

abisner (approved):
Was the issue before that `MeshParticleList` was not trivially copyable to the device in the lambda captures inside `particle_list_unit_test`? Because of the `std::shared_ptr<>`'s?

At any rate, this looks good, and the name change back to just 'Cajita::ParticleList' makes sense.

PR #589, opened 2022-11-09
Zero particle Silo output

streeve:
Add test for empty case and fix.


abisner (approved):

abisner:
DamBreak output seems to work for multiple ranks without warnings.

streeve:
HIP failure unrelated

PR #590, opened 2022-11-10
Adds an HDF5 output option.

brtnfld:
This feature will create an HDF5 file for a timestep and use XDMF (https://xdmf.org/) to standardize the meaning of the HDF5 data for visualization packages.

- [x]     Add an HDF5 reader program
- [x]     Check the HDF5 data in the unit test.
- [x]     Add a github action testing HDF5 feature.
- Post merge: Performance tuning (maybe additional tuning parameters added), at-scale testing needed.


brtnfld:
**TODO**

1. Add an HDF5 reader program
2. Add error checking for the HDF5 feature.
3. Check the HDF5 data in the unit test.
4. Add a github action testing HDF5 feature.

brtnfld:
To enable the feature:

1. Set HDF5_DIR/HDF5_ROOT to the install directory of HDF5.
2. -D Cabana_REQUIRE_HDF5=ON

streeve:
format this please

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,627 @@
...
+              host_view.data() );
+
+    H5Pclose( plist_id );
+    H5Sclose( memspace_id );
+    H5Dclose( dset_id );
+    H5Sclose( filespace_id );
+
+    if ( 0 == comm_rank )
+    {
+        std::string dtype;
```
streeve, line 173: Do you think it's worth helper functions for extracting these?

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,627 @@
...
+    herr_t status;
+
+    offset[0] = (hsize_t)n_offset;
+    count[0] = (hsize_t)n_local;
+    dimsf[0] = (hsize_t)n_global;
+
+    // Reorder in a contiguous blocked format.
+    Kokkos::View<typename SliceType::value_type*,
+                 typename SliceType::device_type>
+        view( "field", slice.size() );
```
streeve, line 136:
```suggestion
        view( Kokkos::ViewAllocateWithoutInitializing( "field" ), slice.size() );
```
This is something we also need to fix in the Silo writer (no need to zero init since we immediately fill)

streeve (changes requested):
I think this is pretty close to ready to merge - @kwitaechong can you test this branch with MPM APIC matrix outputs?

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,621 @@
...
+                coords_view( i, d0 ) = coords_slice( i, d0 );
+        } );
+
+    // Mirror the coordinates to the host.
+    auto host_coords =
+        Kokkos::create_mirror_view_and_copy( Kokkos::HostSpace(), coords_view );
+
+    // Add the point mesh.
+    std::string mesh_name = "particles";
+    double* ptrs[3] = { &host_coords( 0, 0 ), &host_coords( 0, 1 ),
```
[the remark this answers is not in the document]
streeve: See #591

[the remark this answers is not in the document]
streeve: See #589

core/src/Cabana_HDF5ParticleOutput.hpp
streeve: See #591

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 #------------------------------------------------------------------------------#
 cmake_minimum_required(VERSION 3.16)
 
-project(Cabana LANGUAGES CXX)
+project(Cabana LANGUAGES CXX C)
```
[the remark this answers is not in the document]
brtnfld: I think this can be removed. It seems to work fine without it.

core/src/Cabana_HDF5ParticleOutput.hpp
brtnfld: Yes, that would probably be better.

core/src/Cabana_HDF5ParticleOutput.hpp
brtnfld: fixed.

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,621 @@
...
+                        &host_coords( 0, 2 ) };
+
+    int* tmp{ new int[comm_size]{} };
```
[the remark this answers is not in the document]
brtnfld: Actually, ptrs is not needed anymore, I just use host_coords.data.

kwitaechong:
After install hdf5-mpi, I have this compiling error,

`Cabana/core/src/Cabana_HDF5ParticleOutput.hpp:161:9: error: use of undeclared identifier 'H5Pset_dxpl_mpio'
        H5Pset_dxpl_mpio( plist_id, H5FD_MPIO_COLLECTIVE );`

streeve:
We're seeing some strange position output when we test with our proxy app ExaMPM: https://github.com/ECP-copa/ExaMPM. Can you take a look at replacing the Silo calls with HDF5 in one of the examples?

brtnfld:
Are the ExaMPM examples something I can run on a desktop? What command line parameters should I use to reproduce what you are seeing?

I think there are some linking issues with HDF5 in Cabana, I had to change the -lHDF5:hdf5 to just -lhdf5, and I had to set the -L path manually.

kwitaechong:
> Are the ExaMPM examples something I can run on a desktop? What command line parameters should I use to reproduce what you are seeing?
> 
> I think there are some linking issues with HDF5 in Cabana, I had to change the -lHDF5:hdf5 to just -lhdf5, and I had to set the -L path manually.

Yes you can run on desktop. For DamBreak example, `./DamBreak 0.05 2 0 0.001 10.0 50 serial`. As you can see below, left is correct particles for DamBreak by Silo and right is incorrect particles by HDF5

Screen Shot 2022-11-11 at 8 25 50 PM

brtnfld:
Thanks, if I look at the HDF5 file, many of the point coordinates are repeated. Did I misunderstand host_coords.data is not x, y, z, x, y, z,x, y, z .... for each point?

 DATASET "coord_xyz" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 6144, 3 ) / ( 6144, 3 ) }
      DATA {
      (0,0): 0.0125, 0.0375, 0.0125,     --> point 1?
      (1,0): 0.0375, 0.0125, 0.0375,     --> point 2?
      (2,0): 0.0125, 0.0375, 0.0625,
      (3,0): 0.0875, 0.0625, 0.0875,
      (4,0): 0.0625, 0.0875, 0.0625,
      (5,0): 0.0875, 0.1125, 0.1375,
      (6,0): 0.1125, 0.1375, 0.1125,
      (7,0): 0.1375, 0.1125, 0.1375,
      (8,0): 0.1625, 0.1875, 0.1625,
      (9,0): 0.1875, 0.1625, 0.1875,
      (10,0): 0.1625, 0.1875, 0.2125,
      (11,0): 0.2375, 0.2125, 0.2375,
      (12,0): 0.2125, 0.2375, 0.2125,
      (13,0): 0.2375, 0.2625, 0.2875,
      (14,0): 0.2625, 0.2875, 0.2625,
      (15,0): 0.2875, 0.2625, 0.2875,
      (16,0): 0.3125, 0.3375, 0.3125,
      (17,0): 0.3375, 0.3125, 0.3375,
      (18,0): 0.3125, 0.3375, 0.3625,
      (19,0): 0.3875, 0.3625, 0.3875,
      (20,0): 0.3625, 0.3875, 0.3625,
      (21,0): 0.3875, 0.0125, 0.0375,
      (22,0): 0.0125, 0.0375, 0.0125, <--- seems to repeat
      (23,0): 0.0375, 0.0125, 0.0375,
      (24,0): 0.0625, 0.0875, 0.0625,

brtnfld@euler ~ %

brtnfld:
I've verified that the hdf5 and silo coordinates are the same for all the points, with repeated coordinates and such.

streeve:
> I've verified that the hdf5 and silo coordinates are the same for all the points, with repeated coordinates and such. It looks correct in gnuplot too if I plot the points. I'll have to check into the xdmf specs.

Sounds like we need to check on the Silo output too if we have repeated output. For the layouts, we can force row/column ordering in `host_coords` by adding the `Kokkos::LayoutLeft`/`Kokkos::LayoutRight` template parameters

brtnfld:
It's strange because I can create a 2D array with those coordinates, write the 2D array, and it looks fine in ParaView. It is only when I use host_coords that it has an issue, but I can dump the hdf5 file, and the coordinates looks fine.

brtnfld:
ok, so host_coords are storing the coordinates x1,x2,x3,...y1,y2,y3,...z1,z2,z3. Are you saying there is a way to store the coordinates x1,y1, and z1 instead?

streeve:
> ok, so host_coords os storing the coordinates x1,x2,x3,...y1,y2,y3,...z1,z2,z3 in memory. Are you saying there is a way to store the coordinates x1,y1, and z1 instead?

Yes - we can replace all the cases of `LayoutLeft` with `LayoutRight`

kwitaechong:
Yes, it seems works!

Screen Shot 2022-11-15 at 11 32 58 AM

cmake/FindHDF5.cmake
```diff
@@ -0,0 +1,20 @@
+pkg_check_modules(PC_HDF5 hdf5 QUIET)
```
junghans, line 1: Should we instead use the upstream one? https://cmake.org/cmake/help/latest/module/FindHDF5.html

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,649 @@
...
+
+    std::stringstream filename_xdmf;
+    filename_xdmf << prefix << "_" << time_step_index << ".xmf";
+
+    plist_id = H5Pcreate( H5P_FILE_ACCESS );
+    status = H5Pset_fapl_mpio( plist_id, comm, MPI_INFO_NULL );
+
+    status = H5Pset_libver_bounds( plist_id, H5F_LIBVER_LATEST, H5F_LIBVER_LATEST );
+
+#if 0
```
streeve, line 501: Is this for debug?

cmake/FindHDF5.cmake
streeve: We did have issues downstream unless we copied the current added `FindHDF5` at each level (conflicts with the one you linked?)

core/src/Cabana_HDF5ParticleOutput.hpp
brtnfld: I moved this to HDF5Traits.

core/src/Cabana_HDF5ParticleOutput.hpp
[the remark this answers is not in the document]
brtnfld: made tmp vectors.

streeve (commented):
Looks great! Few more fixups here - would you prefer to add the read in a separate PR or add here? And do you want to clean up the commits or just squash merge?

cmake/FindHDF5.cmake
brtnfld: I've never used find_dependency, only find_package. I can implement a CMake schema to find HDF5 that we've had success with for other projects. It is complicated because it matters whether HDF5 was built using CMake or Autotools and whether it was built static/shared.

core/src/Cabana_Core.hpp
```diff
@@ -49,4 +49,13 @@
 #include <Cabana_Experimental_NeighborList.hpp>
 #endif
 
+#ifdef Cabana_ENABLE_SILO
+#include <Cabana_SiloParticleOutput.hpp>
+#endif
+
+#ifdef Cabana_ENABLE_HDF5
+#include <Cabana_HDF5ParticleOutput.hpp>
+#include <xdmf_header_info.h>
```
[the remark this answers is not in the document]
brtnfld: I'll append it. 
I could do a separate PR later for reading if you don't mind there not being a valid check for the I/O. 
I don't need to track the development history of the commits so it would be better to squash merge it.

cmake/FindHDF5.cmake
junghans: @streeve FindHDF5.cmake is part of the CMake Distributions, so if we required a high enough CMake Version, we should just have it.

core/src/Cabana_HDF5ParticleOutput.hpp
brtnfld: It is just a reminder for something to try and see if it should be included as an option.

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,615 @@
...
+        } );
+
+    // Mirror the coordinates to the host.
+    auto host_coords =
+        Kokkos::create_mirror_view_and_copy( Kokkos::HostSpace(), coords_view );
+
+    // Add the point mesh.
+    std::string mesh_name = "particles";
+
+    std::vector<int> tmp( comm_size );
```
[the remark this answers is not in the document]
brtnfld: fixed.

core/src/Cabana_Core.hpp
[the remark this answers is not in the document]
brtnfld: fixed.

cmake/FindHDF5.cmake
brtnfld: I think cmake 3.16 has the find package HDF5 which handles both HDF5 built with CMake and Autotools.

streeve: Perfect - that's already our minimum required. So we should be able to just remove locally?

streeve: @junghans I gave it a shot, but I'm having trouble with C/C++ difference in CMake. The CMake FindHDF5 uses `try_compile`, so I don't know a good way around globally enabling `C` or keeping a custom FindHDF5

brtnfld: Is it possible to set CMAKE_C_STANDARD_REQUIRED ON and then turn it off after find package?

streeve: Looks like it works now specifying CXX only (either required your most recent change or I did it wrong the first time)

brtnfld:
Are there any remaining issues for me to address?

streeve (commented):
I rewrote the read functions to work with Views/slices, but didn't try to add the multiple variadic label+slice. I had a couple more questions on unused variables, but otherwise I think it's ready. @kwitaechong or @abisner can you test out and review?

core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,770 @@
...
+    hid_t plist_id;
+    hid_t dset_id;
+    hid_t filespace_id;
+    hid_t memspace_id;
+
+    // HDF5 hyperslab parameters
+    hsize_t offset[1];
+    hsize_t dimsf[1];
+    hsize_t count[1];
+    herr_t status;
```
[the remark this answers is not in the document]
brtnfld: Since it is a 1D array instead of the 2D array in the main write function, the two will be different.

CMakeLists.txt
```diff
@@ -110,6 +110,9 @@ if(SILO_FOUND)
     DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/Cabana )
 endif()
 
+# find HDF5 (XDMF)
+Cabana_add_dependency( PACKAGE HDF5 COMPONENTS CXX )
```
brtnfld, line 114: CXX should be C. We are not using  CXX bindings.

cmake/CabanaConfig.cmakein
```diff
@@ -42,3 +42,7 @@ set(Cabana_ENABLE_SILO @Cabana_ENABLE_SILO@)
 if(Cabana_ENABLE_SILO)
   find_dependency(SILO @SILO_VERSION@ REQUIRED)
 endif()
+set(Cabana_ENABLE_HDF5 @Cabana_ENABLE_HDF5@)
+if(Cabana_ENABLE_HDF5)
+  find_dependency(HDF5 @HDF5_VERSION@ REQUIRED COMPONENTS C)
```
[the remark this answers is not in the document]
brtnfld: I could not find  HDF5 for an Autotools build of HDF5, at least with CMake 3.20.4. I think we still need to include the legacy find packages alternative.

cmake/FindHDF5.cmake
streeve: (continued in thread above)

cmake/CabanaConfig.cmakein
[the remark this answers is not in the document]
streeve: I'm okay keeping it if we add a note that we should try to remove on new CMake releases that update the internal version

CMakeLists.txt
streeve: fixed

streeve:
> Are there any remaining issues for me to address?

Yeah, I think just the Actions test. We usually add one single run for each TPL and then add to a combined run with all TPL. @junghans would you prefer to install HDF5 in the container?

junghans:
> > Are there any remaining issues for me to address?
> 
> Yeah, I think just the Actions test. We usually add one single run for each TPL and then add to a combined run with all TPL. @junghans would you prefer to install HDF5 in the container?

Yeah, that is what I would do!

brtnfld:
Added attribute for the current timestep for the file and removed unused variables.

streeve:
@junghans confused on the CI output here - HDF5 is found, but still errors right away?

brtnfld:
It looks to me that HDF5 was not built with parallel support.

junghans:
Hmm, no idea. You could run a build with `-DHDF5_FIND_DEBUG=ON` to see what is going on....

cmake/FindHDF5.cmake
```diff
@@ -0,0 +1,20 @@
...
+
+mark_as_advanced(HDF5_INCLUDE_DIR HDF5_LIBRARY)
+
+if(HDF5_INCLUDE_DIR AND HDF5_LIBRARY AND NOT TARGET HDF5::hdf5)
+  add_library(HDF5::hdf5 UNKNOWN IMPORTED)
+  set_target_properties(HDF5::hdf5 PROPERTIES
+    IMPORTED_LOCATION ${HDF5_LIBRARY}
+    INTERFACE_INCLUDE_DIRECTORIES ${HDF5_INCLUDE_DIR})
+ # if(HDF5_LINK_FLAGS)
+    set_property(TARGET HDF5::hdf5 APPEND_STRING PROPERTY INTERFACE_LINK_LIBRARIES "${HDF5_LIBRARY}")
```
junghans, line 17: This one seems wrong to me, isn't that the same as `IMPORTED_LOCATION` ?

cmake/CabanaConfig.cmakein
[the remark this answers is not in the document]
streeve: Just added in case we end up keeping it

streeve:
I was able to build with docker only when explicitly passing the HDF5 include/lib directories (`/usr/include/hdf5/openmpi` and `/usr/lib/x86_64-linux-gnu/hdf5/openmpi/libhdf5.so` for ubuntu). Not sure of a good way to get the right paths in the CI for each distro

brtnfld:
It should find it if you set HDF5_ROOT to the install directory.

streeve:
@junghans @dalg24 @brtnfld adding `enable_language(C)` certainly makes life easier for finding HDF5 - not sure how to disable if not found as we discussed. I don't really like adding `C` to downstream projects that want to use HDF5, but it is an easy change

junghans:
I think we should fill an upstream bug as well.

brtnfld:
I think we also decide to change n_local to size_t (or at least unsigned int).

CMakeLists.txt
```diff
@@ -72,16 +72,23 @@ endif()
...
   option(
     Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION}
     "Require Cabana to build with ${CABANA_DEPENDENCY_PACKAGE} support" ${CABANA_DEPENDENCY_PACKAGE}_FOUND)
   if(Cabana_REQUIRE_${CABANA_DEPENDENCY_OPTION})
-    find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} REQUIRED )
+    find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} REQUIRED COMPONENTS ${CABANA_DEPENDENCY_COMPONENTS} )
   endif()
   set(Cabana_ENABLE_${CABANA_DEPENDENCY_OPTION} ${${CABANA_DEPENDENCY_PACKAGE}_FOUND})
+  if(NOT ${HDF5_FOUND})
+    #disable C
```
streeve, line 90: Not sure how to do this

core/unit_test/tstHDF5ParticleOutput.hpp
```diff
@@ -0,0 +1,187 @@
...
+    checkVector( coords_mirror, coords_read );
+    EXPECT_DOUBLE_EQ( time, time_read );
+}
+
+//---------------------------------------------------------------------------//
+// RUN TESTS
+//---------------------------------------------------------------------------//
+TEST( TEST_CATEGORY, write_read_test ) { writeReadTest(); }
+
+// FIXME: Add test with ghosted particles
```
junghans, line 183: Huh?

junghans:
> I think we should file an upstream bug as well.

https://gitlab.kitware.com/cmake/cmake/-/issues/24241

CMakeLists.txt
```diff
@@ -72,16 +72,23 @@ endif()
 
 # standard dependency macro
 macro(Cabana_add_dependency)
-  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE;VERSION" "" ${ARGN})
-  find_package( ${CABANA_DEPENDENCY_PACKAGE} ${CABANA_DEPENDENCY_VERSION} QUIET )
+  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE;VERSION;COMPONENTS" "" ${ARGN})
+  # Workaround for using HDF5 C-bindings
+  if(${CABANA_DEPENDENCY_PACKAGE} STREQUAL HDF5)
```
junghans, line 77:
```suggestion
  if(CMAKE_VERSION VERSION_LESS 2.26 AND CABANA_DEPENDENCY_PACKAGE STREQUAL "HDF5")
```

CMakeLists.txt
junghans: I don't believe this is possible, so drop this block.

CMakeLists.txt
brtnfld: I talked to Kitware, and one can not disable a language, but he did merge a change to findHDF5 where it does not depend on C being enabled, https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8015/diffs

It should be in the next CMake release, so at that time, you can check the CMake version and not load the C language.

core/unit_test/tstHDF5ParticleOutput.hpp
streeve: Removed - seemed like a good extra test at the time, but now it seems out of scope

.github/workflows/CI.yml
```diff
@@ -347,12 +361,18 @@ jobs:
...
           SILO: ${{ matrix.silo }}
+          HDF5: ${{ matrix.hdf5 }}
         run: |
           if [[ ${SILO} == 'Silo' ]]; then
             cabana_cmake_opts+=( -DCabana_REQUIRE_SILO=ON )
           else
             cabana_cmake_opts+=( -DCMAKE_DISABLE_FIND_PACKAGE_SILO=ON )
           fi
+          if [[ ${HDF5} == 'HDF5' ]]; then
+            cabana_cmake_opts+=( -DCabana_REQUIRE_HDF5=ON -DHDF5_FIND_DEBUG=ON )
```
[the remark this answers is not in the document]
streeve: fixed

CMakeLists.txt
streeve: Removed

streeve:
@junghans I added all your suggestions and also tried to avoid breaking pure CXX downstream apps by not automatically finding HDF5 (will not be found with older CMake unless `Cabana_REQUIRE_HDF5=ON`)

core/src/CMakeLists.txt
```diff
@@ -70,14 +74,19 @@ if(Cabana_ENABLE_SILO)
   target_link_libraries(cabanacore INTERFACE SILO::silo)
 endif()
 
+if(Cabana_ENABLE_HDF5)
+  target_link_libraries(cabanacore INTERFACE ${HDF5_LIBRARIES})
```
junghans, line 78:
```suggestion
  target_link_libraries(cabanacore INTERFACE HDF5::HDF5)
```

core/src/CMakeLists.txt
```diff
@@ -70,14 +74,19 @@ if(Cabana_ENABLE_SILO)
...
 if(Cabana_ENABLE_MPI)
   target_link_libraries(cabanacore INTERFACE MPI::MPI_CXX)
 endif()
 
 target_include_directories(cabanacore INTERFACE
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
-  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  ${HDF5_INCLUDE_DIRS})
```
junghans, line 89:
```suggestion
  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
```

core/src/CMakeLists.txt
streeve: That's what I had originally, but I can't build with this change

junghans: And with `hdf5::hdf5` (the c library) ?

streeve:
Intel builds are failing with `Could NOT find Threads` - I haven't seen this before @junghans

junghans:
> Intel builds are failing with `Could NOT find Threads` - I haven't seen this before @junghans

Hmm, `find_package(Threads)` ?

core/src/CMakeLists.txt
streeve: Looks like this was added/fixed somewhere between CMake 3.18 and 3.25. I can bisect if we want to bump the min for this change

streeve:
Other than the CI issues I think this is ready. @brtnfld are you okay if I clean up the commit history?

brtnfld:
> Other than the CI issues I think this is ready. @brtnfld are you okay if I clean up the commit history?

Yes, I don't think the development history needs to be preserved, so squashing the commits is fine with me.

core/src/CMakeLists.txt
streeve: Added fixme note

core/src/CMakeLists.txt
streeve: Also added fixme

streeve:
@kwitaechong / @abisner can you give final review/test results?

streeve:
retest this please.

streeve:
| | MPI-1 |    MPI-4 |     Speedup(MPI) |
---|---|---|---
| HDF5 | 4.054137 |  1.66753725 | 2.43 |
| Silo | 4.940384 | 3.3220715 | 1.49 |
| Speedup | 1.22  | 1.99 |

(I/O time for 150k particles on my Mac)

streeve (approved):

abisner (approved):

streeve:
| | PPC | MPI-1 |    MPI-6 |     Speedup(MPI) |
---|---|---|---|---
| HDF5 | 1 | 1.966035 | 1.044  | 1.88  |
| Silo | 1 | 1.967855 | 1.6241 | 1.21  |
| Speedup | 1 | 1.0   | 1.56   |  |
| HDF5 | 2 | 7.642659 | 3.605  | 2.12 |
| Silo | 2 | 14.81104 | 3.875  | 3.82 |
| Speedup | 2 | 1.94  | 1.07   |  |

(I/O time for 1.3M/10M particles for 10 steps on Summit V100)

Just quick node level tests - will do more in separate profiling/tuning work. Note this includes data copies and I/O. They're pretty even at ~1M per rank

PR #598, opened 2022-12-08
Require Kokkos 3.6

streeve:
Update Kokkos support - plan to support 2 versions at a time, currently 3.6 and 3.7


dalg24 (approved):

streeve:
@masterleinad I'm missing something here where updating the minimum Kokkos is resulting in MPI issues only in the SYCL build. Seen anything like this?

masterleinad:
> @masterleinad I'm missing something here where updating the minimum Kokkos is resulting in MPI issues only in the SYCL build. Seen anything like this?

You mean
```
[ RUN      ] sycl.comm_plan_test_1
PI CUDA ERROR:
	Value:           717
	Name:            CUDA_ERROR_INVALID_ADDRESS_SPACE
	Description:     operation not supported on global/shared address space
	Function:        cuda_piQueueFinish
	Source Location: /scratch/llvm/sycl/plugins/cuda/pi_cuda.cpp:2244

PI CUDA ERROR
	Value:           717
	Name:            CUDA_ERROR_INVALID_ADDRESS_SPACE
	Description:     operation not supported on global/shared address space
	Function:        cuda_piQueueFinish
	Source Location: /scratch/llvm/sycl/plugins/cuda/pi_cuda.cpp:2244

terminate called after throwing an instance of 'std::runtime_error'
  what():  There was a synchronous SYCL error:
Native API failed. Native API returns: -999 (Unknown OpenCL error code) -999 (Unknown OpenCL error code)
```
? I am seeing similar failures with https://github.com/kokkos/kokkos/pull/5480 bu that one messes with address spaces directly. I'm not quite sure what the problem would be here. It would be interesting to see the backtrace and to know if it works on other `SYCL` backends (Intel GPUs).

streeve:
retest this please.

streeve:
Disabling a subset of failing SYCL tests (not seen on other systems) and ignoring HIP failures

Issue #596, opened 2022-12-02
C++17 updates


abisner:
- [] Remove boilerplate workarounds for fold expressions.
- [] `KOKKOS_CLASS_LAMBDA` for member variable captures.
- [] Remove associated `TODO`'s.

PR #600, opened 2022-12-08
Require C++17

streeve:
In anticipation of requirement in Kokkos 4. Tracking related future updates in #596 

Removes override of CMake standard and instead uses `INTERFACE_COMPILE_FEATURES` to require a minimum standard

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,8 +23,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CXX_FLAGS)
   set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE)
 endif(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CXX_FLAGS)
 
-# Use C++14
-set(CMAKE_CXX_STANDARD 14)
+# Use C++17
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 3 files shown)


core/src/CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ set(HEADERS_IMPL
 
 add_library(cabanacore INTERFACE)
 
+# Require minimum of C++17
+set_target_properties(cabanacore PROPERTIES INTERFACE_COMPILE_FEATURES cxx_std_17)
```
streeve, line 64: As I understood from discussion Cajita will inherit this since it links to the core, but I'm not sure

CMakeLists.txt
```diff
@@ -23,8 +23,6 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CXX_FLAGS)
   set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE)
 endif(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CXX_FLAGS)
 
-# Use C++14
-set(CMAKE_CXX_STANDARD 14)
```
junghans, line 27: we could remove the next two lines as well.

core/src/CMakeLists.txt
junghans: I would still make it explicit!

streeve: Done

CMakeLists.txt
streeve: Done

PR #602, opened 2023-01-09
Intel CI updates

streeve:


abisner (approved):

junghans:
Sorry to be late, but why are we adding `-O0`? `-Wno-debug-disables-optimization` was just used to silence a warning about compiling with `-O0` now we force it to be `-O0`.

streeve:
> Sorry to be late, but why are we adding `-O0`? `-Wno-debug-disables-optimization` was just used to silence a warning about compiling with `-O0` now we force it to be `-O0`.

No that's fair, I was a little quick to merge. That flag we added was somehow causing the build issues in other tests (although very unclear how that manifested as `Threads` not being found in GTest). I assumed it was better to have a consistent intel build with `O0` than nothing

PR #606, opened 2023-02-09
Replace `Kokkos::View::{R -> r}ank` and `Slice::{R->r}ank` and `Slice::{rank->viewRank}()`

dalg24:
Uppercase `Rank` is not documented in Kokkos Core API.

~Even though it is unclear at this time whether it will ever be deprecated or removed, avoid using it to be on the safe side.~
Now deprecated in Kokkos develop

This is a breaking change

### DIFF
--- a/cajita/src/Cajita_IndexSpace.hpp
+++ b/cajita/src/Cajita_IndexSpace.hpp
@@ -354,7 +354,7 @@ KOKKOS_INLINE_FUNCTION auto createSubview( const ViewType& view,
                                            const IndexSpace<1>& index_space )
     -> decltype( Kokkos::subview( view, index_space.range( 0 ) ) )
 {
-    static_assert( 1 == ViewType::Rank, "Incorrect view rank" );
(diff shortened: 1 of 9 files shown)


streeve (approved):

streeve:
This points out a hole in our unit tests - we only use Views (failure was only caught in the performance test)

In `Slice` we have `enum Rank` and `rank()`, but no `enum rank`. What's the plan for Kokkos to transition to a function? Seems like we should take the same route

streeve:
Tested on locally and Frontier with Kokkos 3.6 and develop with no issues. @sslattery can you review?

dalg24:
Looks good to me.  Thanks for dealing with it.

sslattery (approved):
This is good. I like the inclusion of the `viewRank()` function

streeve:
> This is good. I like the inclusion of the `viewRank()` function

I think it's more clear - but just to repeat one more time (mostly for posterity) that breaking change is necessary due to the change of the `rank` member variable

Issue #607, opened 2023-02-17
Enable Views or slices in algorithms


streeve:
Remove the requirement for slices only in algorithms and ensure direct Views work. Some cases should just require modified `enable_if`, others new overloads

- [ ] VerletList
- [ ] LinkedCellList

New overloads
- [ ] CommPlan, Halo, Distributor

PR #608, opened 2023-02-17
Test interpolation with both Views and slices

streeve:
Add interpolation tests with particles in slices as well as Views. Already worked with both (see #607), but previous tests only used Views

Found in #606


streeve:
jenkins failure is a real Cuda issue - not ready to merge

PR #609, opened 2023-02-20
Grid-based particle initialization and random init rewrite

streeve:
- Both uniform and random creation from Cajita grids
- Updates the core library init to make it more similar (type tags and non-uniform box). Behavior changes to ignore particles for the variants with minimum particle separation (and therefore may not return the number of particles requested)
- Deprecates previous simple random init versions (these still loop until all particles have been created which is extremely slow)

~~Core versions should also accept `ParticleList` + user init functors~~ done


streeve:
@abisner this is finally ready for review

streeve:
Now blocked by #654

streeve:
@abisner this actually passes all tests. I removed the filter empty routine for now because of SYCL issues, but I think it's best to revisit that later

cajita/src/Cajita_ParticleInit.hpp
```diff
@@ -0,0 +1,586 @@
...
+ *                                                                          *
+ * This file is part of the Cabana library. Cabana is distributed under a   *
+ * BSD 3-clause license. For the licensing terms see the LICENSE file in    *
+ * the top-level directory.                                                 *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Copyright (c) 2021 by the Picasso authors                                *
```
lebuller, line 13: Should the picasso copyright header be left in here?

streeve:
Failures unrelated

PR #610, opened 2023-03-08
Fix comands to run the tutorials

dalg24:
Was getting errors on Crusher because the Cajita exe were not launched via `srun`
I assumed the intent was to run with `${MPIEXEC_MAX_NUMPROCS`, similar as for the Cabana Core distributed examples.


streeve (approved):
Looks good

streeve:
Ignoring missing HIP - run manually

PR #611, opened 2023-03-09
Store Cajita comm in shared_ptr following core comm classes

streeve:
Issue with the `LocalGrid` tutorial example found by @dalg24 in Crusher testing, but does not show up in the CI. 

This uses the same strategy as the core `CommunicationPlan` and fixed when I ran on Crusher


streeve:
Note that this follows from #295

sslattery (approved):

streeve:
retest this please.

PR #617, opened 2023-04-28
Default space examples

streeve:
Many of the core examples used explicit backends limiting their use in many cases. New default backends match the Cajita examples


example/cajita_tutorial/01_types/types_example.cpp
```diff
@@ -113,10 +113,11 @@ void typesExample()
...
 int main( int argc, char* argv[] )
 {
-    Kokkos::ScopeGuard scope_guard( argc, argv );
-
-    typesExample();
+    {
+        Kokkos::ScopeGuard scope_guard( argc, argv );
 
+        typesExample();
+    }
```
dalg24, line 120: Why would you add a scope?

streeve: I was copying from the MPI examples - I'll revert these

abisner (approved):

streeve:
Ignoring HIP timeout

PR #618, opened 2023-04-29
Move opensuse builds from CI to nightly

streeve:
Same issue as #604


junghans (approved):

streeve:
@junghans now I see this is failing in the fedora builds as well - what do you think we should do until we require Kokkos 4.0?

streeve:
Closing in favor of #620

PR #628, opened 2023-05-16
Add extra CTest option for nightly builds

streeve:
Same as existing extra CMake args


junghans (approved):

streeve:
HIP timeout unrelated

PR #630, opened 2023-05-24
Add HDF5 overload for position only output

streeve:
Previously required output of at least one field other than position


streeve:
@kwitaechong can you review?

kwitaechong (approved):
Looks good

streeve:
Ignoring HIP timeout

PR #631, opened 2023-05-25
Add hypre calls for init and finalize

lebuller:
Add calls to `HYPRE_Init` and `HYPRE_Finalize` in the `HypreStructuredSolver` which is necessary for the current hypre master (post release 2.28) branch and should be done for hypre standard practice.

### DIFF
--- a/cajita/src/Cajita_HypreStructuredSolver.hpp
+++ b/cajita/src/Cajita_HypreStructuredSolver.hpp
@@ -120,8 +120,6 @@ class HypreStructuredSolver
         : _comm( layout.localGrid()->globalGrid().comm() )
         , _is_preconditioner( is_preconditioner )
     {
-        HYPRE_Init();
-
         static_assert( is_array_layout<ArrayLayout_t>::value,
                        "Must use an array layout" );
         static_assert(
@@ -223,8 +221,6 @@ class HypreStructuredSolver
             HYPRE_StructMatrixDestroy( _A );
             HYPRE_StructStencilDestroy( _stencil );
             HYPRE_StructGridDestroy( _grid );
-
-            HYPRE_Finalize();
         }
     }
 
--- a/cajita/src/Cajita_HypreStructuredSolver.hpp
+++ b/cajita/src/Cajita_HypreStructuredSolver.hpp
@@ -120,6 +120,8 @@ class HypreStructuredSolver
         : _comm( layout.localGrid()->globalGrid().comm() )
         , _is_preconditioner( is_preconditioner )
     {
+        HYPRE_Init();
+
         static_assert( is_array_layout<ArrayLayout_t>::value,
                        "Must use an array layout" );
         static_assert(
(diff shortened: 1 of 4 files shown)


cajita/src/Cajita_HypreStructuredSolver.hpp
```diff
@@ -120,6 +120,8 @@ class HypreStructuredSolver
         : _comm( layout.localGrid()->globalGrid().comm() )
         , _is_preconditioner( is_preconditioner )
     {
+        HYPRE_Init();
```
dalg24, line 123: Is it legal to call `HYPRE_Init()` multiple times in a program?

lebuller: It is legal to call `HYPRE_Init()` and then `HYPRE_Finalize()` followed by another call to `HYPRE_Init()` to create multiple different HYPRE solvers at discrete times. I am not sure if it is legal or if there is an issue with multiple calls to `HYPRE_Init()` before a call to `HYPRE_Finalize()`. I do not see any specific info on such a case in the documentation for hypre.

streeve:
From the call: add the init/final to each example rather than in the solver, with clear comments on user responsibility for that

streeve:
Looks good, but needs to be formatted

streeve:
@lebuller you can format with `make cabana-format`

cajita/unit_test/tstHypreStructuredSolver2d.hpp
```diff
@@ -196,6 +199,9 @@ poissonTest( const std::string& solver_type, const std::string& precond_type,
         for ( int j = owned_space.min( Dim::J ); j < owned_space.max( Dim::J );
               ++j )
             EXPECT_FLOAT_EQ( lhs_host( i, j, 0 ), lhs_ref_host( i, j, 0 ) );
+
+    // finalize hypre
+    HYPRE_Finalize();
```
dalg24, line 204: Does the solver need to be out of scope before HYPRE is finalized?

cajita/unit_test/tstHypreStructuredSolver2d.hpp
```diff
@@ -203,52 +203,72 @@ poissonTest( const std::string& solver_type, const std::string& precond_type,
 //---------------------------------------------------------------------------//
 TEST( structured_solver, pcg_none_test )
 {
+    HYPRE_Init();
     poissonTest( "PCG", "none", TEST_MEMSPACE{} );
+    HYPRE_Finalize();
```
dalg24, line 208: Did you consider doing it in `main()` instead?

streeve: I agree - a single new scope within the poisson function would be cleaner

cajita/unit_test/tstHypreStructuredSolver2d.hpp
streeve: @lebuller can you point to the docs where they discuss this kind of thing?

streeve (approved):

streeve:
format this please

streeve:
HIP timeout unrelated

PR #640, opened 2023-06-09
Improve HDF5 CMake requirements

streeve:
- Fail if non-MPI HDF5 is enabled
- Prefer parallel over serial HDF5 (since serial isn't supported)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,7 @@ if(Cabana_ENABLE_HDF5)
   endif()
 
   include(CheckSymbolExists)
+  LIST(APPEND CMAKE_REQUIRED_INCLUDES ${HDF5_INCLUDE_DIRS})
   check_symbol_exists(H5_HAVE_PARALLEL "H5pubconf.h" HDF5_IS_PARALLEL)
   if(NOT HDF5_IS_PARALLEL)
     message(FATAL_ERROR "Cabana HDF5 support requires parallel HDF5.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,9 @@ if(SILO_FOUND)
 endif()
 
 # find HDF5 (XDMF)
+# Set this to always get the parallel version in case both serial and parallel are installed.
+# Parallel is required (see below).
+set(HDF5_PREFER_PARALLEL TRUE)
 # FIXME: not automatically finding since we need to enable C to use it (which
 #        would break pure CXX projects downstream) unless CMake is new enough
 if(CMAKE_VERSION VERSION_LESS 3.26)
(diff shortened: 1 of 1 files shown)


streeve:
@brtnfld just making sure this is reasonable

streeve:
@guangyechen hopefully this makes your recent issue more clear

guangyechen:
I don't seem to have hdf5-parallel in my system, but compilation with module load hdf5 seems working.

guangyechen (approved):

brtnfld:
Looks reasonable.

streeve:
@junghans I don't understand why the ubuntu hdf5 thinks it's not parallel

junghans:
I have no idea.

CMakeLists.txt
```diff
@@ -123,6 +126,9 @@ endif()
 if(Cabana_ENABLE_HDF5 AND NOT Cabana_ENABLE_MPI)
   message(FATAL_ERROR "Cabana HDF5 support requires MPI.")
 endif()
+if(NOT HDF5_IS_PARALLEL)
+  message(FATAL_ERROR "Cabana HDF5 support requires parallel HDF5.")
+endif()
```
brtnfld, line 132: Changed check for HDF5_IS_PARALLEL to: 

  include(CheckSymbolExists)
  check_symbol_exists(H5_HAVE_PARALLEL "H5pubconf.h" HDF5_IS_PARALLEL)

cc: @junghans

brtnfld: I'm not seeing any output from the find HDF5 package.

streeve: You're right - I updated the new check for only the cases where HDF5 is enabled

junghans:
@rbberger could you review this?

CMakeLists.txt
```diff
@@ -120,8 +123,16 @@ else()
...
 
-if(Cabana_ENABLE_HDF5 AND NOT Cabana_ENABLE_MPI)
-  message(FATAL_ERROR "Cabana HDF5 support requires MPI.")
+if(Cabana_ENABLE_HDF5)
+  if(NOT Cabana_ENABLE_MPI)
+    message(FATAL_ERROR "Cabana HDF5 support requires MPI.")
+  endif()
+
+  include(CheckSymbolExists)
+  check_symbol_exists(H5_HAVE_PARALLEL "H5pubconf.h" HDF5_IS_PARALLEL)
```
rbberger, line 132: Doesn't this need to first set `CMAKE_REQUIRED_INCLUDES` to add the HDF5 include directory, otherwise it will only find the "system" installation in default locations such as `/usr`?

jhendersonHDF (approved):

streeve:
HIP timeout unrelated

PR #644, opened 2023-06-28
Adds HDF5 option to use subfiling

brtnfld:
https://github.com/HDFGroup/hdf5doc/blob/master/RFCs/HDF5_Library/VFD_Subfiling/user_guide/HDF5_Subfiling_VFD_User_s_Guide.pdf

Subfiling can be configured either through environment variables or APIs.


core/src/Cabana_HDF5ParticleOutput.hpp
```diff
@@ -31,6 +31,11 @@
 #include <Kokkos_Core.hpp>
 
 #include <hdf5.h>
+#if H5_HAVE_SUBFILING_VFD
```
jhendersonHDF, line 34: These should probably all be #ifdefs instead of #ifs or else I believe it will generate compiler warnings when Subfiling isn't enabled.

jhendersonHDF (approved):

core/src/Cabana_HDF5ParticleOutput.hpp
brtnfld: Fixed.

brtnfld:
@streeve, I've finished the advanced HDF5 example.

brtnfld:
I don't think CI will test these PR features since it needs HDF5 > =1.14

brtnfld:
Added to CI: building HDF5 version 1.14.3 for testing with subfiling.

streeve:
Just rebased and pushed (I kept a backup copy of your branch as well) - one quick question for when subfiling isn't available and I think it's ready to merge. Thanks for all your time on this!

brtnfld:
What was the question about if subfiling is not available?

streeve (approved):
Sorry, this comment got left in "pending" - looks like the CI worked, so this is ok with me

example/core_tutorial/CMakeLists.txt
```diff
@@ -37,4 +37,7 @@ if(Kokkos_ENABLE_CUDA)
 endif()
 if(Cabana_ENABLE_HDF5)
   add_subdirectory(13_hdf5_output)
+  if(HDF5_VERSION VERSION_GREATER_EQUAL 1.14.0 AND HDF5_ENABLE_SUBFILING_VFD)
```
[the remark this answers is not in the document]
brtnfld: Looks fine.

streeve:
HIP timeout unrelated

PR #646, opened 2023-07-04
add sub communicator in the GlobalGrid

kwitaechong:
add sub communicator in the designated or reduction direction

### DIFF
--- a/cajita/src/Cajita_GlobalGrid_impl.hpp
+++ b/cajita/src/Cajita_GlobalGrid_impl.hpp
@@ -212,6 +212,74 @@ GlobalGrid<MeshType>::blockRank( const int i, const int j ) const
 }
 
 //---------------------------------------------------------------------------//
+// Get the MPI sub communicator for given direction
+template <class MeshType>
+MPI_Comm GlobalGrid<MeshType>::subComm( const int dim ) const
+{
+    MPI_Comm sub_comm;
+
+    // Set colors same within the dimBlockId of the dim
+    MPI_Comm_split( comm(), dimBlockId( dim ), blockId(), &sub_comm );
+    return sub_comm;
+}
+
+//---------------------------------------------------------------------------//
+// Get the MPI sub communicator for reduction in the given direction
+template <class MeshType>
+MPI_Comm GlobalGrid<MeshType>::reductionSubComm( const int dim ) const
+{
+    MPI_Comm sub_comm;
+
+    // non reduciton directions
+    std::array<int, num_space_dim - 1> no_dir;
+    if ( num_space_dim == 2 )
+    {
+        switch ( dim )
+        {
+        case Dim::I:
+            no_dir[0] = Dim::J;
+            break;
+        case Dim::J:
+            no_dir[0] = Dim::I;
+            break;
+        }
+    }
+    else if ( num_space_dim == 3 )
+    {
+        switch ( dim )
+        {
+        case Dim::I:
+            no_dir[0] = Dim::J;
+            no_dir[1] = Dim::K;
+            break;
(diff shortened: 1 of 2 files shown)


cajita/src/Cajita_GlobalGrid_impl.hpp
```diff
@@ -211,6 +211,74 @@ GlobalGrid<MeshType>::blockRank( const int i, const int j ) const
...
+
+    // Set colors same within the dimBlockId of the dim
+    MPI_Comm_split( comm(), dimBlockId( dim ), blockId(), &sub_comm );
+    return sub_comm;
+}
+
+//---------------------------------------------------------------------------//
+// Get the MPI sub communicator for reduction in the given direction
+template <class MeshType>
+MPI_Comm GlobalGrid<MeshType>::reductionSubComm( const int dim ) const
```
streeve, line 229: I think this will be easier if you pass 2 dimensions here (and restrict to 3d only) and rename to `subComm`

cajita/src/Cajita_GlobalGrid_impl.hpp
```diff
@@ -211,6 +211,74 @@ GlobalGrid<MeshType>::blockRank( const int i, const int j ) const
...
+{
+    MPI_Comm sub_comm;
+
+    // Set colors same within the dimBlockId of the dim
+    MPI_Comm_split( comm(), dimBlockId( dim ), blockId(), &sub_comm );
+    return sub_comm;
+}
+
+//---------------------------------------------------------------------------//
+// Get the MPI sub communicator for reduction in the given direction
```
streeve, line 227: This creates a sub communicator in 2 dimensions, right?

cajita/src/Cajita_GlobalGrid_impl.hpp
```diff
@@ -211,6 +211,74 @@ GlobalGrid<MeshType>::blockRank( const int i, const int j ) const
...
+template <class MeshType>
+MPI_Comm GlobalGrid<MeshType>::reductionSubComm( const int dim ) const
+{
+    MPI_Comm sub_comm;
+
+    // non reduciton directions
```
streeve, line 233: reduction

streeve (changes requested):

cajita/src/Cajita_GlobalGrid_impl.hpp
kwitaechong: Yes. create sub communicator in 2D plane for 3D and sub communicator in 1D line for 2D.

cajita/src/Cajita_GlobalGrid_impl.hpp
kwitaechong: `subComm` is used for creating sub communicator in a given direction and this `reductionSubComm` is used for creating sub communicator when there is reduction in the given direction.

cajita/src/Cajita_GlobalGrid_impl.hpp
kwitaechong: fixed

cajita/src/Cajita_GlobalGrid_impl.hpp
streeve: But can't `subComm` be used for a reduction into 1d and `reductionSubComm` be used for reduction into 2d? Why not pass the 2 "non-reduction" dimensions here like you do in the case above?

streeve:
@sslattery this is to make our grid-based binning operations downstream simpler

streeve:
Closing for now - planning to add this and use it only interally within an Array reduction operation instead

PR #656, opened 2023-08-02
Remove Cajita Halo stored comm

streeve:
This relies on the global grid having created a new cartesian comm (in a shared_ptr) and using that directly, rather than storing a duplicated copy


cajita/src/Cajita_Halo.hpp
```diff
@@ -457,29 +456,19 @@ class Halo
...
-    void getComm( const Array_t& array, const ArrayTypes&... arrays )
+    MPI_Comm getComm( const Array_t&, const ArrayTypes&... arrays ) const
     {
         // Recurse.
-        getComm( arrays... );
-
-        // Check that the communicator of this array is the same as the halo
-        // comm.
-        int result;
-        MPI_Comm_compare( array.layout()->localGrid()->globalGrid().comm(),
```
sslattery, line 479: Shouldn't we still do this comparison when we have multiple arrays?

sslattery (approved):
Looks good - just one comment on comparison we may want to resolve.

cajita/src/Cajita_Halo.hpp
```diff
@@ -457,29 +456,27 @@ class Halo
...
+        auto comm = getComm( array );
 
-        // Check that the communicator of this array is the same as the halo
-        // comm.
+        // Check that the communicator of this array is the same as the other
+        // arrays.
         int result;
-        MPI_Comm_compare( array.layout()->localGrid()->globalGrid().comm(),
-                          _comm, &result );
+        MPI_Comm_compare( comm, getComm( arrays... ), &result );
```
sslattery, line 475: nice! This will achieve the proper recursion.

streeve:
retest this please.

streeve:
@sslattery in this case the comm is identical, so we just needed to add another comparison

streeve:
Ignoring HIP timeout

PR #658, opened 2023-08-03
Draft: FindHYPRE came

abisner:
Added FindHYPRE cmake for autotools install in lieu of available CMake installation.


abisner:
This PR adds a FindHYPRE cmake to be used in the case of Kokkos HIP builds, in which there is currently no supported CMake installation method for hypre builds on AMD.

cmake/FindHYPRE.cmake
```diff
@@ -0,0 +1,20 @@
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_HYPRE hypre QUIET)
```
junghans, line 2: We could use `IMPORTED_TARGET` here and then do something like:
```
if(HIP_BUILD)
  pkg_check_modules(PC_HYPRE hypre IMPORTED_TARGET)
  add_library(ALIAS hypre::hypre pkgconfig:HYPRE)
else()
  find_package(HYPRE)
endif()
```

cmake/FindHYPRE.cmake
```diff
@@ -0,0 +1,20 @@
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_HYPRE hypre QUIET IMPORTED_TARGET)
+
+find_path(HYPRE_INCLUDE_DIR hypre.h HINTS ${PC_HYPRE_INCLUDE_DIRS})
```
junghans, line 4:
```suggestion
find_path(HYPRE_INCLUDE_DIR HYPRE.h HINTS ${PC_HYPRE_INCLUDE_DIRS})
```

CMakeLists.txt
```diff
@@ -92,7 +92,12 @@ Cabana_add_dependency( PACKAGE ALL )
 find_package( CLANG_FORMAT 10 )
 
 # find hypre
-Cabana_add_dependency( PACKAGE HYPRE VERSION 2.22.1 )
+if(Kokkos_ENABLE_HIP)
+  pkg_check_modules(PC_HYPRE hypre IMPORTED_TARGET)
+  add_library(ALIAS hypre::hypre pkgconfig:HYPRE)
```
junghans, line 97:
```suggestion
  add_library(ALIAS HYPRE::HYPRE pkgconfig:HYPRE)
```

cmake/FindHYPRE.cmake
```diff
@@ -0,0 +1,20 @@
+find_package(PkgConfig QUIET)
```
junghans, line 1: I am not sure if we even need this file at all.

streeve: Agreed - removed

streeve:
Okay @junghans failure here is finally just missing the right target `"pkgconfig:PC_HYPRE" does not already exist.`

CMakeLists.txt
```diff
@@ -92,7 +92,13 @@ Cabana_add_dependency( PACKAGE ALL )
 find_package( CLANG_FORMAT 10 )
 
 # find hypre
-Cabana_add_dependency( PACKAGE HYPRE VERSION 2.22.1 )
+if(Kokkos_ENABLE_HIP)
+  # FIXME: remove when hypre supports HIP CMake builds
+  pkg_check_modules(PC_HYPRE hypre IMPORTED_TARGET)
+  add_library(HYPRE::HYPRE ALIAS pkgconfig:PC_HYPRE)
```
junghans, line 98:
```suggestion
  add_library(HYPRE::HYPRE ALIAS PkgConfig::PC_HYPRE)
```
I had wrong prefix (see https://cmake.org/cmake/help/latest/module/FindPkgConfig.html)

streeve:
@junghans I tried some more in a container, but I don't see any pkg-config content from hypre anywhere.. If we point to that correctly I think it will finally work

CMakeLists.txt
```diff
@@ -92,7 +92,13 @@ Cabana_add_dependency( PACKAGE ALL )
 find_package( CLANG_FORMAT 10 )
 
 # find hypre
-Cabana_add_dependency( PACKAGE HYPRE VERSION 2.22.1 )
+if(Kokkos_ENABLE_HIP)
+  # FIXME: remove when hypre supports HIP CMake builds
+  pkg_check_modules(PC_HYPRE hypre REQUIRED IMPORTED_TARGET)
```
junghans, line 97: We need to add logic for `Cabana_ENABLE_HYPRE` here, too.

streeve:
Now blocked by #697

streeve:
Should be able to close once https://github.com/hypre-space/hypre/tree/cmake-hip is merged @victorapm

PR #662, opened 2023-08-10
Allow Kokkos::Views for neighbor lists

streeve:


streeve:
@abisner can you take a look at this?

lebuller (approved):

PR #675, opened 2023-08-25
Use Kokkos math functions in kernel

streeve:
Promoted in Kokkos 3.7


streeve:
Ignoring unrelated failures for neighbor list and HIP timeout

PR #677, opened 2023-08-26
Modify 05_slice example to use DefaultExecutionSpace

dineshadepu:


example/core_tutorial/05_slice/slice_example.cpp
```diff
@@ -56,16 +56,15 @@ void sliceExample()
...
-      allocated on NVIDIA devices use `Kokkos::CudaSpace` instead of
-      `Kokkos::HostSpace`.
+      We declare the memory space in which the AoSoA will be allocated
+      on the default execution space we are dealing with.
     */
-    using MemorySpace = Kokkos::HostSpace;
-    using ExecutionSpace = Kokkos::DefaultHostExecutionSpace;
+    using ExecutionSpace = Kokkos::DefaultExecutionSpace;
+    using MemorySpace = typename ExecutionSpace::memory_space;
     using DeviceType = Kokkos::Device<ExecutionSpace, MemorySpace>;
```
streeve, line 68: Just a note that #585 will change most of the examples to use `memory_space` directly - it may be best to wait until that is merged

example/core_tutorial/05_slice/slice_example.cpp
```diff
@@ -78,6 +77,9 @@ void sliceExample()
     Cabana::AoSoA<DataTypes, DeviceType, VectorLength> aosoa( "my_aosoa",
                                                               num_tuple );
 
+    // Create a mirror view of the aosoa on the host for accessing it legally
```
streeve, line 80:
```suggestion
    // Create a mirror of the AoSoA on the host (this will not make a new copy if the memory is already allocated on the host, but will if it is in device memory) in order to print out the data.
```

example/core_tutorial/05_slice/slice_example.cpp
```diff
@@ -56,16 +56,15 @@ void sliceExample()
...
-      HostSpace allocates memory in standard CPU RAM.
+      The current example uses default spaces to work across all supported
+      backends, but explicit choices like HostSpace and CudaSpace can also be
+      used.
 
-      Kokkos also supports execution on GPUs. For example, to create an AoSoA
-      allocated on NVIDIA devices use `Kokkos::CudaSpace` instead of
-      `Kokkos::HostSpace`.
+      We declare the memory space in which the AoSoA will be allocated
+      on the default execution space we are dealing with.
```
streeve, line 64:
```suggestion
      We create the AoSoA particle data in the memory space associated
      with the default execution space (see https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Initialization.html).
```

streeve (changes requested):
The code will need to be formatted - see https://github.com/ECP-copa/Cabana/blob/master/CONTRIBUTING.md for clang-format and pre-commit details

example/core_tutorial/05_slice/slice_example.cpp
dineshadepu: Sure, I will update the rest of the examples after #585 is merged. Will keep the PR open until then.

streeve: @dineshadepu should be okay to continue this PR now that we've merged the blocking PR - let me know if those changes are confusing in any way. But I do not plan to include this in the imminent release, so feel free to wait until that's finished

PR #681, opened 2023-09-01
Remove format action

streeve:
Not used and clang-format version is out of date


.github/workflows/format.yml
```diff
@@ -1,47 +0,0 @@
-name: Format
-on:
-  issue_comment:
-    types: [created]
-jobs:
-  edit:
-    name: clang-format
-    if: ${{ github.event.comment.body == 'format this please' }}
-    runs-on: ubuntu-20.04
```
junghans, line 9: We could use a never version of clang-format by using a newer ubuntu!

junghans (approved):

Issue #457, opened 2021-10-04
Load balancing roadmap


streeve:
Before we merge #444 I think we need to unify it with out existing capability now that Yuxing is back. I would suggest (feel free to edit):

- [ ] Create a base `LoadBalancer` class with the particle workload that exists in `SparseDimPartitioner`
- [ ] Inherit from `LoadBalancer` and overload a couple methods for the `SparseLoadBalancer` class
- [ ] Rename #444 `LoadBalancerALL` (or similar) and inherit from `LoadBalancer`

Maybe we need to discuss again what really belongs in the global grid vs partitioner vs load balancer, but the current steps may be enough

streeve:
Draft in #541

PR #682, opened 2023-09-01
Add LoadBalancer to top-level header

streeve:
Missing since addition to the library

### DIFF
--- a/cajita/src/Cajita.hpp
+++ b/cajita/src/Cajita.hpp
@@ -60,4 +60,8 @@
 #include <Cajita_SiloParticleOutput.hpp>
 #endif
 
+#ifdef Cabana_ENABLE_ALL
+#include <Cajita_LoadBalancer.hpp>
+#endif
+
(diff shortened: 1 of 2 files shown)


cajita/src/Cajita.hpp
```diff
@@ -60,4 +60,8 @@
 #include <Cajita_SiloParticleOutput.hpp>
 #endif
 
+#ifdef Cabana_ENABLE_ALL
```
lebuller, line 63: Not sure I understand why the load balancer is dependent on `Cabana_ENABLE_ALL`? More so a question for my understanding than anything

lebuller (approved):

cajita/src/Cajita.hpp
streeve: It uses the external library. There's more info in #457 - basically I think this should be named `LoadBalancerALL`, but since it's the only one we have right now it's not

streeve:
Failures unrelated

PR #686, opened 2023-09-13
Update readme overall description

streeve:
Update wording to be more clear on applications and capability


kwitaechong (approved):

streeve:
Ignoring test timeouts (only changing readme)

PR #531, opened 2022-04-26
Update to clang-format-14 in CI

streeve:

PR #688, opened 2023-09-13
Update min required clang-format in CMakeLists

streeve:
Was updated in #531, but never in CMake


junghans (approved):

streeve:
Ignoring unrelated device CI failures (only updating format version)

PR #689, opened 2023-09-14
fixup: HIP CI heffte not compiled due to version

streeve:
Make tpl required in HIP CI


streeve:
Now blocked by #697; would also require a follow-on that installs rocm-complete

streeve:
Closing until we figure out the best way to get rocfft

PR #692, opened 2023-09-19
Introducing Gaussian Mixture Models

petschge:
which is a number of weightes Gaussians that describe a particle distribution function in velocity space. The mixture model can be computed from a set of particles and we also allow to draw new particles from such a Gaussian Mixture Model. The code follows the papers by Figueiredo and Jain, 2002
(https://dx.doi.org/https://dl.acm.org/doi/10.1109/34.990138), but adds particles weights like the paper by Chen, Chacon, Nguyen, 2021 (DOI:https://doi.org/10.1016/j.jcp.2021.110185)

This is not against the current head of master, but I assume that there is other changes that I need to make before rebasing and submitting for good. Feedback (here or on Slack) welcome!

PR #696, opened 2023-09-29
Add random velocity initialization functions

streeve:
Related to #692 

Pass a second init tag (so far just `InitRandom`) to create velocities. Version for `ParticleList` and for `Slice`, both fused with the particle creation kernel 

Needs more testing, more convenience interfaces, potentially more thought on sampling from distributions as compared to fully random creation

### DIFF
--- a/core/src/Cabana_ParticleInit.hpp
+++ b/core/src/Cabana_ParticleInit.hpp
@@ -150,7 +150,6 @@ int createParticles(
     return count_host( 0 );
 }
 
-//---------------------------------------------------------------------------//
 /*!
   \brief Initialize random particles given an initialization functor.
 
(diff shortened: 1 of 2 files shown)


core/src/Cabana_ParticleInit.hpp
```diff
@@ -494,6 +717,138 @@ createRandomParticlesMinDistance( ExecutionSpace exec_space,
...
+  already in the container (and should be unchanged).
+  \param shrink_to_fit Optionally remove unused allocated space after creation.
+  \param seed Optional random seed for generating particles.
+
+  \return Number of particles created.
+*/
+template <class ExecutionSpace, class InitFunctor, class ParticleListType,
+          class PositionTag, class VelocityTag, class ArrayType>
+int createParticles(
+    InitRandom tag, InitRandom, ExecutionSpace exec_space,
```
kwitaechong, line 751: remove the last `InitRandom`? or one is for position and the other is for velocity?

core/src/Cabana_ParticleInit.hpp
```diff
@@ -494,6 +717,138 @@ createRandomParticlesMinDistance( ExecutionSpace exec_space,
...
+
+        // If we created a new particle insert it into the list.
+        if ( create )
+        {
+            // Particle velocity.
+            double pv[3];
+            auto gen = pool.get_state();
+            for ( int d = 0; d < 3; ++d )
+                Cabana::get( particle, VelocityTag(), d ) =
+                    Kokkos::rand<RandomType, double>::draw( gen, 0.0, 1.0 );
```
kwitaechong, line 790: we don't necessarily assume its value between 0.0 and 1.0. Can we assign this value from the argument?

Issue #634, opened 2023-05-30
fedora:rawhide MPI I/O issues


streeve:
HDF5 and BOV write timeout failures for >1 MPI rank (currently disabled in nightly builds)

PR #697, opened 2023-10-05
HIP build CI

streeve:
Current frequent build failures due to container size


streeve:
@junghans this is ready to merge - failures are unrelated (but same timeout issue we have in #634 nightly builds with the recent ci-containers updates)

streeve:
@lebuller I'm going to merge this since the Cuda failure is entirely unrelated to this HIP CI change, but can you take a look at whether this UVM failure on Jenkins is quickly fixable? I've never seen it before, but we don't use UVM much

PR #698, opened 2023-10-09
Fix heFFTe overflow

patrickb314:
Heffte FFTs would crash on MPI communication in some cases because ~the workspace passed in wasn't ever allocated.~ of integer overflow. Also use the proper heffte routine to query the size of the workspace to allocate.

### DIFF
--- a/grid/src/Cabana_Grid_FastFourierTransform.hpp
+++ b/grid/src/Cabana_Grid_FastFourierTransform.hpp
@@ -555,7 +555,7 @@ class HeffteFastFourierTransform
         _fft = Impl::createHeffteFft3d(
             heffte_execution_space, heffte_backend_type{}, inbox, outbox,
             layout.localGrid()->globalGrid().comm(), heffte_params );
-        int fftsize = std::max( _fft->size_outbox(), _fft->size_inbox() );
+        long long fftsize = std::max( _fft->size_outbox(), _fft->size_inbox() );
 
         // Check the size.
         auto entity_space =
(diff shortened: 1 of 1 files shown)


patrickb314:
Hmmm, not sure why this lets things run on Tioga. the previous code was allocating but just not initializing. Let me double-check before we accept this PR.

grid/src/Cabana_Grid_FastFourierTransform.hpp
```diff
@@ -568,8 +568,7 @@ class HeffteFastFourierTransform
             Kokkos::ViewAllocateWithoutInitializing( "fft_work" ),
             2 * fftsize );
         _workspace = Kokkos::View<Scalar* [2], memory_space>(
-            Kokkos::ViewAllocateWithoutInitializing( "workspace" ),
-            4 * fftsize );
+            "workspace", _fft->size_workspace() );
```
streeve, line 571: Change makes sense - let me know if you're satisfied with the change. We'll want to add a comment along the lines of "Purposely using zero-init on allocation" to remind us not to make this mistake again

patrickb314: I'm not sure I am. The initialization shouldn't be necessary. I suspect the problem was actually that fftsize is an int and 4*fftsize was overflowing it. Using _fft->size_workspace() avoids this overflow, and so I just coincidentally fixed the overflow. Let me test and submit a revised PR instead.

patrickb314: Okay, I think the latest change to this PR fixes it. fftwork was stored as an int when fft sizes could overflow that. Also changed to use the heffte interface to determine the right workspace size to allocate.

streeve (approved):
Looks good

patrickb314:
Not sure why the relevant checks aren't passing so this won't auto-merge. Do I need to do something to resolve those, because they mainly seem to be failing from timeouts in the CI system...

streeve:
> Not sure why the relevant checks aren't passing so this won't auto-merge. Do I need to do something to resolve those, because they mainly seem to be failing from timeouts in the CI system...

No, that's unfortunately the state of our CI right now. Failures are unrelated, so I'll merge

PR #699, opened 2023-10-12
Guard adding HDF5 include directories

dalg24:
Fix
```
CMake Error in core/src/CMakeLists.txt:
  Target "cabanacore" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "<Cabana>/core/src/HDF5_C_INCLUDE_DIR-NOTFOUND"

  which is prefixed in the source directory.
```
Bug was introduced in #590


dalg24:
(fixup indentation)

streeve (approved):
Thanks

streeve (approved):
Looks good. Jenkins timeout unrelated (hdf5 not installed in those containers). I/O timeouts not fixed

PR #701, opened 2023-10-12
Link against namespaced targets in examples and benchmarks

dalg24:
Rational: Showcasing what users are expected to do downstream.  Unscoped targets are not exported.


streeve (approved):
Looks good - HIP and I/O timeouts unrelated

PR #703, opened 2023-10-16
Update benchmark plotting scripts

streeve:
Small improvements and fixes


streeve:
@pabloseleson can you test, review, and add any local changes that you think are useful?

streeve:
@abisner @kwitaechong I think this just needs a quick review. It does a lot, but only in the benchmarks and plotting functionality

kwitaechong (approved):

PR #712, opened 2023-11-07
Adds linked cell parallel capability

lebuller:


core/src/Cabana_Parallel.hpp
```diff
@@ -251,7 +253,8 @@ template <class FunctorType, class NeighborListType, class... ExecParameters>
 inline void neighbor_parallel_for(
     const Kokkos::RangePolicy<ExecParameters...>& exec_policy,
     const FunctorType& functor, const NeighborListType& list,
-    const FirstNeighborsTag, const SerialOpTag, const std::string& str = "" )
+    const FirstNeighborsTag, const SerialOpTag, const std::string& str = "",
+    typename std::enable_if<( !is_linked_cell_list<NeighborListType>::value ), int>::type* = 0 )
```
lebuller, line 257: Currently checking that it is not a linked cell list for non-linked cell neighbor parallel as I didn't see a `is_neighbor_list` check. Not sure if this is the best way to differentiate the linked cell approach from this approach.

core/unit_test/tstNeighborList.hpp
```diff
@@ -24,6 +24,7 @@ namespace Test
 {
 //---------------------------------------------------------------------------//
 // Linked cell list cell stencil test.
+/*
```
lebuller, line 27: I commented out the portion of the neighbor list test that checks the `linkedCellStencil` since that has been moved into the linked cell list directly. We should probably move the contents of this test into the `linkedCellList` test and alter it to use the stencil as part of the linked cell list

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -75,21 +76,33 @@ class LinkedCellList
...
+      \param neighborhood_radius Radius for neighbors
+
+      \param cell_size_ratio Ratio of the cell size to the neighborhood size
     */
     template <class SliceType>
     LinkedCellList(
         SliceType positions, const typename SliceType::value_type grid_delta[3],
         const typename SliceType::value_type grid_min[3],
         const typename SliceType::value_type grid_max[3],
+        const Scalar neighborhood_radius, const Scalar cell_size_ratio = 1,
```
streeve, line 89: We require `const typename SliceType::value_type` for all the other scalars, so we can do the same for the cutoff

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -75,21 +76,33 @@ class LinkedCellList
...
                  grid_max[1], grid_max[2], grid_delta[0], grid_delta[1],
                  grid_delta[2] )
+        , _sorted( false )
+        , _rsqr( neighborhood_radius * cell_size_ratio )
     {
         std::size_t np = positions.size();
         allocate( totalBins(), np );
         build( positions, 0, np );
+
+        // Stencil constructor carry-over
```
streeve, line 102: I'm realizing I probably never said it, but I imaged keeping a separate struct and constructing a stencil instance inside the LCL

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -222,6 +268,29 @@ class LinkedCellList
     KOKKOS_INLINE_FUNCTION
     std::size_t rangeEnd() const { return _bin_data.rangeEnd(); }
 
+    /*!
+      /bried Given a cell, get the index bounds of the cell stencil
+    */
+    KOKKOS_INLINE_FUNCTION
+    void getCells( const int cell, int& imin, int& imax, int& jmin, int& jmax,
```
streeve, line 275: Not sure exactly how, but we need to decide how to differentiate this from `getBins`

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -431,15 +566,128 @@ void permute(
...
+
+    // Update internal state.
+    linked_cell_list.update( true );
 }
 
 //---------------------------------------------------------------------------//
+//! Stencil of cells surrounding each cell.
+/*
+template <class Scalar>
+struct LinkedCellStencil
```
streeve, line 584: I would rather keep this, then add an interface `LinkedCellList::getStencil` which calls `getCells`

core/src/Cabana_Parallel.hpp
streeve: Yeah, we don't have one. If we figure out the `getNeighbor` we can remove it anyway (but it would require some kind of other tag for the LCL only variant)

core/unit_test/tstNeighborList.hpp
streeve: I agree, let's move this test

core/src/Cabana_LinkedCellList.hpp
streeve, line 89: Let's default the cutoff to be equal to the `grid_delta`, first because it's most often how we use it and second because it keeps backwards compatibility for free

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -75,21 +76,33 @@ class LinkedCellList
...
         const typename SliceType::value_type grid_min[3],
         const typename SliceType::value_type grid_max[3],
+        const Scalar neighborhood_radius, const Scalar cell_size_ratio = 1,
         typename std::enable_if<( is_slice<SliceType>::value ), int>::type* =
             0 )
         : _grid( grid_min[0], grid_min[1], grid_min[2], grid_max[0],
                  grid_max[1], grid_max[2], grid_delta[0], grid_delta[1],
                  grid_delta[2] )
+        , _sorted( false )
+        , _rsqr( neighborhood_radius * cell_size_ratio )
```
streeve, line 96: typo?

streeve (changes requested):

core/src/Cabana_LinkedCellList.hpp
lebuller: Maybe renaming this to `getStencilCells` and `getBins` to `getParticleBins`? I think those names would be better descriptions of what the functions are actually doing while also being differentiable from each other

core/src/Cabana_LinkedCellList.hpp
lebuller: What is the typo?

core/src/Cabana_LinkedCellList.hpp
lebuller: Do you mean default the `neighborhood_radius` to be the `grid_delta` similar to defaulting the `cell_size_ratio` to 1? Or do you mean assume that they are equal implicitly and when the `neighborhood_radius` is used in the constructor to replace it with the `grid_delta`? I don't believe the first option is legal since  `grid_delta` is an input to this function

core/src/Cabana_LinkedCellList.hpp
lebuller: So are you thinking to leave the `linkedCellStencil` mainly as it is and then just storing a copy of it when creating the `linkedCellList`? Then creating a wrapper function similar to those for the grid and bindata functions that calls `getCells` from the stencil stored as a member of the list?

streeve: Yeah, I think that keeps a good encapsulation

core/src/Cabana_LinkedCellList.hpp
streeve: I like those names - still a bit unsure of whether we should just call everything "bin" instead of "cell". But stencil vs particle is helpful no matter what

core/src/Cabana_LinkedCellList.hpp
streeve: You're right - we can just add another constructor without the cutoff argument that defaults it during construction

core/src/Cabana_LinkedCellList.hpp
streeve: `rsqr` should `radius * radius`, right?

core/src/Cabana_LinkedCellList.hpp
lebuller: Tried this, but to have the stencil as a member, we need a Scalar type to be used for it and I couldn't get it to work. I'm sure it's possible but I'm not sure how to setup the templating correctly for it

streeve:
Initial performance check on Frontier (with Verlet results for comparison)
linkedcell
verlet

Clearly important to keep the linked cell interface separate from the neighbor iteration

streeve:
CI timeouts unrelated (and manually tested on HIP)

streeve (approved):

core/src/Cabana_LinkedCellList.hpp
```diff
@@ -141,10 +141,13 @@ class LinkedCellList
                  grid_delta[2] )
         , _cell_stencil( grid_delta[0], 1.0, grid_min, grid_max )
         , _sorted( false )
+        , _binned( false )
     {
         std::size_t np = positions.size();
         allocate( totalBins(), np );
         build( positions, 0, np );
+        getParticleBins();
```
streeve, line 149: Can we just call this in `build` instead (called in every constructor) and then not have to store/check if it's binned?

lebuller: We probably can. I'll try that We also need to call it in permute as well since we change the particle IDs there.

streeve: Good point

streeve:
Test failures unrelated (docker, jenkins, coverage)

PR #717, opened 2023-11-20
Create neighbor list histograms

streeve:
Add method to construct neighbor histograms. ~Missing test and need to decide whether we output bin edge vs bin center~

### DIFF
--- a/core/src/Cabana_NeighborList.hpp
+++ b/core/src/Cabana_NeighborList.hpp
@@ -18,6 +18,8 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <Cabana_Sort.hpp>
+
 namespace Cabana
 {
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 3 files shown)


core/src/Cabana_NeighborList.hpp
```diff
@@ -114,6 +116,50 @@ maxNeighbor( const ListType& list, const std::size_t num_particles )
...
+    Kokkos::RangePolicy<ExecutionSpace> particle_policy( exec_space, 0,
+                                                         num_particles );
+    Kokkos::parallel_for( particle_policy, extract_functor );
+    Kokkos::fence();
+
+    auto bin_data = Cabana::binByKey( num_neigh, num_bin );
+
+    auto histogram = Kokkos::View<int* [2], typename ListType::memory_space>(
+        "particle_count", num_bin, 2 );
+    int max_neigh = NeighborList<ListType>::maxNeighbor( list );
```
lebuller, line 148: It looks like we call this on host here, but the `maxNeighbor` interface calls `numNeighbor` which accesses a device view, labelled `num_neighbors`, which leads to an inaccessible memory issue when using the neighbor histogram with HIP.

core/src/Cabana_NeighborList.hpp
```diff
@@ -108,12 +110,67 @@ maxNeighbor( const ListType& list, const std::size_t num_particles )
...
+    Kokkos::fence();
+
+    auto bin_data = Cabana::binByKey( num_neigh, num_bin );
+
+    auto histogram = Kokkos::View<int* [2], typename ListType::memory_space>(
+        "particle_histogram", num_bin, 2 );
+
+    int max_neigh;
+    Kokkos::Max<int> max_reduce( max_neigh );
+    Kokkos::parallel_reduce(
```
lebuller, line 153: I setup this reduce similar to the way `maxNeighbor` is called in `neighbor_unit_test.hpp` but it looks like we may be double looping over particles this way. We loop over particles in the `parallel_reduce` and then again in `Impl::maxNeighbor`. This version appears to work but we should probably look at changing the `maxNeighbor` interface so that we don't have this double loop

streeve:
retest this please.

PR #729, opened 2024-01-03
Implement `Impl::max_reduce` since `ArborX::max` has been deprecated

dalg24:
The `ArborX::max` algorithm was never intended to be public and was deprecated in arborx/ArborX#998
This PR introduces `Impl::max_reduce`.  An alternative would be to use the Kokkos (experimental) parallel algorithms.

Loosely related to arborx/ArborX#1013


dalg24:
Addresses https://github.com/ECP-copa/Cabana/pull/728#issuecomment-1874802641

streeve:
HIP and docker CI failures unrelated

streeve (approved):

PR #735, opened 2024-01-29
Re-enable jenkins HIP CI

streeve:


docker/Dockerfile.hipcc
```diff
@@ -100,7 +100,7 @@ RUN ARBORX_VERSION=4834bff44c23c9510c6ed93366638dcdf85ab217 && \
       -D CMAKE_INSTALL_PREFIX=${ARBORX_DIR} \
       -D CMAKE_BUILD_TYPE=Debug \
       -D CMAKE_CXX_COMPILER=hipcc \
-      -D CMAKE_CXX_FLAGS=-amdgpu-target=gfx906 \
+      -D CMAKE_CXX_FLAGS=-amdgpu-target=gfx908 \
```
dalg24, line 103: Wait why do we add flags here?

streeve:
retest this please.

streeve:
retest this please.

docker/Dockerfile.hipcc
dalg24: ping

streeve: Looks like we've had that since we added TPL to HIP builds. All the jenkins builds need version/option updates - are you okay if I do that separately?

dalg24: Sure.  You might want (whenever you handle this) to pass `-D GPU_TARGETS=gfx908`.
This is intended for rocThrust, making sure that it does not make you compile for all AMD GPU architectures that cmake knows about.

PR #737, opened 2024-02-02
Replace {push,pop}Region with ScopedRegion

### LINKED ISSUE #736: Possibly unclosed Kokkos Profiling Region in Halo gather
A Kokkos profiling region is created in `Cabana::Grid::Halo::gather` at [grid/src/Cabana_Grid_Halo.hpp:291](https://github.com/ECP-copa/Cabana/blob/9cc3824a967eadad1c017622c53f976a8115e5c8/grid/src/Cabana_Grid_Halo.hpp#L291). When the neighbor size is zero, `gather` will immediately return without closing the profiling region.

This resulted in a SEGFAULT when profiling a test case on a single rank.

@kwitaechong

kwitaechong:
replace pushRegion/popRegion with ScopedRegion for cases where the function returns early, etc.

closes #736

### DIFF
--- a/core/src/Cabana_Distributor.hpp
+++ b/core/src/Cabana_Distributor.hpp
@@ -21,6 +21,7 @@
 #include <Cabana_Slice.hpp>
 
 #include <Kokkos_Core.hpp>
+#include <Kokkos_Profiling_ScopedRegion.hpp>
 
 #include <mpi.h>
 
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 # Dependencies
 #------------------------------------------------------------------------------#
 # find kokkos
-find_package(Kokkos 3.7 REQUIRED)
+find_package(Kokkos 4.1 REQUIRED)
```
streeve, line 39: Need to change the wiki as well (once we merge)

streeve (changes requested):
We'll need to update all the CI from 3.7 to 4.1

CMakeLists.txt
dalg24: I think you should bump the minimum requirement in a separate PR.

streeve:
@kwitaechong this can be rebased on master

streeve (approved):

streeve:
Ignoring unrelated jenkins failures

PR #740, opened 2024-03-20
Comment spelling using codespell

streeve:
From https://github.com/codespell-project/codespell


streeve:
@junghans even though we try to force re-installation of `icpx` in the CoPA `ci-containers` it's lost again:

` icpx  is not a full path and was not found in the PATH.`

I'm not sure what to try next

junghans (approved):

streeve:
jenkins failures unrelated

PR #742, opened 2024-03-21
Added prefix argument to Grid writeTimeStep

DavidJoy8:
Grid's BovWriter currently does not allow the user any control over the filename.  This change makes it work the same as the HDF5 particle output.


grid/src/Cabana_Grid_BovWriter.hpp
```diff
@@ -172,16 +172,17 @@ reorderView( TargetView& target, const SourceView& source,
...
   \param time The current time
   \param array The array to write
   \param gather_array Gather the array before writing to make parallel
   consistent.
 */
 template <class ExecutionSpace, class Array_t>
-void writeTimeStep( ExecutionSpace, const int time_step_index,
-                    const double time, const Array_t& array,
-                    const bool gather_array = true )
+void writeTimeStep( ExecutionSpace, const std::string& prefix,
```
streeve, line 183: We'll want a backwards compatibility wrapper that passes an empty string to this function (that we deprecate)

streeve (changes requested):
Test failures aren't your fault - see #743

grid/src/Cabana_Grid_BovWriter.hpp
DavidJoy8: I tried to create this (instead of passing an empty string, the deprecated version replicates the previous behavior of creating a prefix based on the array label).  However, there's an issue stemming from there already being another version of this function that infers the ExecutionSpace:
When one (like in the test cases) calls `writeTimeStep( "grid_cell_field_2d", 302, 3.43, *cell_field )`, it should be going to the new, non-deprecated method, but the compiler instead matches it to the templated `writeTimeStep( ExecutionSpace, const int...)`, templating ExecutionSpace to std::string.
I don't know how to fix this; I tried some stuff that would say "template ExecutionSpace to any class but std::string" but couldn't get it to work.

streeve: We'll just need to add an `enable_if` to restrict that first argument to a valid execution space - we can chat about it Monday

grid/src/Cabana_Grid_BovWriter.hpp
```diff
@@ -393,18 +394,40 @@ void writeTimeStep( ExecutionSpace, const int time_step_index,
...
+                    const double time, const Array_t& array,
+                    const bool gather_array = true )
+{
+    using exec_space = typename Array_t::execution_space;
+    writeTimeStep( exec_space{}, prefix, time_step_index, time, array,
+                   gather_array );
+}
+
+template <class ExecutionSpace, class Array_t>
+[[deprecated( "writeTimeStep now requires a filename prefix" )]] void
```
streeve, line 415: Looking at this again, I would rather not deprecate and just directly use the array label as the prefix in the variant

DavidJoy8:
Messed up my repo; closing this and will create a new pull request

streeve:
> Messed up my repo; closing this and will create a new pull request

You can just reopen, `git rebase master`, and force push to this branch if you want

PR #752, opened 2024-05-09
Fix configure error for hdf5 mpi C bindings

streeve:
```
CMake Error at /github/home/hdf5/cmake/hdf5-targets.cmake:68 (set_target_properties):
  The link interface of target "hdf5::hdf5-shared" contains:

    MPI::MPI_C

  but the target was not found.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ endif()
 
 # standard dependency macro
 macro(Cabana_add_dependency)
-  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE;VERSION;COMPONENTS" "" ${ARGN})
+  cmake_parse_arguments(CABANA_DEPENDENCY "" "PACKAGE;VERSION" "COMPONENTS" ${ARGN})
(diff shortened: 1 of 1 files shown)


dalg24:
```
<kokkos>/include/Kokkos_ExecPolicy.hpp:272:14: error: no matching conversion for static_cast from 'member_type' (aka 'unsigned long') to 'Kokkos::Impl::AtomicDataElement<Kokkos::ViewTraits<int[2], Kokkos::HostSpace, Kokkos::MemoryTraits<4>>>'
    warn |= (static_cast<IndexType>(static_cast<member_type>(bound)) != bound);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
@ldh4 this is interesting.

https://github.com/kokkos/kokkos/blob/506da184f0aa8af9fbbca5bccd8c500b81865af5/core/src/Kokkos_ExecPolicy.hpp#L126-L131
In the constraint we ensure convertibility from the user-provided argument to the policy index type but in the precondition check we do a roundtrip conversion and here it turns out the reverse conversion is not possible.

lebuller (approved):

streeve:
Docker/jenkins failures unrelated

PR #758, opened 2024-06-24
Fix BovWriter 

### LINKED ISSUE #745: BOV output failure
From https://github.com/ORNL-MDF/Finch/issues/12

Possibly missing wait/barrier or missing copy

abisner:
Adds missing host mirror copy.

Closes #745.

### DIFF
--- a/grid/src/Cabana_Grid_BovWriter.hpp
+++ b/grid/src/Cabana_Grid_BovWriter.hpp
@@ -267,7 +267,9 @@ void writeTimeStep( ExecutionSpace, const std::string& prefix,
     local_space_max.back() = owned_extents.back();
     IndexSpace<num_space_dim + 1> local_space( local_space_min,
                                                local_space_max );
-    auto owned_subview = createSubview( array.view(), local_space );
+
+    auto array_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace{}, array.view());
+    auto owned_subview = createSubview( array_host, local_space );
 
     std::array<long, num_space_dim + 1> reorder_space_size;
     for ( std::size_t d = 0; d < num_space_dim; ++d )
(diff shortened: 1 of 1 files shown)


grid/src/Cabana_Grid_BovWriter.hpp
```diff
@@ -276,9 +278,11 @@ void writeTimeStep( ExecutionSpace, const std::string& prefix,
     }
     reorder_space_size.back() = owned_extents.back();
     IndexSpace<num_space_dim + 1> reorder_space( reorder_space_size );
-    auto owned_view = createView<value_type, Kokkos::LayoutRight, memory_space>(
+
+    auto owned_view = createView<value_type, Kokkos::LayoutRight, Kokkos::HostSpace>(
         array.label(), reorder_space );
-    reorderView( owned_view, owned_subview, reorder_space, ExecutionSpace() );
+
+    reorderView( owned_view, owned_subview, reorder_space, Kokkos::DefaultHostExecutionSpace{} );
```
streeve, line 285: I don't see anything that would stop us from reordering on device, then mirroring after that kernel - what am I missing?

kwitaechong: @abisner I have segment fault error when encounter `node_halo->gather` and `Grid_Halo` CUDA test fails as well. Do you have same issue on `summit`

streeve: @kwitaechong that may be unrelated - this should only affect the grid output tests

grid/src/Cabana_Grid_BovWriter.hpp
```diff
@@ -267,7 +267,9 @@ void writeTimeStep( ExecutionSpace, const std::string& prefix,
     local_space_max.back() = owned_extents.back();
     IndexSpace<num_space_dim + 1> local_space( local_space_min,
                                                local_space_max );
-    auto owned_subview = createSubview( array.view(), local_space );
+
+    auto array_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace{}, array.view());
```
streeve, line 271: Can you try replacing this with a fence instead?

PR #759, opened 2024-07-03
Add grid_parallel_for with View of IndexSpaces

kwitaechong:
add the case of parallel_for with view of IndexSpace


streeve:
Closing until we have a need for this - note that one performance issue here is needing to launch a kernel every time to get the range indices from the Views (in contrast to the Array version). If we do ever need this, we should create a `GridPolicy` that stores this information instead

PR #761, opened 2024-07-30
Remove configure error for non-parallel HDF5 support

streeve:
Warn and disable HDF5 support instead


streeve:
ping @kwitaechong

kwitaechong (approved):

Issue #766, opened 2024-09-04
Running in to Installation issues while installing on Mac OS


dineshadepu:
Hi all,

I am developing my code on my new Macbook, and in process I am installing Cabana on it. I was able to install `Kokkos` successfully.  However, I am running into an error, while installing `Cabana`. Here is the following error message:

```
dineshadepu@Dineshs-MacBook-Pro:~/softwares/Cabana/build$ cmake     -D CMAKE_BUILD_TYPE="Release"     -D CMAKE_PREFIX_PATH="$KOKKOS_INSTALL_DIR;$SILO_INSTALL_DIR;$HDF_INSTALL_DIR"     -D CMAKE_INSTALL_PREFIX="$CABANA_INSTALL_DIR"     -D Cabana_REQUIRE_OPENMP=ON     -D Cabana_ENABLE_EXAMPLES=ON     -D Cabana_ENABLE_TESTING=OFF     -D Cabana_REQUIRE_HDF5=ON     -D Cabana_ENABLE_PERFORMANCE_TESTING=OFF     .. ;
-- The CXX compiler identification is Clang 18.1.8
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/homebrew/Cellar/llvm/18.1.8/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found OpenMP_CXX: -fopenmp=libomp (found version "5.1")
-- Found OpenMP: TRUE (found version "5.1") found components: CXX
-- Enabled Kokkos devices: OPENMP;SERIAL
-- Found Kokkos_DEVICES: OPENMP
-- The C compiler identification is Clang 18.1.8
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/homebrew/Cellar/llvm/18.1.8/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found CLANG_FORMAT: /opt/homebrew/opt/llvm/bin/clang-format (found suitable version "18.1.8", minimum required is "14")
[...]

PR #771, opened 2024-09-16
fixup: use hdf5 library if available

streeve:


streeve:
@dineshadepu can you check if this resolves the issue (without needing to modify Cabana) from #766?

junghans (approved):

PR #774, opened 2024-09-24
Rewrite grid particle init with parallel_scan filter after creation

streeve:
Removes the atomic and should provide more robust support


streeve:
@abisner let me know if this helps you at all

PR #775, opened 2024-09-24
Add non-MPI remove feature

streeve:
Current way to remove particles goes through an MPI codepath with `migrate`. This adds a non-MPI feature that will be used in particle creation and anywhere random particles need to be removed


junghans (approved):

streeve:
Ignoring docker failure

PR #776, opened 2024-09-24
fixup: MPI test type

streeve:
Warnings from mpich


streeve:
Ignoring docker failure

PR #787, opened 2024-10-05
cmake: remove obsolete pc file

junghans:
The pkg-config with kokkos is just incomplete, so let's drop it.


streeve (approved):
Sounds good to me

PR #798, opened 2024-10-23
Get ParticleList of Vector/Matrix

kwitaechong:
get view of particle vector/matrix


streeve:
I would just add this in #797

streeve:
Should be moved inside #797

PR #799, opened 2024-10-30
Add support for all heFFTe MPI communication backends

rodrigoacb:
This PR extends the available backends available in heFFTe using a enum class to pick the chosen reshape/communication pattern. 

Previously supported: alltoallv and p2p

This adds alltoall and p2p_plined

This changes were done in cooperation with @rhalver

### DIFF
--- a/grid/src/Cabana_Grid_FastFourierTransform.hpp
+++ b/grid/src/Cabana_Grid_FastFourierTransform.hpp
@@ -19,8 +19,6 @@
 #include <Cabana_Grid_Array.hpp>
 #include <Cabana_Grid_Types.hpp>
 
-#include <Cabana_Utils.hpp> // FIXME: remove after next release.
-
 #include <Kokkos_Core.hpp>
 #include <Kokkos_Profiling_ScopedRegion.hpp>
 
@@ -98,13 +96,24 @@ struct is_matching_array<
 {
 };
 
+/*!
+  \brief Choices of heFFTe MPI communication patterns
+*/
+enum class FFTCommPattern : unsigned int
+{
+    alltoallv = 0u,
+    p2p = 1u,
+    alltoall = 2u,
+    p2p_plined = 3u
+};
+
 //---------------------------------------------------------------------------//
 /*!
   \brief Parameters controlling details for fast Fourier transforms.
 */
 class FastFourierTransformParams
 {
-    bool alltoall = true;
+    FFTCommPattern FFTcomm = FFTCommPattern::alltoallv;
     bool pencils = true;
     bool reorder = true;
 
@@ -113,7 +122,18 @@ class FastFourierTransformParams
       \brief Set MPI communication strategy.
       \param value Use all to all MPI communication.
     */
-    void setAllToAll( bool value ) { alltoall = value; }
+    void setAlltoAll( FFTCommPattern value ) { FFTcomm = value; }
+    /*!
+      \brief Set MPI communication strategy.
+      \param value Use all to all MPI communication.
+    */
+    void setAlltoAll( bool value )
+    {
+        if ( value )
+            FFTcomm = FFTCommPattern::alltoallv;
+        else
+            FFTcomm = FFTCommPattern::p2p;
+    }
     /*!
       \brief Set data exchange type (pencil or slab).
       \param value Use pencil (true) or slab (false) decomposition.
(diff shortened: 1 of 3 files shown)


streeve:
Thanks for the improvements! Do you mind splitting into two PRs? I think I'll want someone else to at least co-review the spline details, but the FFT should be quick to merge

grid/src/Cabana_Grid_FastFourierTransform.hpp
```diff
@@ -97,13 +95,21 @@ struct is_matching_array<
     : public std::true_type
 {
 };
+enum class FFTCommPattern : unsigned int
```
streeve, line 98: @sr7cb hey Sanil, what kind of communication options would we expect to interface with in fftx? Trying to make sure our interface is generic enough

grid/unit_test/tstFastFourierTransform.hpp
```diff
@@ -39,7 +39,8 @@ void calculateFFT( bool use_default, bool use_params,
     Experimental::FastFourierTransformParams params;
 
     // Set MPI communication
-    params.setAllToAll( true );
+    params.setAlltoAll( Cabana::Grid::Experimental::FFTCommPattern::alltoallv );
+    params.setAlltoAll( true );
```
streeve, line 43: Do we need to set with both?

grid/src/Cabana_Grid_FastFourierTransform.hpp
sr7cb: Unfortunately I don't have enough familiarity with our MPI backend to provide a concrete answer. Please go ahead with your process and we can update/revisit down the road.

rodrigoacb:
> Thanks for the improvements! Do you mind splitting into two PRs? I think I'll want someone else to at least co-review the spline details, but the FFT should be quick to merge

Hi @streeve Sam! We also thank you for maintaining Cabana! I don't mind splitting the PRs. Should I leave this one for the FFT and create a new one for the Splines?

I had a look into the actions, and I think I have already addressed it with fixes that I can commit once the PRs are at the places.

grid/unit_test/tstFastFourierTransform.hpp
rodrigoacb: You are right, only one is needed. I will remove the boolean variation. Its behavior and usage is commented at the example [example/grid_tutorial/10_fft_heffte/heffte_fast_fourier_transform_example.cpp](https://github.com/ECP-copa/Cabana/pull/799/files#diff-0ed0f7a30244d6118a480989c9185939d650f85b0c619c95ec62719d57664458) for backward compatibility

streeve:
> > Thanks for the improvements! Do you mind splitting into two PRs? I think I'll want someone else to at least co-review the spline details, but the FFT should be quick to merge
> 
> Hi @streeve Sam! We also thank you for maintaining Cabana! I don't mind splitting the PRs. Should I leave this one for the FFT and create a new one for the Splines?
> 
> I had a look into the actions, and I think I have already addressed it with fixes that I can commit once the PRs are at the places.

Yes, that sounds great!

grid/src/Cabana_Grid_FastFourierTransform.hpp
```diff
@@ -129,7 +149,7 @@ class FastFourierTransformParams
       \brief Get MPI communication strategy.
       \return Using AllToAll or not.
     */
-    bool getAllToAll() const { return alltoall; }
```
streeve, line 132: @jstewart28 @patrickb314 just wanted to check with you both since you're using Cabana+heffte. I don't think this will affect you, but this get method will break compatibility if I missed where you use it. I noticed you have a large switch/case block - does this help you?

streeve:
@rodrigoacb sorry for the delay - we will want to rebase this to remove the splines. I'm happy to do that if needed

rodrigoacb:
Hi @streeve, no problem! Thanks, I will be happy if you can rebase to remove the splines and get the repo to the point it can be merged

grid/src/Cabana_Grid_FastFourierTransform.hpp
JStewart28: @streeve as long as `setAllToAll` still accepts `true` (it looks like it does) this won't break anything. The switch block allows us to dynamically set Alltoall parameters at runtime, so if anything, we can expand this switch block to cover the additional Alltoall parameters.

JStewart28 (approved):

junghans (approved):

PR #801, opened 2024-11-17
fix Cabana_Remove

kwitaechong:
fix Cabana_Remove


streeve:
Closing: not necessary

PR #810, opened 2025-02-14
CI: run on ubuntu-latest

streeve:
GitHub retiring 20.04


streeve:
@junghans should we use a tagged version instead?

junghans:
As we are spinning up our container right after I don't think it matters much. I would just use `ubuntu-latest`.

junghans (approved):

PR #811, opened 2025-03-02
cmake: append to CMAKE_MODULE_PATH instead of overwrite

junghans:
Then the user can set `CMAKE_MODULE_PATH` from the outside as well.


junghans:
ping @streeve

Issue #757, opened 2024-05-31
Variable radius neighbor search


streeve:
Add support for a per-particle search radius. In the context of the `VerletList`, a fixed fine (maybe intermediate) `LinkedCellList` background grid should enable local variation in `LinkedCellStencil` sizes based on the local cutoff

streeve:
Added in #805

PR #812, opened 2025-03-06
Guarantee reciprocal neighbors

streeve:
In cases where non-uniform cutoff radii are used. This now ensures that if `i` neighbors `j`, `j` also neighbors `i`. Follow up to #757 

Test updated to reflect this reciprocity

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -348,9 +348,11 @@ struct VerletListBuilder
                             {
                                 // See if we should actually check this box for
                                 // neighbors.
-                                if ( linked_cell_list.cellStencil()
-                                         .grid.minDistanceToPoint(
-                                             x_p, y_p, z_p, i, j, k ) <= rsqr )
+                                if ( withinCutoff(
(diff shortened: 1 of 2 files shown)


lebuller (approved):

core/src/Cabana_VerletList.hpp
```diff
@@ -295,9 +348,11 @@ struct VerletListBuilder
                             {
                                 // See if we should actually check this box for
                                 // neighbors.
-                                if ( linked_cell_list.cellStencil()
-                                         .grid.minDistanceToPoint(
-                                             x_p, y_p, z_p, i, j, k ) <= rsqr )
+                                if ( withinCutoff(
```
streeve, line 351: @lebuller I don't think this was here when you reviewed. Trying to decide if we need more testing of this feature or not, but will merge if not

lebuller: It was not, just looked over it. Looks good to me. I'm not sure if this needs more testing or not. Not sure the right way to test this block more rigorously than we already do.

PR #817, opened 2025-04-28
Collector

JStewart28:
Additions:
  1. Requires CXX standard 20
  2. Added a Collector class.
  3. Added createFromImportsOnly and createFromImportsAndTopology functions to CommunicationPlan.
  4. Added tests for Collector.
  5. Added tests for createFromImportsOnly and createFromImportsAndTopology.
  6. Modified cmake so test binaries are installed when testing is enabled.


CMakeLists.txt
```diff
@@ -14,9 +14,13 @@
 #------------------------------------------------------------------------------#
 cmake_minimum_required(VERSION 3.16)
 
-project(Cabana LANGUAGES CXX)
+project(Cabana LANGUAGES C CXX)
 set(PROJECT_VERSION "0.8.0-dev")
 
+# Set the C++ version we need - 20 for requires
+set(CMAKE_CXX_STANDARD 20)
```
streeve, line 21: We set this in the `src/CMakeLists` to allow for using newer standards - were things not working without this change?

cmake/test_harness/test_harness.cmake
```diff
@@ -25,8 +25,14 @@ macro(Cabana_add_tests_nobackend)
     set(_target Cabana_${CABANA_UNIT_TEST_PACKAGE}_${_test}_test)
     add_executable(${_target} tst${_test}.cpp ${TEST_HARNESS_DIR}/unit_test_main.cpp)
     target_link_libraries(${_target} PRIVATE ${CABANA_UNIT_TEST_PACKAGE} ${gtest_target})
+
+    # Install the test executable
+    install(TARGETS ${_target}
```
streeve, line 30: I've never installed testing files - what does this enable? I would merge this separately

streeve (commented):
Can you add a bit more description on use cases? I understand the concept, but am trying to think of where I would use this

cmake/test_harness/test_harness.cmake
junghans: Same, and so I would install them in /usr/share/... not /usr/bin.

JStewart28: I chose to install the test binaries to make it easier to debug the new `CommunicationPlan` and `Collector` tests on Tioga using Spack. On systems like Tioga, manually building Cabana can be more difficult, and Spack simplifies the process. However, Spack removes its staging directories after installation, so without installing the tests, you’d need to run `spack install --keep-stage` and then manually locate the staging directory to access the test binaries. 

By installing the test executables, I can run them directly from the install tree in job scripts without having to `spack uninstall cabana` and `spack install --keep-stage` before each job I submit.

Do you want me to create a new pull request just for this and move the tests to `/usr/share/tests`?

CMakeLists.txt
JStewart28: This is residual from previous development that used `requires` from C++20. We don't need it anymore. I removed these lines from CMakeLists.txt

cmake/test_harness/test_harness.cmake
junghans: > Do you want me to create a new pull request just for this and move the tests to /usr/share/tests?

I think you can make the change of the install location part of this PR.

JStewart28:
> Can you add a bit more description on use cases? I understand the concept, but am trying to think of where I would use this

For the Collector? Where do you want me to add it?

streeve:
> For the Collector? Where do you want me to add it?

I should have specified more detail in the pull request description itself

JStewart28:
> > For the Collector? Where do you want me to add it?
> 
> I should have specified more detail in the pull request description itself

I updated the description with two examples.

JStewart28:
@streeve @junghans do you have any other comments on this pull request?

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 #------------------------------------------------------------------------------#
 cmake_minimum_required(VERSION 3.16)
 
-project(Cabana LANGUAGES CXX)
+project(Cabana LANGUAGES C CXX)
```
junghans, line 17: Why is C needed?

cmake/CabanaConfig.cmakein
```diff
@@ -44,3 +44,8 @@ set(Cabana_ENABLE_HDF5 @Cabana_ENABLE_HDF5@)
 if(Cabana_ENABLE_HDF5)
   find_dependency(HDF5 REQUIRED COMPONENTS C)
 endif()
+set(Cabana_ENABLE_TESTING @Cabana_ENABLE_TESTING@)
+if(Cabana_ENABLE_TESTING)
+  install(TARGETS ${TEST_EXECUTABLE}
```
junghans, line 49: I don't understand why this needs to be in the config file.

cmake/test_harness/test_harness.cmake
```diff
@@ -87,6 +93,11 @@ macro(Cabana_add_tests)
       target_include_directories(${_target} PRIVATE ${_dir}
         ${TEST_HARNESS_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
       target_link_libraries(${_target} PRIVATE ${CABANA_UNIT_TEST_PACKAGE} ${gtest_target})
+
+      # Install the test executable
+      install(TARGETS ${_target}
+              RUNTIME DESTINATION share/tests)
```
junghans, line 99: Use CMake_INSTALL_DATADIR instead of share.

junghans (changes requested):

cmake/CabanaConfig.cmakein
JStewart28: I removed these lines from the config file

CMakeLists.txt
JStewart28: C is not needed, so I removed it.

cmake/test_harness/test_harness.cmake
```diff
@@ -25,8 +25,14 @@ macro(Cabana_add_tests_nobackend)
     set(_target Cabana_${CABANA_UNIT_TEST_PACKAGE}_${_test}_test)
     add_executable(${_target} tst${_test}.cpp ${TEST_HARNESS_DIR}/unit_test_main.cpp)
     target_link_libraries(${_target} PRIVATE ${CABANA_UNIT_TEST_PACKAGE} ${gtest_target})
+
+    # Install the test executable
+    install(TARGETS ${_target}
+            RUNTIME DESTINATION CMake_INSTALL_DATADIR)
```
junghans, line 31:
```suggestion
            RUNTIME DESTINATION ${CMAKE_INSTALL_DATADIR}/tests)
```

cmake/test_harness/test_harness.cmake
```diff
@@ -87,6 +93,11 @@ macro(Cabana_add_tests)
       target_include_directories(${_target} PRIVATE ${_dir}
         ${TEST_HARNESS_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
       target_link_libraries(${_target} PRIVATE ${CABANA_UNIT_TEST_PACKAGE} ${gtest_target})
+
+      # Install the test executable
+      install(TARGETS ${_target}
+              RUNTIME DESTINATION CMake_INSTALL_DATADIR)
```
junghans, line 99:
```suggestion
              RUNTIME DESTINATION ${CMake_INSTALL_DATADIR}/Cabana/tests)
```

JStewart28:
@junghans changes made

cmake/test_harness/test_harness.cmake
```diff
@@ -87,6 +93,11 @@ macro(Cabana_add_tests)
       target_include_directories(${_target} PRIVATE ${_dir}
         ${TEST_HARNESS_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
       target_link_libraries(${_target} PRIVATE ${CABANA_UNIT_TEST_PACKAGE} ${gtest_target})
+
+      # Install the test executable
+      install(TARGETS ${_target}
```
streeve, line 98: How do either of you feel about adding a CMake option to enable this? Still seems like a non-standard need to me

streeve (changes requested):
Looks like a couple of easy warning fixes in the CI and one failure that I saw on my end. Can you also add a simple example mirroring the current migrate case? We can rearrange the numbers or add it as `example/core_tutorial/11_collector`

cmake/test_harness/test_harness.cmake
JStewart28: I'm for it. You two have more experience with what is standard and non-standard than me.

core/unit_test/tstCollector.hpp
```diff
@@ -0,0 +1,766 @@
...
+    Kokkos::fence();
+
+    // Do the migration
+    data.resize( collector->numOwned() + collector->totalNumImport() );
+    Cabana::migrate( *collector, data );
+
+    // Check the change in size and import counts.
+    if ( my_rank == 0 )
+    {
+        EXPECT_EQ( data.size(), 4 ) << "Rank " << my_rank << "\n";
```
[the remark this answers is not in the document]
JStewart28: Which system are you running on?

JStewart28:
> Looks like a couple of easy warning fixes in the CI and one failure that I saw on my end. Can you also add a simple example mirroring the current migrate case? We can rearrange the numbers or add it as `example/core_tutorial/11_collector`

I added an example for the Collector as `example/core_tutorial/11_collector`

JStewart28:
The code for the `distributeData` / `collectData` and `migrate` functions is duplicated in the Distributor and Collector classes. I think these functions should be generalized to accept a Distributor or a Collector to eliminate duplication. Thoughts? If so, how should this be designed?

cmake/test_harness/test_harness.cmake
junghans: I agree let’s wrap it in an option (e.g. `Cabana_INSTALL_TEST_EXECUTABLES`) that is off by default

core/unit_test/tstCollector.hpp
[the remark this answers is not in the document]
JStewart28: This was a flaw in the test logic for communicator sizes less than 3 in `tstCommunicationPlan.hpp` and not equal to 4 in `tstCollector.hpp`. The tests have been fixed.

streeve (commented):
Tests look good now - can we move the Collector and Distributor into a shared `Migrate.hpp` file like we discussed? 

Not sure if it's possible or worthwhile to split the CommPlan functions into some helpers that are shared - what do you think?

JStewart28:
> Tests look good now - can we move the Collector and Distributor into a shared `Migrate.hpp` file like we discussed?

Just pushed the change that combines Collector and Distributor into `Cabana_Migrate.hpp`. 

> Not sure if it's possible or worthwhile to split the CommPlan functions into some helpers that are shared - what do you think?

We will make a sperate pull request for the restructuring that will add various communication spaces. I'll follow-up on our email chain about what we think makes sense for the restructuring.

CMakeLists.txt
```diff
@@ -238,6 +242,10 @@ if(Cabana_ENABLE_TESTING)
   set(TEST_HARNESS_DIR ${CMAKE_SOURCE_DIR}/cmake/test_harness)
   include(cmake/test_harness/test_harness.cmake)
   enable_testing()
+
+  install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/core/unit_test
```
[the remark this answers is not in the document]
JStewart28: Yes. The change has been made.

JStewart28:
@streeve if all the CI tests pass, is there anything else needed in this pull request before it can be completed and closed?

JStewart28:
@streeve all tests have passed. Are we ready to merge this pull request?

core/src/Cabana_Migrate.hpp
```diff
@@ -632,47 +641,294 @@ void migrate( ExecutionSpace, const Distributor_t& distributor,
...
+void migrate(
+    ExecutionSpace exec_space, const Distributor<MemorySpace>& distributor,
+    const AoSoA_t& src, AoSoA_t& dst,
+    typename std::enable_if<( is_distributor<Distributor<MemorySpace>>::value &&
+                              is_aosoa<AoSoA_t>::value ),
+                            int>::type* = 0 )
+{
+    // Check that src and dst are the right size.
+    if ( src.size() != distributor.exportSize() )
+        throw std::runtime_error( "Source is the wrong size for migration!" );
```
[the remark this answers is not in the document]
JStewart28: Added.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -924,6 +924,530 @@ class CommunicationPlan
...
+            send_to( i ) = mpi_statuses[i].MPI_SOURCE;
+            _total_num_export += send_counts( i );
+        }
+
+        // Extract the export ranks and number of exports and then flag the
+        // send ranks.
+        for ( int r = 0; r < num_recvs; ++r )
+        {
+            int export_to = send_to( r );
+            if ( export_to > -1 )
```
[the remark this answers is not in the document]
JStewart28: `send_to(i)` is populated from `mpi_statuses[i].MPI_SOURCE`. (See the code a few lines above) `mpi_statuses[i].MPI_SOURCE` will never return <0. If the user used an import rank of -1 there will be an (unhandled) error on line 1226. But I did add an else statement (that should never be reached) that throws a runtime error.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -924,6 +924,530 @@ class CommunicationPlan
...
+        auto element_export_ranks = Kokkos::create_mirror_view_and_copy(
+            memory_space(), element_export_ranks_h );
+
+        auto counts_and_ids2 = Impl::countSendsAndCreateSteering(
+            exec_space, element_export_ranks, comm_size,
+            typename Impl::CountSendsAndCreateSteeringAlgorithm<
+                ExecutionSpace>::type() );
+
+        // Copy indices_send to device mempry before returning
+        // auto export_indices_d = Kokkos::create_mirror_view_and_copy(
```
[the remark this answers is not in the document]
JStewart28: I removed the commented code. Since we use gpu-to-gpu communication, `export_indices` can be used in MPI communication no matter which memory space it is in, so it doesn't need to be copied to device memory here.

core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -924,6 +924,530 @@ class CommunicationPlan
...
```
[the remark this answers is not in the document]
JStewart28: Same as above

core/src/Cabana_Migrate.hpp
```diff
@@ -162,60 +162,222 @@ struct is_distributor
 {
 };
 
+//---------------------------------------------------------------------------//
+/*!
+  \brief A communication plan for importing data from one uniquely-owned
```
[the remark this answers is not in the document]
JStewart28: I clarified that the Collector duplicates data and the Distributor moves data.

streeve:
> With clarifying that the Collector duplicates data and the Distributor moves data, I changed the language around the term "migrate". When using a Distributor, "distribute" is used, and when using a Collector, "collect" is used. If a function takes either a Distributor or a Collector, I use "migrate".
> 
> Is migrate still the best term for the general case? Migrate implies moving data from one decomposition to another, no duplication. The Distributor does this but the Collector does not. I'm open to refactoring "migrate" to a different name to avoid confusion.

Sorry to delay here, but I misunderstood this point early on, despite your clear descriptions. This sounds more like an import `Halo` - it similarly uses an offset and duplicates rather than keeping unique ownership (but otherwise simply changes the import/export perspective). This would also align with the `gather/scatter` concepts, rather than `migrate`

Let me know if I'm fundamentally missing something here (or if you want to discuss again) and I'll keep up with this better to get this merged.

streeve:
> I know this is a 90-degree turn with a bit of back-tracking - but what if we moved the Collector to be in `Cabana_Halo.hpp`, renamed the Collector to an `ImportHalo` and the existing `Halo` to an `ExportHalo`? (And adjusted the helpers in `Cabana_Halo.hpp` appropriately) Or did something with templating to distinguish between import and export halos, to avoid breaking existing code? And then reverted the `Distributor` back to its original state? I think this design makes a lot more sense, because like you said, the Collector functions more like a Halo than a Distributor.
> 
> What do you think? I'll discuss this with Patrick next week when he is back from vacation.

I think this is the right idea - a tag for `Import`/`Export` (default) sounds good (`ImportHalo`/`ExportHalo` could work too, but the backwards compatibility is a bit more cumbersome)

Send a meeting invite if you think that's helpful

streeve:
> Going with a tag is good. Since the import/export decision influences how the communication plan is built, maybe `Cabana::CommDriver::Export` and `Cabana::CommDriver::Import`?
> 
> Do you want me to do the restructuring in this pull request, or close this one and open a new pull request? Keeping this named as "Collector" will be misleading.

Yeah, a new PR that links here makes sense - that will also separate the discussion pre/post Distributor/Halo

I would default to the simplest tag possible `Cabana::Export/Import` (most often in the past we've eventually reused tags but needed renaming to make them general enough)

PR #818, opened 2025-04-30
Address Clang-Tidy diagnostics

jngrad:


core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -1281,12 +1281,14 @@ class CommunicationData
         setData( particles );
 
         auto send_capacity = sendCapacity();
-        std::size_t new_send_size = total_send * _overallocation;
+        auto new_send_size = static_cast<std::size_t>(
+            static_cast<double>( total_send ) * _overallocation );
```
dalg24, line 1285: What warning did you get?
This fix looks silly.

jngrad: `clang-diagnostic-implicit-int-float-conversion`, `clang-diagnostic-float-conversion`

dalg24: Did we commit to build w/o warnings with that flag?
On the one hand, it is only one occurrence and sure we can confirm the intent, on the other hand I feel like it is a mouthful for conversions that in that occurrence are somewhat obvious.
Do we have a policy for this?

streeve: I don't have a problem with it, especially as a single instance. But no, we don't have any real policy/setup for clang-tidy at the moment

streeve (approved):

PR #820, opened 2025-04-30
CI: bump hypre version in docker build

### LINKED ISSUE #819: CI: Docker workflow breaks with GCC 15.0.1
Dependency hypre-space/hypre v2.22.1 generates compiler errors in CI with GNU 15.0.1:
```
In file included from /home/kokkos/src/hypre/src/distributed_ls/pilut/DistributedMatrixPilutSolver.h:115,
                 from /home/kokkos/src/hypre/src/distributed_ls/pilut/comm.c:25:
/home/kokkos/src/hypre/src/distributed_ls/pilut/./struct.h:43:26: error: 'bool' cannot be defined via 'typedef'
   43 |    typedef unsigned char bool;
      |                          ^~~~
/home/kokkos/src/hypre/src/distributed_ls/pilut/./struct.h:43:26: note: 'bool' is a keyword with '-std=c23' onwards
/home/kokkos/src/hypre/src/distributed_ls/pilut/./struct.h:43:4: warning: useless type name in empty declaration
   43 |    typedef unsigned char bool;
      |    ^~~~~~~
```

junghans:
Fix #819


junghans:
@streeve docker works, but now the CI fails somewhere else ;-)

dalg24:
(On my cell) it looks like the compiler raises warnings in the Kokkos header. Are we not treating them as "system" headers?

streeve:
> (On my cell) it looks like the compiler raises warnings in the Kokkos header. Are we not treating them as "system" headers?

Isn't that the default for `find_package`? It seems like only fedora is having a problem

junghans:
> > (On my cell) it looks like the compiler raises warnings in the Kokkos header. Are we not treating them as "system" headers?
> 
> Isn't that the default for `find_package`? It seems like only fedora is having a problem

I thought so as well. Fedora is the relatively new gcc-15.0.1

junghans:
And it actually says `-isystem /github/home/kokkos/include` on the compile line.

streeve (approved):

PR #823, opened 2025-06-17
Added namespaces to (most) error messages

JStewart28:
> [streeve](https://github.com/streeve): I like that you added the details of the call site in the error messages - can you add that for the Distributor?

Per our discussion in PR #817


JStewart28:
Accidently merged branches the wrong way. Need to close this and re-open.

PR #831, opened 2025-07-21
Fix `VerletList` constructor

FabienPean-Virtonomy:
Enable using `Kokkos::View` in `Cabana::VerletList` constructor overload with per-particle cutoff radius. It seems it was omitted by mistake (or by design?) from the constructor. 

It could be beneficial to add the constraint on `RadiusSlice` template parameter too ?

### DIFF
--- a/core/src/Cabana_VerletList.hpp
+++ b/core/src/Cabana_VerletList.hpp
@@ -864,7 +864,8 @@ class VerletList
                 const typename PositionSlice::value_type grid_min[3],
                 const typename PositionSlice::value_type grid_max[3],
                 const std::size_t max_neigh = 0,
-                typename std::enable_if<( is_slice<PositionSlice>::value ),
+                typename std::enable_if<( is_slice<PositionSlice>::value ||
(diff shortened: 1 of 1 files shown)


FabienPean-Virtonomy:
@streeve This line will be triggered in debug build when using `Kokkos::View` with variable radius per particle due to difference of semantics for `size()` between AoSoA and View. Is there a generic way to recover the same information for the assert or there must be a constexpr split depending on the input type?

https://github.com/ECP-copa/Cabana/blob/d9900ea07dd0c4c518448c0fc71f518b693093e9/core/src/Cabana_VerletList.hpp#L193

streeve:
> @streeve This line will be triggered in debug build when using `Kokkos::View` with variable radius per particle due to difference of semantics for `size()` between AoSoA and View. Is there a generic way to recover the same information for the assert or there must be a constexpr split depending on the input type?

Sorry for the delay, yes you're right we should be using the free function `Cabana::size( positions )` in that instance to handle Views as well

PR #833, opened 2025-08-05
fail when legacy kokkos views are off

cwsmith:
If kokkos was built with non-legacy views then the cmake var Kokkos_ENABLE_IMPL_VIEW_LEGACY is not defined.  This was tested with kokkos @ develop b06886a37.


streeve:
Thanks for the fix - I tried to simplify here and also noticed that the version was an issue. Does this still work on your end?

streeve (approved):

PR #835, opened 2025-08-15
Fix race condition when constructing CommunicationPlan from imports

JStewart28:
Sometimes `CommunicationPlan` test 10 hangs. I believe there is a race condition between the two rounds of `Isends` and `Irecvs` when constructing a `CommunicationPlan` from imports only. The test does not hang after adding an `MPI_Barrier` after the first `MPI_Waitall`.

This pull request adds the barriers to fix the bug.

### DIFF
--- a/core/src/Cabana_CommunicationPlan.hpp
+++ b/core/src/Cabana_CommunicationPlan.hpp
@@ -1063,6 +1063,12 @@ class CommunicationPlan
         if ( MPI_SUCCESS != ec )
             throw std::logic_error( "Failed MPI Communication" );
 
+        // This barrier is needed to ensure all the above Isends and IRecvs
+        // complete before the next echange starts. If there is no barrier
+        // sometimes the send_to data will be populated incorrectly and cause
+        // the code to hang.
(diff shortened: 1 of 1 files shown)


core/src/Cabana_CommunicationPlan.hpp
```diff
@@ -1083,7 +1084,8 @@ class CommunicationPlan
             for ( std::size_t j = 0; j < _num_export[i]; j++ )
             {
                 MPI_Irecv( export_indices.data() + idx, 1, MPI_INT,
-                           _neighbors[i], mpi_tag, comm(), &mpi_requests[idx] );
+                           _neighbors[i], mpi_tag + 1, comm(),
```
streeve, line 1087: Can you add a comment mentioning that incrementing the tag is a way to ensure the messages are processed in the right order? Probably just need one outside the loop in both functions

streeve:
Nice solution! This is great. I'll take another look at the derived class barrier usage to see if we can close that old issue

streeve (approved):

PR #845, opened 2025-09-12
Fix Cabana::Remove when nothing to remove

kwitaechong:
Fix Cabana_Remove to return without do anything when there is nothing to remove

Expand tests for:
- Slice vs View removal
- Number of particles removed

### DIFF
--- a/core/unit_test/tstRemove.hpp
+++ b/core/unit_test/tstRemove.hpp
@@ -20,7 +20,7 @@
 namespace Test
 {
 
-void testRemove()
+void testRemoveSlice()
 {
     int num_particle = 200;
     Cabana::AoSoA<Cabana::MemberTypes<int>, TEST_MEMSPACE> aosoa(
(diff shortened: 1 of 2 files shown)


streeve (changes requested):
Can you add a test that doesn't remove ?

core/unit_test/tstRemove.hpp
```diff
@@ -25,40 +25,45 @@ void testRemove()
...
             {
                 keep_slice( p ) = 0;
             }
         } );
 
-    // Remove only odd particles.
     int new_num_particle = num_particle / 2;
-    Cabana::remove( TEST_EXECSPACE{}, new_num_particle, keep_particle, aosoa );
+    Cabana::remove( TEST_EXECSPACE{}, new_num_particle, keep_slice, aosoa );
     EXPECT_EQ( aosoa.size(), new_num_particle );
```
streeve, line 57: Can you add one more check that each particle left after the removal has a value of 1 (that we removed the right particles)? That was intended to be included in the original version

PR #855, opened 2025-10-22
Allow `ParameterPack` to be constructed from elements that are not default constructible

JBludau:
The docstrings state that `ParameterPack` requires the elements to be copyable to the device.
When `makeParameterPack` is used it implicitly requires default constructability and assignability of the elements. This would not be necessary if `ParameterPack` would provide a ctor that allows to go through the element's copyctor.

This PR adds a ctor to `ParameterPack` that uses the element's copy ctor.
I am not sure if we want the ctor to be part of the public interface or make it private and have a new convenience function (like `makeParameterPack`) that is friend of the class and can construct it.

This came up in CabanaPD for multiforce models where we would like to have a model that redirects to a multitude of models that are stored in the `ParameterPack`. Nevertheless, the models are not default constructible. An alternative to this PR would be to make the models in CabanaPD default constructible and assignable.

### DIFF
--- a/core/src/Cabana_ParameterPack.hpp
+++ b/core/src/Cabana_ParameterPack.hpp
@@ -62,6 +62,7 @@ struct PackTypeAtIndex
 template <std::size_t N, typename T>
 struct ParameterPackElement
 {
+    //! Ctor from variadic template args
     explicit ParameterPackElement( T value )
         : _m( value )
     {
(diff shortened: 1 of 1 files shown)


JBludau:
@streeve could you trigger the CI on this?

JBludau:
hmm ok. I guess the doxygen build canceled the other ones

streeve (approved):

PR #856, opened 2025-10-23
Fixes to use hypre 3.0 and hypre with ROCM

patrickb314:
Changes CMakeLists.txt to require hypre 3.0.0 and the cabana hypre solver to work with ROCMSpace now that hypre 3.0 supports it.


patrickb314:
Meant to send this to a different fork on which I'm testing, not here. Apologies.

PR #858, opened 2025-10-28
Fix MySoA struct breaking CI tests

JStewart28:
The CI errors have been fixed by editing the struct `MySoA` in `core/src/unit_test/tstAoSoA.hpp`:
```c++
template <int VLEN, int D1, int D2, int D3>
struct MySoA
{
    float m0[D1][D2][D3][VLEN];
    int m1[VLEN];
    double m2[D1][VLEN];
    double m3[D1][D2][VLEN];
    
    KOKKOS_INLINE_FUNCTION
    MySoA(){};
};
```
Adding the constructor fixes the Ubuntu/OpenSUSE warnings and adding `KOKKOS_INLINE_FUNCTION` fixes a "calling a host function on the device" error when building with HIP.

### DIFF
--- a/core/unit_test/tstAoSoA.hpp
+++ b/core/unit_test/tstAoSoA.hpp
@@ -531,8 +531,8 @@ struct MySoA
     double m2[D1][VLEN];
     double m3[D1][D2][VLEN];
 
-    KOKKOS_DEFAULTED_FUNCTION
-    MySoA() = default;
+    KOKKOS_INLINE_FUNCTION
+    MySoA(){};
 };
 
 // Test an unmanaged AoSoA.


dalg24 (commented):
What ROCm version were you using?

dalg24:
> @dalg24 what's the advantage of using KOKKOS_DEFAULTED_FUNCTION? I did not know about this macro.

In the past some toolchains would complain when host device annotations were missing on a defaulted default constructor or destructor, and other toolchains would raise errors if you had them.  Kokkos picks a definition that works for the toolchain that is used.

streeve (approved):

dalg24:
> It's a warning about an uninitialized value in `Kokkos_ViewMapping.hpp`: [see here](https://github.com/ECP-copa/Cabana/actions/runs/18914097423/job/53992631800)

Sure, and that uninitialized value is the `MySoA` struct after default-initialization.
t's not like the empty body solves that problem, what is does it makes `MySoA` non trivially constructible.  This has implications in terms of view initialization performance.  It is not critical but the right thing to do would be to think about what we want that type to do.
Also I haven't checked what version of Kokkos was used but I can tell you that more recent versions of Kokkos would not have yielded a warning.

PR #868, opened 2026-04-02
Add support for large message sizes

JStewart28:
Added Cabana wrappers for `MPI_Isend` and `MPI_Irecv` that support message sizes larger than INT_MAX. If `MPI_VERSION >= 4.0` use `MPI_Isend_c` and `MPI_Irecv_c` versions, otherwise truncate messages to INT_MAX.

I am open to moving these functions to a different file, renaming them, or placing them in a new namespace.

### DIFF
--- a/core/src/impl/Cabana_CommunicationPlan_Mpi.hpp
+++ b/core/src/impl/Cabana_CommunicationPlan_Mpi.hpp
@@ -34,6 +34,99 @@
 namespace Cabana
 {
 
+/*!
+    \brief Helper to post non-blocking receive(s) for a single neighbor, handling
+    messages larger than INT_MAX when MPI version is less than 4 and
+    large-count messages are unavailable.
+
+    \param subview The receive buffer.
+
+    \param source Who to receive from.
+
+    \param tag The MPI message tag.
+
+    \param comm The MPI Communicator.
+
+    \param requests Requests vector to append this message to.
+*/
+template <class ViewType>
+inline void cabanaIrecv( const ViewType& subview, int source, int tag,
+                    MPI_Comm comm,
+                    std::vector<MPI_Request>& requests )
+{
+    using value_type = typename ViewType::value_type;
+    std::size_t total_bytes = subview.size() * sizeof( value_type );
+ 
+#if MPI_VERSION >= 4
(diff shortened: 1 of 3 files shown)


core/src/impl/Cabana_Halo_Mpi.hpp
```diff
@@ -69,39 +69,37 @@ Gather<HaloType, AoSoAType,
 
     // Post non-blocking receives.
     int num_n = _comm_plan.numNeighbor();
-    std::vector<MPI_Request> requests( num_n );
+    std::vector<MPI_Request> requests;
+    requests.reserve( num_n * 2 );
```
streeve, line 73: Clarification: I assume this is technically unrelated, but good practice?

core/src/impl/Cabana_CommunicationPlan_Mpi.hpp
```diff
@@ -34,6 +34,94 @@
...
+    \param source Who to receive from.
+
+    \param tag The MPI message tag.
+
+    \param comm The MPI Communicator.
+
+    \param requests Requests vector to append this message to.
+*/
+template <class ViewType>
+inline void cabanaIrecv( const ViewType& subview, int source, int tag,
```
streeve, line 53: I don't have a better name, but maybe there's something more specific to the functionality?

streeve (commented):
Quick clarification and this looks good to me. Sorry for the delay on Cabana work in general

core/src/impl/Cabana_Halo_Mpi.hpp
JStewart28: Yes. Posting as Isends and Irecvs and waiting on all of them is better practice.

core/src/impl/Cabana_CommunicationPlan_Mpi.hpp
JStewart28: What about largeMessageIsend and largeMessageIrecv?

JStewart28: Or LargeMsgIsend and LargeMsgIrecv if you want shorter function names

PR #874, opened 2026-08-04
Slice rewrite for compatibility with Kokkos mdspan Views

streeve:
This removes the incompatibility with Kokkos mdspan-based Views by removing custom `LayoutCabanaSlice` and directly using the `Kokkos::LayoutStride`. The intent is to later replace this with a fully customized Kokkos layout 

Performance details will be posted prior to merging, but initial testing shows performance degradation on the order of 10%  as compared to the previous implementation (with some cases showing small speedup) depending on AoSoA member data types

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,12 +41,6 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 # find kokkos
 find_package(Kokkos 4.1 REQUIRED)
 
-# FIXME: remove when custom layout support with mdspan Views is added.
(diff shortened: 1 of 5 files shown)


streeve:
@JBludau fyi since you've also looked into this

junghans (approved):
Works for me!

sslattery (approved):

streeve:
For posterity: 

New benchmark comparing master and this branch (still using the legacy Views) on Frontier:
slice_legacy-master_legacy-stride

New benchmark comparing master and this branch (and comparing legacy to mdspan Views) on Frontier:
slice_legacy-master_mdspan-stride

junghans:
On a related note, could be make a v0.9 (or v0.8.1) as using an old version of kokkos (or legacy views) is annoying.

